PR #130, opened 2016-05-01
aurbuild: add support for package and database signing

Celti:
Outside of chroots, this works flawlessly. Repose has had database signing support for quite a while now, just undocumented. Makepkg likewise can sign packages without an issue.

Inside the chroot, things are more complicated. You need to pass two things into the chroot, and indeed all the way to makepkg: the GPG agent socket (with a fixed location of `$GNUPGHOME/S.gpg-agent`, i.e. `$HOME/.gnupg/S.gpg-agent`), and an environment variable telling GPG where to tell the agent to open the pinentry at: either $DISPLAY or $GPG_TTY.

Unmodified makechrootpkg can handle the socket just fine; it accepts the `-d <path>` option, which is translated into the `--bind` option for systemd-nspawn — thus, `-d $HOME/.gnupg/S.gpg-agent:/build/.gnupg/S.gpg-agent` works. With that alone, you can sign packages within the chroot as long as your signing key passphrase is pre-cached in the agent. If it's not, the agent tries to open pinentry-curses on /dev/console and fails.

I have submitted a [patch](https://lists.archlinux.org/pipermail/arch-projects/2016-May/004342.html) for devtools upstream that adds a `-e <var>` option to makechrootpkg, which gets translated into `--setenv` for systemd-nspawn. Thus, `-e DISPLAY=$DISPLAY -e GPG_TTY=$GPG_TTY` will pass those into the environment. A [second patch](https://lists.archlinux.org/pipermail/arch-projects/2016-May/004343.html) allows those variables through the sudo wall in the chroot, letting makepkg and gpg see them and everything function normally.

Finally, this PR includes cleanup of option handling, more error verification, and solves another option-related error.

### DIFF
--- a/bin/aurbuild
+++ b/bin/aurbuild
@@ -16,10 +16,10 @@ mkchr_args=(-cu)
 # that pacman now considers packages inside the official repositories
 # as "local", and warns if they are newer than a custom counterpart.
 build_local() {
-    #global database pool root mkpkg_args
+    #global database pool root sign mkpkg_args
 
-    PKGDEST="$pool" LC_MESSAGES=C makepkg "${mkpkg_args[@]}"
-    repose -vf "$database" -r "$root" -p "$pool"
+    PKGDEST="$pool" LC_MESSAGES=C makepkg "${mkpkg_args[@]}" "$sign"
+    repose "$sign" -vf "$database" -r "$root" -p "$pool"
 
     printf '%s\n%s\n' "[$database]" "$(pacconf --repo="$database")" > "$tmp"/config
     sudo pacman -Syu --config="$tmp"/config
@@ -30,13 +30,13 @@ build_local() {
 # '/pkgdest' is hardcoded in makechrootpkg, packages are rebuilt even
 # if already available.
 build_chroot() {
-    #global CHROOT database pool root mkchr_args
+    #global CHROOT database pool root sign mkchr_args
 
     # Skip verification checks in the chroot instance as .gnupg paths
     # are hardcoded in makechrootpkg. Signatures are still verified
     # outside the chroot. Preserve PKGDEST with sudo -E, see FS#44827.
-    sudo -E PKGDEST="$pool" makechrootpkg "${mkchr_args[@]}" -d "$pool" -r "$CHROOT" -- --skippgpcheck
-    repose -vf "$database" -r "$root" -p "$pool"
+    sudo -E PKGDEST="$pool" makechrootpkg -r "$CHROOT" -d "$pool" "${mkchr_args[@]}" -- "$sign"
+    repose "$sign" -vf "$database" -r "$root" -p "$pool"
 }
 
 # If a package is rebuilt (same pkgver), checksums in the database no
@@ -57,9 +57,8 @@ clean_cache() {
 }
 
 usage() {
-    plain "Usage: $argv0 [-c] -a <queue> -d <database> -p <pool> -r <root> [--] [<makepkg/makechrootpkg args>]"
+    plain "Usage: $argv0 [-c] [-s] -a <queue> -d <database> -p <pool> -r <root> [--] [<makepkg/makechrootpkg args>]"
 }
-
 
 trap 'exit' INT
 trap 'rm -rf "$tmp"' EXIT
@@ -68,13 +67,14 @@ source /usr/share/makepkg/util.sh
 
 [[ -t 2 ]] && colorize
 
-while getopts :a:cd:p:r: OPT; do
+while getopts :a:cd:p:r:s OPT; do
     case $OPT in
         a) queue="$OPTARG"    ;;
         c) build=build_chroot ;;
         d) database="$OPTARG" ;;
         p) pool="$OPTARG"     ;;
         r) root="$OPTARG"     ;;
+        s) sign="--sign"      ;;
         :) case $OPTARG in
                 a) error "$argv0: No package queue specified (-a)"; usage; exit 1 ;;
                 d) error "$argv0: No repository database specified (-d)"; usage; exit 1 ;;
@@ -114,6 +114,11 @@ else
     mkpkg_args+=("${@:$OPTIND}")
 fi
 
+if [[ $sign == --sign ]]; then
+    # Pass socket and environment to the build chroot.
+    mkchr_args+=("-d" "$HOME/.gnupg/S.gpg-agent:/build/.gnupg/S.gpg-agent")
+    mkchr_args+=("-e" "GPG_TTY=$(tty)" "-e" "DISPLAY=$DISPLAY")
+fi
 
 if [[ $build == build_chroot ]]; then
     CHROOT=/var/lib/aurbuild
--- a/bin/aurbuild
+++ b/bin/aurbuild
@@ -56,6 +56,11 @@ clean_cache() {
     done < <(makepkg --packagelist)
 }
 
+usage() {
+    plain "Usage: $argv0 [-c] -a <queue> -d <database> -p <pool> -r <root> [--] [<makepkg/makechrootpkg args>]"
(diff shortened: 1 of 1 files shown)


AladW:
Thanks! It would be great if you could rebase this and split it into several pull requests, as the patches haven't landed to makechrootpkg yet.

bin/aurbuild
```diff
@@ -133,6 +133,10 @@ if [[ $build == build_chroot ]]; then
     sudo cp -v "$tmp"/pacman.conf "$CHROOT"/root/etc/pacman.conf
 fi
 
+# Canonicalise paths
```
AladW: I'd summarize the commit here, something like "Avoid issues with relative paths", as everybody knows what realpath does, but perhaps not why you've used it here.

bin/aurbuild
```diff
@@ -133,6 +133,10 @@ if [[ $build == build_chroot ]]; then
...
+pool="$(realpath "${pool}")"
+root="$(realpath "${root}")"
```
AladW: Use `realpath -e` to get built-in error reporting (all components of the path must exist), also the braces for pool/root are redundant here.

bin/aurbuild
```diff
@@ -65,32 +70,50 @@ source /usr/share/makepkg/util.sh
...
+        ?) error "$argv0: Option not recognized (-$OPTARG)"; usage; exit 1; ;;
     esac
 done
 
+# Validate arguments
+re_blank='^[[:space:]]*$'
+if [[ $queue =~ $re_blank || $database =~ $re_blank || $pool =~ $re_blank || $root =~ $re_blank ]]; then
+    error "$argv0: Missing arguments" && usage
+    exit 1
+elif [[ ! -r $queue ]]; then
```
AladW: elif lists end on the first succesful test, so in case of multiple wrong arguments, you'd only see the first. A bit more if/fi boilerplate wouldn't hurt here.

bin/aurbuild
```diff
@@ -65,32 +70,50 @@ source /usr/share/makepkg/util.sh
...
+    error "$argv0: $queue: Permission denied"
+    exit 13
```
AladW: Besides having to be readable, queue should also not be a directory. You could error on this with errno 21 (Is a directory).

bin/aurbuild
```diff
@@ -65,32 +70,50 @@ source /usr/share/makepkg/util.sh
...
+                d) error "$argv0: No repository database specified (-d)"; usage; exit 1 ;;
+                p) error "$argv0: No package pool specified (-p)"; usage; exit 1 ;;
+                r) error "$argv0: No repository root specified (-r)"; usage; exit 1 ;;
+            esac ;;
+        ?) error "$argv0: Option not recognized (-$OPTARG)"; usage; exit 1; ;;
     esac
 done
 
+# Validate arguments
+re_blank='^[[:space:]]*$'
```
AladW: It seems odd to report "Missing arguments" here, when it should already be done in the getopts loop above. I assume this is working around a getopts limitation?

bin/aurbuild
```diff
@@ -57,24 +57,24 @@ clean_cache() {
...
 
-while getopts :a:cd:p:r: OPT; do
+while getopts :a:cd:p:r:s OPT; do
     case $OPT in
         a) queue="$OPTARG"    ;;
         c) build=build_chroot ;;
         d) database="$OPTARG" ;;
         p) pool="$OPTARG"     ;;
         r) root="$OPTARG"     ;;
+        s) sign="--sign"      ;;
```
AladW: Arguably it's more flexible to create a repose option array, than a single option stored in a variable.

bin/aurbuild
```diff
@@ -114,6 +114,11 @@ else
     mkpkg_args+=("${@:$OPTIND}")
 fi
 
+if [[ $sign == --sign ]]; then
+    # Pass socket and environment to the build chroot.
+    mkchr_args+=("-d" "$HOME/.gnupg/S.gpg-agent:/build/.gnupg/S.gpg-agent")
```
AladW: This makes the same mistake as makechrootpkg, in that it breaks when `GNUPGHOME` is set to a non-standard location. Also, `/build` ?

bin/aurbuild
```diff
@@ -114,6 +114,11 @@ else
...
+    mkchr_args+=("-e" "GPG_TTY=$(tty)" "-e" "DISPLAY=$DISPLAY")
```
AladW: What if `DISPLAY` is not set?

bin/aurbuild
```diff
@@ -30,13 +30,13 @@ build_local() {
...
 build_chroot() {
-    #global CHROOT database pool root mkchr_args
+    #global CHROOT database pool root sign mkchr_args
 
     # Skip verification checks in the chroot instance as .gnupg paths
     # are hardcoded in makechrootpkg. Signatures are still verified
     # outside the chroot. Preserve PKGDEST with sudo -E, see FS#44827.
-    sudo -E PKGDEST="$pool" makechrootpkg "${mkchr_args[@]}" -d "$pool" -r "$CHROOT" -- --skippgpcheck
-    repose -vf "$database" -r "$root" -p "$pool"
+    sudo -E PKGDEST="$pool" makechrootpkg -r "$CHROOT" -d "$pool" "${mkchr_args[@]}" -- "$sign"
```
AladW: As below, this does not account for `GNUPGHOME`.

bin/aurbuild
Celti: Getopts won't tell you if you didn't give it `-a <queue>`, only if you gave it `-a` without an argument. You need both sets of validation.

bin/aurbuild
Celti: Given that all of these exit, you won't see more than one anyway. If you'd prefer I write it to set a switch and exit after all the ifs if the switch is set, I can do that.

bin/aurbuild
Celti: Noted. Will add.

bin/aurbuild
Celti: I figured that "canonicalse" was a clear enough definition, but I can expand.

bin/aurbuild
Celti: The error checking above after getopts should already ensure this doesn't error — we *know* the path exists, we just need the full path and not the relative path.

bin/aurbuild
Celti: Good point, I'll fix that.

bin/aurbuild
Celti: The same option is used for both makepkg and repose to tell them to sign the package and the database respectively. A repose options array would allow specifying compression for the database, but you can't conveniently pass options to it by sticking them on the end of your command line, so it's a bit of a wash.

bin/aurbuild
Celti: `/build` is $HOME for the build user. There is no such thing as $GNUPGHOME inside the chroot.

bin/aurbuild
Celti: Then it passes an empty variable into the chroot and gpg looks for GPG_TTY instead — I thought about and tested that.

bin/aurbuild
AladW: Then please add a comment clarifying this.

AladW:
Looks like upstream isn't interested in implementing this, so I guess I'll need to come up with a less ugly implementation of find_cached_packages in `commitpkg`.

AladW:
via 9ede771af784eb43db9ff4c70a631c02589065e3, 4b8903161b4beba07dc375a1623378355e2a3f6a and 8c9d3229f86509aa07db939ba609a436310c2936. I may implement stricter root/pool checks later, in any case, thanks for the help! :heart:

PR #143, opened 2016-05-29
add database compression options from repose to aurbuild & aursync

cmdd:


AladW:
Can you add some reasoning on how this would be useful, apart from making more repose options available to aurbuild/aursync (note there's also `--arch`, `--reflink`, `--rebuild`)? My local repository contains 424 packages; `custom.db` is 1.1 MB, `custom.files` 3.1 MB, and all packages combined 1.3 GB. As such, any size benefit from compressing the database would be neglegible.

cmdd:
I was mostly concerned with having features from repose available from the aurutils; I guess I got carried away though! I'll go ahead and close this PR then.

Issue #109, opened 2016-04-19
aurbuild: upgrade chroot once before build


AladW:
It could happen that a new library version is pushed while packages and their dependencies are build.  You could update the chroot before building by removing `makechrootpkg -u`, and using:
```
sudo arch-nspawn "$CHROOT/root" --bind "$pool" pacman -Syu --noconfirm
```

AladW:
This isn't as straightforward as originally thought, as you have to sync the databases to install packages in the chroot from the local repo. Considering the repo is cleaned with -c, I suppose you could do something like:
```
pacman -Syu # repo-only config
pacsync custom
```

PR #147, opened 2016-06-10
Update chroot once

vorph1:
Update the root chroot once before build so that the user chroot doesn't have to update base packages on each subsequent package build

### DIFF
--- a/bin/aurbuild
+++ b/bin/aurbuild
@@ -122,6 +122,9 @@ if ((chroot)); then
         sudo install -d "$CHROOT"
         sudo mkarchroot "$CHROOT"/root base base-devel
     else
+        # since makechrootpkg restores the user chroot to the state from root chroot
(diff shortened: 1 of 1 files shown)


AladW:
This was proposed in #109. The issue is that you need to update the chroot to make it aware of new packages in the local repository. You could update it once, then use a temporary config as in `build_local()`, but I'm unsure if this is worth it.

vorph1:
This does not remove the update that is run by makechrootpkg before each package is built, newly built packages in repo do get picked up correctly.

AladW:
Oh, you mean that `makechrootpkg -c` restores the user copy back to the original in `$CHROOT/root`, which wasn't updated priorly. Please add a comment reflecting this as it isn't obvious from the code.

AladW:
Thanks, merged: :+1:

Issue #156, opened 2016-07-22
make aria2c optional


AladW:
The issue with aria2c, besides the large codebase (comparable to curl) is that it takes much to compile from source (2+ GB RAM on my install). GNU parallel should be able to replace it, at least in aursearch.

AladW:
Via #164. aria2 and parallel are now both optional, relying on curl as fallback.

Issue #150, opened 2016-06-10
aursearch: by-desc and by-maintainer search


AladW:
Besides searching by name, it also possible to search by description and by maintainer. Supporting multiple arguments would also be useful, however this would likely need migration to jq (`jq add` and `jq unique_by`, see #113).

AladW:
As far as I can see, using multiple arguments with the search RPC would be hard without implementing #113, due to `jq add` and `jq unique_by`.

AladW:
Via #164

PR #164, opened 2016-09-19
#150 & #156

vorph1:
#150 & #156

### DIFF
--- a/bin/aursearch
+++ b/bin/aursearch
@@ -9,6 +9,7 @@ tmp=$(mktemp -dt "$argv0".XXXXXXXXXX) || exit
 readonly PS4 argv0 aurweb tmp
 format=short
 list=pkgname
+searchby=''
 
 # Split by 150 arguments to prevent HTTP 414.
 # https://bugs.archlinux.org/task/49089
@@ -29,7 +30,7 @@ split() {
 }
 
 getjson() {
-    if aria2c --download-result=hide --console-log-level=error --stderr=true -d "$tmp" -i -; then
+    if parallel --will-cite "curl -s -g {} -o $tmp/index-{#}.html"; then
         declare out=("$tmp"/index*.html*)
 
         if [[ -s ${out[0]} && $(jq '.resultcount' "${out[0]}") -gt 0 ]]; then
@@ -138,12 +139,16 @@ if [[ -t 2 ]]; then
     colorize
 fi
 
-while getopts :Fbrv OPT; do
+while getopts :Fbrvmnd OPT; do
     case $OPT in
         F) list=fixed   ;;
         b) list=pkgbase ;;
         r) format=none  ;;
         v) format=long  ;;
+        m|n|d) list=fixed ;;&
+        m) searchby=maintainer ;;
+        n) searchby=name ;;
+        d) searchby=name-desc ;;
     esac
 done
 shift $((OPTIND - 1))
@@ -154,6 +159,13 @@ if ((!$#)); then
     exit 1
 fi
 
-match "$list" "$@" | split | getjson | parse "$format"
+pkgs=("$@")
+
+if [ -n "$searchby" ]; then
+  arg="$@"
+  IFS=$'\n' pkgs=($(curl -s -G --data-urlencode "arg=$arg" -d "v=5" -d "type=search" -d "by=$searchby" "$aurweb/rpc/" | jq -r ".results[].Name"))
+fi
+
+match "$list" "${pkgs[@]}" | split | getjson | parse "$format"
 
 # vim: set et sw=4 sts=4 ft=sh:
--- a/bin/aursearch
+++ b/bin/aursearch
@@ -15,13 +15,14 @@ command_exists () {
     type "$1" &> /dev/null
 }
 
+urlencode() {
+    jq -R -r '@uri'
+}
+
 # Split by 150 arguments to prevent HTTP 414.
 # https://bugs.archlinux.org/task/49089
 split() {
     awk -v rpc="$aurweb/rpc/?v=5&type=info" '{
-        gsub("+","%2b")
-        gsub("@","%40")
-
         if (NR == 1)
             printf "%s&arg[]=%s", rpc, $0
         if (NR > 1)
@@ -39,13 +40,15 @@ getjson() {
     elif command_exists parallel; then
         parallel --will-cite -X "curl -s -g {}" > $tmp/out.json
     else
-        xargs curl -s -g > $tmp/out.json
+        readarray -t uri
+        curl -s  -g "${uri[@]}" > $tmp/out.json
     fi
 
     if [ $? -eq 0 ]; then
         sed -e '1i[' -e 's/}{/},{/g' -e '$a]' -i $tmp/out.json
 
-        declare -i totalresults=$(cat $tmp/out.json | jq '[.[].resultcount] | reduce .[] as $c (0; . + $c)')
+        declare -i totalresults
+        totalresults=$(cat $tmp/out.json | jq '[.[].resultcount] | reduce .[] as $c (0; . + $c)')
 
         if [[ $totalresults -gt 0 ]]; then
             cat $tmp/out.json
@@ -174,10 +177,11 @@ fi
 
 if [ -n "$searchby" ]; then
   { for q in "$@"; do
-    echo "$aurweb/rpc/?v=5&type=search&by=$searchby&arg=${q/ /%20}"
+    echo -n "$aurweb/rpc/?v=5&type=search&by=$searchby&arg="
+    echo "$q" | urlencode
   done } | getjson | parse "$format"
 else
-  match "$list" "$@" | split | getjson | parse "$format"
+  match "$list" "$@" | urlencode | split | getjson | parse "$format"
 fi
 
 # vim: set et sw=4 sts=4 ft=sh:
--- a/bin/aursearch
+++ b/bin/aursearch
@@ -30,7 +30,7 @@ split() {
 }
 
 getjson() {
-    if parallel --will-cite "curl -s -g {} -o $tmp/{#}.json"; then
+    if parallel -j$(( $(nproc) * 3 )) --will-cite "curl -s -g {} -o $tmp/{#}.json"; then
         declare out=("$tmp"/*.json)
 
         if [[ -s ${out[0]} && $(jq '.resultcount' "${out[0]}") -gt 0 ]]; then
--- a/bin/aursearch
+++ b/bin/aursearch
@@ -9,7 +9,7 @@ tmp=$(mktemp -dt "$argv0".XXXXXXXXXX) || exit
 readonly PS4 argv0 aurweb tmp
 format=short
 list=pkgname
-searchby=''
+searchby=name
 
 # Split by 150 arguments to prevent HTTP 414.
 # https://bugs.archlinux.org/task/49089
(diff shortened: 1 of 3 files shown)


AladW:
Nice, thanks for the work you're putting in!

A few notes:
+ It seems using parallel/curl is 4-6x slower than aria2, e.g.
```
[archie@vbox bin]$ time aursearch .+ | wc -l
75448

real    0m26.774s
user    0m5.940s
sys     0m2.470s
[archie@vbox bin]$ time ./aursearch .+ | wc -l
75448

real    2m45.164s
user    0m19.450s
sys     0m3.380s
[archie@vbox bin]$ time ./aursearch python | wc -l
6964

real    0m16.200s
user    0m2.030s
sys     0m0.390s
[archie@vbox bin]$ time aursearch python | wc -l
[DL:1.5MiB][#c7294b 0B/0B][#ba983b 0B/0B][#b7ab81 0B/0B][#44565d 0B/0B][#dffa13 0B/0B]
6964

real    0m3.255s
user    0m0.750s
sys     0m0.210s
```
Is this a limitation with parallel, or can it be improved on? In any case, good to see parallel + curl misses  no results compared to aria2.
+ I wonder if it makes sense to use split() for URL encoding, passing the search type as argument. It is an adhoc approach though, unlike `--data-urlencode`.
+ All arguments are concatenated to a single argument for bytype search. It would be nice if multiple targets return the result union, as with an info type search.

vorph1:
In my case curl was only 2x slower, i was able to fix this by increasing concurrency in parallel.

```
[szymon@desktop bin]$ time aursearch .+ | wc -l
[DL:2.5MiB][#463f94 0B/0B][#dea982 0B/0B][#8ad8fd 0B/0B][#e0b807 0B/0B][#597ce9 0B/0B]
[DL:2.2MiB][#a7326b 0B/0B][#329635 0B/0B][#1d2671 0B/0B][#7e76ac 0B/0B][#dd999a 0B/0B]
[DL:2.0MiB][#cc4087 0B/0B][#0a22e5 0B/0B][#c9ef60 0B/0B][#6ec320 0B/0B][#289034 0B/0B]
[DL:1.7MiB][#5d6ce5 0B/0B][#34ea86 0B/0B][#29bca6 0B/0B][#70b0e5 0B/0B][#7bff88 0B/0B]
[DL:1.5MiB][#29a81a 0B/0B][#f29cc6 0B/0B][#99b4fd 0B/0B][#316c38 0B/0B][#efca7b 0B/0B]
[DL:1.3MiB][#229271 0B/0B][#a7e395 0B/0B][#5175c7 0B/0B][#088863 0B/0B][#ec193a 0B/0B]
[DL:1.1MiB][#229271 0B/0B][#a7e395 0B/0B][#5175c7 0B/0B][#088863 0B/0B][#ec193a 0B/0B]
[DL:1.0MiB][#588f59 0B/0B][#2c3c70 0B/0B][#f9bced 0B/0B][#d3e674 0B/0B][#ee2961 0B/0B]
[DL:1.0MiB][#0ec2b3 0B/0B][#a46f7f 0B/0B][#dc7b4f 0B/0B][#a4ee01 0B/0B][#1d0270 0B/0B]
[DL:0.9MiB][#2bf9d6 0B/0B][#8c32df 0B/0B][#db9c00 0B/0B][#d89875 0B/0B][#bbc4cf 0B/0B]
[DL:0.9MiB][#a83702 0B/0B][#585f98 0B/0B][#fb1cab 0B/0B][#c962dd 0B/0B][#2166f7 0B/0B]
[DL:0.9MiB][#cbfa6b 0B/0B][#14d388 0B/0B][#95fc52 0B/0B][#7caab5 0B/0B][#15186d 0B/0B]
[DL:1.0MiB][#ea1abe 0B/0B][#6ea8d9 0B/0B][#bd47c9 0B/0B][#8d1115 0B/0B][#8c58ef 0B/0B]
[DL:1.1MiB][#eeb758 0B/0B][#956051 0B/0B][#58b067 0B/0B][#187626 0B/0B][#e1a051 0B/0B]
[DL:1.5MiB][#728ff4 0B/0B][#399183 0B/0B][#7d9fc6 0B/0B][#85137b 0B/0B][#58c52b 0B/0B]
75450

real	0m19.139s
user	0m2.490s
sys	0m0.717s
[szymon@desktop bin]$ time ./aursearch .+ | wc -l
75450

real	0m12.680s
user	0m5.867s
sys	0m1.370s
[szymon@desktop bin]$ time ./aursearch python | wc -l
6964

real	0m1.990s
user	0m0.737s
sys	0m0.143s
[szymon@desktop bin]$ time aursearch python | wc -l
6964

real	0m1.177s
user	0m0.350s
sys	0m0.063s
```

> All arguments are concatenated to a single argument for bytype search. It would be nice if multiple targets return the result union, as with an info type search.

I believe this should be the case for name-desc search.

> I wonder if it makes sense to use split() for URL encoding, passing the search type as argument. It is an adhoc approach though, unlike --data-urlencode.

split is needed mainly because of the 150 args limit right?

AladW:
>In my case curl was only 2x slower, i was able to fix this by increasing concurrency in parallel.

After some trial and error `-j24 --nice 19` reaches the same speed for me with parallel (on a VM with 1 CPU and 512MB RAM), more processes and RAM usage shoots up with no time benefit. 

However, CPU is at 100% while with aria2 it's at 7-10%. I think we could improve on this by using multiple arguments for each curl process (`parallel -X`), but I'm not sure how to construct the command line that way.

On a similar note, checking each file for results is probably the right thing to do, but perhaps this could be done in one go (`jq --slurp`?) rather than one process per file.

>I believe this should be the case for name-desc search.

Hmm, what's the reason for using "$*" with name-desc, and "$@" for the other types?

>split is needed mainly because of the 150 args limit right?

Yeah. The only real clean solution to this (and the need for parallel downloads) would be to either implement POST requests or extend the by-name interface, see:

https://bugs.archlinux.org/task/49089
https://bugs.archlinux.org/task/49099

I'm not sure how hard it is to write a patch for POST requests, but it's probably easier than regex support -- also because people are divided on how latter should be implemented, see [aur-dev](https://lists.archlinux.org/pipermail/aur-dev/2016-April/004027.html).

vorph1:
> I think we could improve on this by using multiple arguments for each curl process (parallel -X), but I'm not sure how to construct the command line that way.

This doesn't give us much since all urls will be downloaded sequentially. All thing considered I think aria is the superior solution.

> Hmm, what's the reason for using "$*" with name-desc, and "$@" for the other types?

Multi word searches make sense in description (i.e. media player) while maintainers and names are always one word. Come to think of it the user could just quote arguments with spaces, so that's up to you which way to go :smile:

AladW:
>This doesn't give us much since all urls will be downloaded sequentially. All thing considered I think aria is the superior solution.

I got some decent results using curl alone, provided all arguments are appended to one process. e.g. `aursearch python`, which has 23 files:
+ aria2: 1s
+ curl + readarray or xargs: 1.6s
+ curl + loop: 4s
The issue is that that curl seems to consider multiple arguments as one transfer, so adding newlines with `-w` will only append one at the end (preventing a successful merge).

Additionally, we could make name search the default, and leave regex support with its parallel downloads as an option (`-P`).

>Multi word searches make sense in description (i.e. media player) while maintainers and names are always one word. Come to think of it the user could just quote arguments with spaces, so that's up to you which way to go

Yeah, if you rely on quoting you can do e.g. `aursearch multi 'media player'` or `aursearch 'multi media' player`, where with one string it's `aursearch 'multi media player'` in both cases.

bin/aursync
```diff
@@ -55,13 +59,19 @@ download() {
 
     declare gz
 
-    parallel --will-cite -X "curl -s --remote-name-all {}"
+    if command_exists aria2c; then
+        aria2c -d "$1" -i -
+    elif command_exists parallel; then
+        parallel --will-cite -X "curl -s --remote-name-all {}"
+    else
+        xargs curl -s --remote-name-all
```
AladW: Use `readarray -t uri; curl <args> "${uri[@]}"` as xargs replaces curl's exit status with its own on failure

bin/aursync
```diff
@@ -55,13 +59,19 @@ download() {
 
     declare gz
 
-    parallel --will-cite -X "curl -s --remote-name-all {}"
+    if command_exists aria2c; then
```
AladW: I'm not too fond of the idea of supporting two methods for parallel downloading. If `-X` and increased jobs makes parallel as effective as aria2c I'd consider parallel, otherwise just stick to aria2c with curl as fallback.

bin/aursync
```diff
@@ -55,13 +59,19 @@ download() {
...
+    elif command_exists parallel; then
+        parallel --will-cite -X "curl -s --remote-name-all {}"
+    else
+        xargs curl -s --remote-name-all
+    fi
+
+    cd $OLDPWD
 
-    for gz in *.tar.gz; do
+    for gz in $1/*.tar.gz; do
```
AladW: You could use a subshell to save the second cd

bin/aursearch
```diff
@@ -30,16 +30,13 @@ split() {
 }
 
 getjson() {
```
AladW: I'm not sure why but I'm missing some results in this section.
```
[archie@vbox bin]$ time ./aursearch python | wc -l
6926
[archie@vbox bin]$ time aursearch python | wc -l
[#37eae8 0B/0B CN:1 DL:0B]
6972
```
Possibly a newline is missing somewhere; the AUR does not add newlines by itself. That's why `awk '{print}' was used instead of cat; `-w \n` may also be necessary for curl.

bin/aursearch
```diff
@@ -164,15 +160,12 @@ if ((!$#)); then
...
-  "name"|"maintainer"|"name-desc")
-    { for u in "${urls[@]}"; do
-      echo "$aurweb/rpc/?v=5&type=search&by=$searchby&arg=${u/ /%20}"
-    done } | getjson | parse "$format"
-  ;;
-  *) match "$list" "$@" | split | getjson | parse "$format"
-esac
+if [ -n "$searchby" ]; then
+  { for q in "$@"; do
+    echo "$aurweb/rpc/?v=5&type=search&by=$searchby&arg=${q/ /%20}"
```
AladW: Since this bypasses the split function, it doesn't URL encode `@` and `+`. A better solution may be to generalize `split` to take arguments, and add URL encoding for the space character. e.g.
```
    declare type=$1 by=$2
    declare rpc="$aurweb/rpc/?v=5&type=$type"

    if [[ -n $by ]] && [[ $by != none ]]; then
        rpc+="&by=$by"
    fi

(...)
        gsub(" ","%20")
```

man1/aursearch.1
```diff
@@ -4,25 +4,30 @@ aursearch \- search AUR package names by pcre pattern
...
-.I pattern
+.I string
 .YS
 
 .SH DESCRIPTION
-\fBaursearch\fR searches the AUR by a PCRE pattern. If the query
-contains multiple patterns, the union of all pattern results is
-returned.
+\fBaursearch\fR searches the AUR by multiple patterns or strings,
+the union of all search results is returned.
```
AladW: Needs some conjunction or simply two sentences instead of a comma

bin/aursearch
vorph1: I added the -P flag and set the default to use by=name search, that's why the two differ.

AladW: Oh, right 👍

bin/aursearch
vorph1: I'd suggest extracting urlencoding from split as another step in the pipeline. The search method in the rpc doesn't support passing an array of arguments so each search term requires a separate request.

AladW: Sounds good to me.

bin/aursearch
```diff
@@ -15,13 +15,14 @@ command_exists () {
     type "$1" &> /dev/null
 }
 
+urlencode() {
+    jq -R -r '@uri'
```
AladW: Huh, didn't know jq could do that. Nice

bin/aursearch
```diff
@@ -39,13 +40,15 @@ getjson() {
...
+        readarray -t uri
+        curl -s  -g "${uri[@]}" > $tmp/out.json
     fi
 
     if [ $? -eq 0 ]; then
         sed -e '1i[' -e 's/}{/},{/g' -e '$a]' -i $tmp/out.json
 
-        declare -i totalresults=$(cat $tmp/out.json | jq '[.[].resultcount] | reduce .[] as $c (0; . + $c)')
+        declare -i totalresults
+        totalresults=$(cat $tmp/out.json | jq '[.[].resultcount] | reduce .[] as $c (0; . + $c)')
```
AladW: Useless cat

AladW:
Comparing `parallel -X -j +8` to aria for all results:
```
$ cat parallel_stat.txt # 17 samples
mean(11.750)    median(11.750)  sstdev(11.750)
14.60925           15.037                2.6846448927186

$ cat aria2_stat.txt # 17 samples
mean(16.411)    median(16.411)  sstdev(16.411)
16.0116875       15.4095              3.0986248072061
```
So surprisingly, parallel does better than aria2c. CPU and RAM usage are similar in both cases.

vorph1:
Why not just keep them both as optional deps? Your call 😄

AladW:
Fair enough I guess. In any case, merged with a few changes. Note the aursearch flags might still change before the next release. Cheers

PR #167, opened 2016-09-24
Fix aurbuild fail on creating a new container

vorph1:
This fixes 2 bugs that caused aurbuild to fail when creating a new container (introduced in 1.3.3)


AladW:
Thanks. We should probably go back to `pacconf --raw` for a more complete fix.

AladW:
Done with 92b6b56b4e522b01cbf76d2f5b72f70045d52d59

Issue #185, opened 2016-10-27
Feature proposal: Pinning of the AUR TLS certificate


clawoflight:
Since the AUR doesn't and probably never will sign packages, it would be a good and fairly straightforward thing to pin it's TLS certificate. That way, we are at least fully safe against MITM attacks.

This could be used in `curl` and `git`, `aria2` doesn't support it.

I propose adding the following features:
- an environment variable along the lines of `AURUTILS_PINNED_AUR_PUBKEY`
- using it to populate `curl --pinnedpubkey` if set
- using it to populate `git clone --config http.pinnedpubkey` if set
- never using `aria2` if it is set

And documenting a `find` command which can be used to set or update the value of `http.pinnedpubkey` for all cloned build repositories in `~/.cache/aursync`.

I am willing to implement this once I get guidelines and a green light.
Do others want this? How exactly do you want to see it implemented?

AladW:
This doesn't look too complex to implement, if you use an option array for curl (c.f. `makepkg_args` in `aurbuild`). Feel free to make a pull request with what you have in mind, and I'll comment from there. :+1:

AladW:
Related: https://lists.archlinux.org/pipermail/pacman-dev/2016-October/021639.html

clawoflight:
I'll probably make an option array, that seems like the most elegant solution.
I will add an explicit warning when I document that :)

> Related: https://lists.archlinux.org/pipermail/pacman-dev/2016-October/021639.html

Cool, it will be nice to have that in pacman, though the signed packages make it less urgent :)

AladW:
I've asked in #archlinux-security and it seems the AUR doesn't support HPKP in the first place... compare
```
[archie@thinkpad ~]$ curl -I https://aur.archlinux.org
HTTP/1.1 200 OK
Server: nginx/1.10.1
Date: Tue, 01 Nov 2016 19:13:54 GMT
Content-Type: text/html; charset=utf-8
Connection: keep-alive
Vary: Accept-Encoding
Cache-Control: no-cache, must-revalidate
Expires: Tue, 11 Oct 1988 22:00:00 GMT
Pragma: no-cache
Strict-Transport-Security: max-age=16070400
```
with github:
```
[archie@thinkpad ~]$ curl -I https://github.com
HTTP/1.1 200 OK
Server: GitHub.com
Date: Tue, 01 Nov 2016 19:16:38 GMT
Content-Type: text/html; charset=utf-8
Status: 200 OK
[...]

PR #189, opened 2016-11-01
Pinning of the AUR's TLS public key

clawoflight:
This is WIP towards implementing #185.

- [x] Patch aursync
- [ ] Patch aursearch
- [ ] Patch aurgrep
- [ ] Document

### DIFF
--- a/bin/aursync
+++ b/bin/aursync
@@ -15,6 +15,10 @@ readonly PS4 argv0 aurweb tmp XDG_CACHE_HOME AURDEST
 declare -i view=1 build=1 chkver=1 fetch=1
 declare -i update=0 snapshot=0 sign=0 rotate=0 chroot=0
 declare -i bindro=0 force=0 keep=0 logdest=0 namcap=0 chtemp=0
+declare -i pubkeypin=0
+
+curl_args=(--cert-status -fLsS --remote-name-all)
+git_clone_args=()
 
 # stdin: /path/to/repo.git
 clone() {
@@ -34,6 +38,13 @@ clone() {
                 warning "remote $remote does not match $repo"
             fi
 
+            # Update the config to get the newest pubkey. Is this smart or not efficient enough?
+            if ((pubkeypin)); then
+                git config http.pinnedpubkey "$AURUTILS_PINNED_AUR_TLS_PUBKEY"
+            else
+                git config --unset http.pinnedpubkey
+            fi
+
             git fetch -v # returns 0 even if nothing was fetched
 
             if [[ $(git rev-parse HEAD) != $(git rev-parse '@{upstream}') ]]; then
@@ -44,7 +55,7 @@ clone() {
 
             cd_safe "$startdir"
         else
-            git clone "$repo"
+            git clone "${git_clone_args[@]}" "$repo"
         fi
     done
 }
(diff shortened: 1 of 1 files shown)


bin/aursync
```diff
@@ -53,12 +64,12 @@ download() {
     declare gzdir=$1 startdir=$PWD
     cd_safe "$gzdir"
 
-    if type -P aria2c >/dev/null 2>&1; then
+    if ((pubkeypin==0)) && type -P aria2c >/dev/null 2>&1; then
         aria2c --stderr=true -i -
     elif type -P parallel >/dev/null 2>&1; then
-        parallel --will-cite -X -j +5 --nice 10 --halt soon,fail=1 "curl -fLsS --remote-name-all {}"
+        parallel --will-cite -X -j +5 --nice 10 --halt soon,fail=1 "curl ${curl_args[*]} {}"
```
clawoflight: I am unsure whether this is the correct expansion to use.
Shellcheck gave me an error when I used `[@]` here.

bin/aursync
```diff
@@ -33,6 +37,13 @@ clone() {
                 warning "remote $remote does not match $repo"
             fi
 
+            # Update the config to get the newest pubkey. Is this smart or not efficient enough?
```
clawoflight: I thought about this while writing the other stuff.

`time` tells me the performance impact is negligible, but that is on a fast machine and you or others might disagree.

bin/aursync
AladW: [*] is the right way; e.g. try `foo=(a b c); printf '%s\n' "foo: ${foo[@]}" && printf '%s\n' "${foo[*]}"`. You can find a more detailed description in the Shellcheck wiki

bin/aursync
AladW: It's something like 0.003s for me... should be negligable compared to the other stuff.

bin/aursync
clawoflight: Thanks, that example cleared it up :)

bin/aursync
clawoflight: Cool. That removes the need for a way to manually update the hash.

clawoflight:
Should I start documenting this, or are you waiting for something? :)

AladW:
It's more that I'm hesitant knowing there's no upstream support for actual HPKP, even when client-side pinning works fine. If you could do `--cert-status` in a separate commit, I'd merge that one directly, though.

clawoflight:
I could do that, but it's not really worth a PR. You can add that to the two calls two `curl` if you want :)

Do you have a specific reason to hesitate? :)
I actually prefer client-side pinning because it's an out-of-band trust model, whereas everything is sent along with HPKP.

clawoflight:
Cool, github kept the comments after a rebase and force-push :)

AladW:
My concern for client-side pinning when HPKP is not supported by the site is, how do you make sure you don't have a fake, MITM certificate at the time you set the public key in your configuration? (I suppose you have a similar issue with HPKP, but only on the very first connection to the website)

clawoflight:
> how do you make sure you don't have a fake MITM certificate at the time you set the public key in your configuration? 

By thinking and being careful. For example, you could
* check several times during the week, and/or
* check from within different networks, e.g. uni, and/or
* use different DNS servers, and/or
* also check through a VPN, and/or
* check after SSHing somewhere else.

Maybe the admin of the machine the AUR runs on would even GPG-sign a hash generated directly from the file and/or post it to aur-general if you ask nicely :)

I don't find this inherently less secure than HPKP, but rather a trade-off. HPKP is nice because it requires no user-interaction, but a MITM on first connection negates it completely.
Whereas this makes out-of-channel checks much easier, and even allows you to *secure a new machine before ever connecting to the internet*, simply by copying a trusted hash over from another one.
Or, for that matter, pass a hash on to close friends, or compare against theirs.

clawoflight:
For reference, I have this in my `~/.zshrc`:

```zsh
function generate-tls-pubkey-hash() {
	openssl s_client -servername $1 -connect $1:443 |
	 openssl x509 -pubkey -noout |
	 tee /dev/stderr |
	 openssl rsa -pubin -outform der |
	 openssl dgst -sha256 -binary |
	 openssl enc -base64
}
```

The resulting hash can be used in curl by prepending `sha256//`.

clawoflight:
Did you make up your mind on this? :)

AladW:
Sorry, I don't have much time to look at aurutils atm

clawoflight:
Okay, thanks for the reply, just wanted to get a lifesign :)

AladW:
See https://github.com/AladW/aurutils/issues/185#issuecomment-326458202. If you want to implement this in a separate repo, let me know and I'll link it from the readme.

PR #204, opened 2016-12-28
aursync: fix order of arguments passed to aurcheck

djmattyg007:


djmattyg007:
I found that when passed after the non-option arguments, the `-r` argument was ignored by `getopts`.

AladW:
Perhaps it's better to use an explicit argument for the database in `aurcheck`, similar to `aurbuild -d`, instead of relying on an exact order?

djmattyg007:
I'm comfortable with that.

AladW:
Thanks!

PR #221, opened 2017-03-07
Fixed aurcheck manpage

Foxboron:
The doc for -c is probably not good enough. Didn't get anything out of it when trying to see how it works.

Signed-off-by: Morten Linderud <morten@linderud.pw>


AladW:
For an example usage, see https://github.com/AladW/aurutils/blob/master/bin/aursync#L189

Issue #237, opened 2017-04-28
--ignore flag doesn't properly read input


Flat:
Using --ignore with aursync -u returns `grep: ignore.txt: No such file or directory`, both when specifying a file with package names, as well as entering a package name directly.

```
cat ignore.txt
synergy-git

```

[error.txt](https://github.com/AladW/aurutils/files/963172/error.txt)

AladW:
Please try https://github.com/AladW/aurutils/commit/34f92cb425b2fc9c8a4e7320f0d6d9e804ff2eda

Flat:
No error, but it fails to ignore the package.

[error.txt](https://github.com/AladW/aurutils/files/965253/error.txt)
```
cat ignore.txt 
keybase-bin
```

AladW:
Is the ignored package actually built or does it only display the `-> keybase-bin 1.0.21_20170424190450+5b1f576-1 -> 1.0.21_20170428160949+36e780b-1` line in the console?

Flat:
Sorry, looks like it is actually ignored, it's just displayed in the input/output of aurchain. So if that is intentional behavior then https://github.com/AladW/aurutils/commit/34f92cb425b2fc9c8a4e7320f0d6d9e804ff2eda fixes the issue. Thanks!

AladW:
Well, there should probably be some indication that a package is ignored, besides the fact that it is not built. I'll take a look before the next release.

AladW:
@Flat please try https://github.com/AladW/aurutils/commit/86b2dc9fe25818b9afb192569185b50e1d7880b3

Flat:
Looks good to me! Any thoughts on passing packages to ignore on the command line without specifying a file?

AladW:
https://github.com/AladW/aurutils/commit/294b989dbb559eb761502bfd373ed56e17fae411

Issue #208, opened 2017-01-09
feature: aursync flag for deleting package from db and/or purge


prietus:
appart from e.g repose -d /var/cache/pacman/custom/custom.db package there is no other way of doing it (I think)
maybe is a good idea having a remove/purge flag. 

cheers.

AladW:
https://github.com/vodik/repose/issues/29#issuecomment-270982941

vodik:
Are we talking about purging packages and their dependencies, or something else?

AladW:
@prietus bump to question above

AladW:
Closing due to inactivity

daurnimator:
Please reopen.

> Are we talking about purging packages and their dependencies, or something else?

e.g.
  - I build a package I though was interesting: `aursync foo`
  - I install it `sudo pacman -S foo`
  - I use it for a little bit, but get bored of it and want to delete it: `sudo pacman -Rs foo`
  - At some point later the package `foo` gets an update in AUR, infact it gets to be quite a large package.
  - I run `aursync -u` to update all my packages installed from AUR. however `foo` now takes a long time to build: I don't use it any more, and want to drop it.

==> Please add a command that allows me to easily delete/purge 'foo', or at least stop it from getting updated by `aursync -u`.

AladW:
Note that aurbuild now uses repo-add directly, so you can use a tool like repoctl to automatically remove packages you no longer need from the database. To ignore packages, create a textfile (e.g. `~/.config/aurutils/ignore.conf` and point `aursync` with `aursync --ignore ~/.config/aurutils/ignore.conf`. If either of this is not clear from the documentation, please open a new issue for that.

aphirst:
I second the idea that aurutils itself being able to commission removing packages from its own custom repository would be a very useful time/effort-saving feature.

maximbaz:
I think having such a script would be very useful, and I'd like to help out with it. The workaround `repo-remove` works, but I think it doesn't account for dependencies, so you have to manually keep track of all unused dependencies and run `repo-remove` for all of them too.

@AladW if you agree, you can reopen the issue and assign to me. It's going in the end of my list though, first I want to finish with VCS packages (#283), then propose an implementation of repo autodetection for all existing scrips (so that you don't have to pass `-d myrepo` if you only have one repo, just like `aursync` already works), then make sorting in `aursearch` configurable (#279), and then take a stab at this 🙂

[...]

Issue #20, opened 2016-03-04
aurweb RPC interface


AladW:
This issue tracks aurweb changes.

rmarquis:
You can add to your list that the RPC currently includes all architecture dependent (make)depends in the same arrays. See [here](https://lists.archlinux.org/pipermail/aur-dev/2014-December/003006.html) for details.

AladW:
Added, thanks for the reminder.

AladW:
Upstream issues for `repose`:
- [x] **repose** performance issues with large databases https://github.com/vodik/repose/issues/42
- [ ] **repose** does not take absolute paths to packages https://github.com/vodik/repose/issues/49
- [ ] **repose** failure on database signing https://github.com/vodik/repose/issues/46
- [ ] **repose** parsing errors
- [ ] **repose** segmentation faults

Removed from the first post as support will be removed from `aurbuild`, see #238.

eli-schwartz:
https://bugs.archlinux.org/task/49946 fixed in git

Error codes for all in https://lists.archlinux.org/pipermail/pacman-dev/2017-September/022156.html
Currently in the patchqueue at https://git.archlinux.org/users/allan/pacman.git/commit/?h=patchqueue&id=ab2be5794de8a57426ec3fc0f631cc766a0b8227

eli-schwartz:
> pacman vercmp alpha '' results in -1 (alpha < "") [FS#49093](https://bugs.archlinux.org/task/49093)

Submitted fix at https://lists.archlinux.org/pipermail/pacman-dev/2017-December/022211.html

eli-schwartz:
>  **pacman** `vercmp alpha ''` results in `-1` (`alpha < ""`) [FS#49093](https://bugs.archlinux.org/task/49093)

What's the dead end here? `vercmp alpha ` is now a fatal error in the latest stable release. I don't think it's strictly necessary to call vercmp with quoted arguments, as any actual version in the wild, with a space in it, is invalid and can be checked as such -- vercmp doesn't do any checks at all to determine whether a pkgver is invalid or not.

From the bug report:
> Well, vercmp is useful in other cases, such as comparing sync/local versions (with pacman --print or expac) against the AUR RPC (with curl/json). If either string is empty, you'd have to be aware of the limitation, and workaround by e.g. setting -1 yourself if the variable is empty.

[...]

PR #238, opened 2017-04-28
aurbuild: remove repose support

AladW:
I haven't used `repose` in some time and it adds a maintenance burden in the form of sporadic bugs and user requests (see #237, https://bbs.archlinux.org/viewtopic.php?pid=1706372#p1706372, #208, #20). Drop it and only leave support for `repo-add`. Existing database may have to be rebuilt accordingly.


vodik:
Shame, but not surprising.

I'm not dead though, and going to be getting myself back on the archlinux irc channels and working on repose again this week.

Earnestly:
@vodik I still use it :P  Nothing else is quite as fast and as simple, even if it might get confused about files sometimes, I simply rebuild the repo.

AladW:
Hi vodik, glad to see you're still alive! :D This gives me a good excuse to delay the 1.6 release indefinitely then. (1.5.3 still supports repose)

PR #243, opened 2017-05-15
aurqueue: add -h option handling

alyptik:


AladW:
I had done something similar before seeing this pull request... (see https://github.com/AladW/aurutils/commit/ef6aa2eb125b5a1f2afc0b292dc2a5e0cec8408e on option-handling branch) 

Regarding https://github.com/AladW/aurutils/pull/243/files#diff-b21534d75a3a443dac260f119611eba6R65, is this to handle the case where someone made a file or folder called `--help`?

alyptik:
> Regarding https://github.com/AladW/aurutils/pull/243/files#diff-b21534d75a3a443dac260f119611eba6R65, is this to handle the case where someone made a file or folder called --help?

Yup, exactly that. I figured it should be handled as an "it could happen" edge case, as you had mentioned earlier that `-f oobar` is a valid repository name.

I am not very familiar with **getopts**; does it allow disambiguations like that? Would it be preferable to rebase and submit PR for that branch?

AladW:
>Yup, exactly that. I figured would be useful to test for as an "it could happen" edge case, as you had mentioned earlier that -f oobar is a valid repository name.

Hmm, technically the only arguments you'd pass on to aurqueue would be package names, and those (unlike repositories) can't start in a hyphen. 

>I am not very familiar with getopts; does it allow disambiguations like that? Would it be preferable to rebase, make changes, and submit PR for that branch?

Master is fine. As to getopts, you can pass it `--` so it doesn`t treat arguments starting in a hyphen as option arguments, e.g. you can do `aurqueue -- *`. The benefit is that it's easily extensible if you want more options later (e.g. `-abc` or `-a optarg -b optarg`.

alyptik:
Integrated your changes from https://github.com/AladW/aurutils/commit/ef6aa2eb125b5a1f2afc0b292dc2a5e0cec8408e as suggested, since I could see no better way to handle usage flags.

I did however also add the ability to read from `standard input` when `aurqueue` is part of a pipe, similar to `pacman`; the script will also to do so when  **-** is part of the path list.

Afterward, the two path lists are merged so the rest of the script can proceed as before. I have updated the PR title accordingly.

AladW:
:+1:

Issue #217, opened 2017-02-22
man1: updates for 2.x


AladW:
## Man pages
- [x]  `aur(1)`: explain how to add custom scripts (#305, #317) -> #540
  + [x] How to keep the size of `AURDEST` small (#317)
  + [x] Remove packages from a local repository (#305)
- [x] `aur-build(1)`: `/var/tmp` PKGDEST, default `makepkg` arguments
- [x] `aur-build(1)`: `GPGKEY` in `makepkg.conf` is ignored for `repo-add`
- [x] `aur-vercmp(1)`: explain `-u` (https://github.com/AladW/aurutils/commit/dd4c302997c5eaee752a54c974c237a2b952d5c8), `-c`
- [x] `aur-repo(1)`: add option descriptions
- [x] `aur-vercmp(1)`: add `--quiet`/`-q`
- [x] `aur-sync(1)`: add `--no-graph`
- [x] `aur-depends`: now takes input as arguments
- [x] `aur-fetch`: now takes input as arguments

## Procedures
- [x] Add information on _hosting_ a custom repository (https://github.com/AladW/aurutils/issues/247#issuecomment-304256312)
- [x] Pointers on general AUR usage ([wiki](https://wiki.archlinux.org/index.php/Arch_User_Repository))

Foxboron:
I made a overview of disparities inn documentation and missing options. Should be a complete list of all the documentations issues.


#### aurbuild:
- [ ] help string difference from manpage
- [ ] man synopsis missing options

#### aurchain:
- [ ] help string difference from man page
- [ ] missing zsh completion

#### aurcheck:
- [ ] missing zsh completion

#### aurfetch:
- [ ] missing man page
- [ ] missing zsh completion

#### aurgrep:
- [ ] missing man page
- [ ] missing help string
- [ ] missing zsh completion

#### aurqueue:
- [ ] missing help string

#### aursearch:
- [ ] help string difference from man page
- [ ] man synopsis missing options
- [ ] man page missing -n option

#### aursync:
- [ ] help string difference from man page
- [ ] missing flag documentation:
  - [ ] -B
    - [ ] Missing inn the zsh completion
    - [ ] The manpage and help says two different things.
  - [ ] -C
    - [ ] Missing inside the manpage synopsis
- [ ] makepkg and makechrootpkg missing from zsh completion

AladW:
Awesome, thanks! I've copied the list to the first post.

AladW:
[...]

PR #258, opened 2017-09-23
aurbuild_chroot documentation

jebrosen:
Completes a few tasks in #217. Feedback appreciated before working on more documentation in the future.

### DIFF
--- a/bin/aurbuild_chroot
+++ b/bin/aurbuild_chroot
@@ -33,7 +33,7 @@ trap_exit() {
 }
 
 usage() {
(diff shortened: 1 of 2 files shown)


AladW:
Thanks! The only remark I'd have on this request is that pacconf is also used in aurbuild, so the reference in the See also section could stay. Arguably the same for makepkg.conf/pacman.conf, but the relevant options (-M, -C, -D) are not documented yet in aurbuild(1).

jebrosen:
I reverted the changes to aurbuild(1) for this PR; I think they should be tackled separately.

AladW:
Cheers :+1:

PR #271, opened 2018-01-10
aurgrep: support pattern beginning with hyphen

abouvier:
Example:
```
$ aurgrep -- -nox
vim-nox
cppcheck-nox
erlang-nox-r17
vlc-nox
gst-plugins-bad-rpi-nox
usbguard-nox
qbittorrent-nox-git
virtualbox-guest-utils-nox-svn
emacs-nox-24bit
sdl2-nox
```

### DIFF
--- a/bin/aurgrep
+++ b/bin/aurgrep
@@ -52,6 +52,6 @@ else
     list_update "$aurweb/$pkglist.gz" "$pkglist" > timestamp
 fi
 
-pcregrep "$@" "$pkglist"
+pcregrep -e "$@" "$pkglist"
 
 # vim: set et sw=4 sts=4 ft=sh:


AladW:
Not sure why `-e` would be required in this case - can you post the error message where it fails?

abouvier:
The man of `pcregrep` for option `-e` says:
> Specify a pattern to be matched. This option can be used multiple times in order to specify several patterns. **It can also be used as a way of specifying a single pattern that starts with a hyphen.**

Without this option the command `aurgrep -- -nox` blocks indefinitely.

AladW:
I expected this might cause some issues with other arguments, which it did (see #285)
```
$ aurgrep libstd++-doc
$ aurgrep -- -F 'libstd++-doc'
pcregrep: Failed to open libstd++-doc: No such file or directory
```
That's why I opted to simply print the list and leave `grep` to the user. Thanks for the feedback.

PR #278, opened 2018-01-26
Propagate -N (--nosync, --no-sync) aurbuild option to aursync

maximbaz:
I'd like to propagate the `-N` option from aurbuild to aursync for the following reason:

After I review the package files, I want aursync to be able to fully run without my interaction until the very end. Right now it stops on the first `sudo pacman -Sy` after it built the first package and waits for sudo password, which is annoying. I am synchronizing repositories anyway when I'm upgrading (by running `pacman -Syu`), so it's totally fine to not synchronize the repo during `aursync`.

### DIFF
--- a/bin/aursync
+++ b/bin/aursync
@@ -74,8 +74,8 @@ fi
 
 longopts=allan,bind:,chroot,directory:,force,ignore:,ignorearch,ignore-arch,log
 longopts+=,noconfirm,no-confirm,nover,no-ver,noview,no-view,print,rmdeps,rm-deps
-longopts+=,sign,temp,tar,repo:,root:,update,aurutils,continue
-shortopts=B:C:D:M:AcfLnprstTu
+longopts+=,nosync,no-sync,sign,temp,tar,repo:,root:,update,aurutils,continue
+shortopts=B:C:D:M:AcfLnNprstTu
 
 if TEMP=$(getopt -o "$shortopts" -l "$longopts" -n "$argv0" -- "$@"); then
     eval set -- "$TEMP"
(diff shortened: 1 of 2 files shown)


AladW:
Ok, but what do you do when building a package with AUR dependencies? For `makepkg -s` to work you would need the previously built packages to be propagated to pacman.

maximbaz:
Ah, so that's what would fail... I see now. I didn't run into such example while testing.

I was also considering to allow `pacman -Sy` in sudoers to be run without password, but aurbuild doesn't run exactly this command, it runs two commands with always changing arguments, which is difficult to whitelist. I don't think that allowing access to the entire pacman without sudo password is a great idea... 

Do you perhaps have some ideas how else to achieve what I want, completely independent process that I don't have to constantly monitor?

AladW:
The simplest way would be to use chroot builds and simply whitelist `aurbuild_chroot`. For regular builds I'm not sure on a satisfactory solution since makepkg has sudo calls to pacman as well.

maximbaz:
I just tried out your idea with whitelisting `aurbuild_chroot`, this works very well. I wonder how much it is of a security hole 😃. However it is definitely better than whitelisting the entire `pacman`, and in general it looks cleaner to build AUR packages in `chroot`.

Thanks for the idea, I'll close this PR since it looks like I have a better approach now.

djmattyg007:
Is there a reason to continue supporting non-chroot builds?

AladW:
Is there a reason not to? The vast majority of AUR packages fail to build in a chroot, and while using devtools in AUR is a good practice it is not mandated.

djmattyg007:
I was not aware of that. Ignore me, carry on.

Issue #65, opened 2016-03-21
aurcheck: retrieve new pkgver for VCS packages


AladW:
Old approach: https://github.com/AladW/aurutils/blob/1eb2271bae5b87c582104e81f52a19d642b401ce/aursync#L52

Which was hacked, as you have to install packages outside `makepkg -s` and hardcode VCS. Running pkgver() a second time isn't very nice either.

rmarquis:
Note that you can technically rely on the VCSCLIENTS array of makepkg.conf if you want to avoid hardcoding VCS clients.

AladW:
Finding VCS packages in a given list could have its place in aurqueue.

AladW:
I've left an attempt in the experimental branch for now. For the needed glue with other scripts, see #79.

AladW:
@vodik's script:

https://ptpb.pw/qBwa/bash

vodik:
I think @holomorph did some work on this stuff too, if you poke him nicely on IRC.

holomorph:
I looked into it a bit, but I think @Earnestly figured this out more thoroughly. I never committed my rebuild script to any repo and don't have access to that machine atm

Earnestly:
Ah but I only support git (for now) and don't support (or rely, however you want to look at it) on the PKGBUILDs: https://github.com/Earnestly/pkgbuilds/blob/master/util.bash

It's allows me to keep it -very- simple because I don't exactly need that level of dynamic convenience which is necessary for a project like this.

AladW:
For more suggested workarounds, see #227.

AladW:
For an interesting approach which uses the github API (github.com sources only), see:

https://github.com/Jguer/yay/blob/a6a0b9b441e3bdc67a9ac9f86c2961956531c31b/vcs.go

maximbaz:
@AladW I noticed this issue has an `upstream` label, I assume it is referring to [FS#56602](https://bugs.archlinux.org/task/56602)? If so, the consensus seems to be to use naming conventions for packages.

But also, I noticed that the discussion there is about how to detect which packages are VCS and VCS-like, not about how to detect if a package actually has an update. To me it sounds that the more complicated part is actually invoking `pkgver()` and detecting whether a given package needs to be updated or not, and it isn't as much of a big a deal to decide _when_ to run this check - we could start with the most simple implementation possible (match package name against hardcoded list of suffixes?) and later improve.

[...]

PR #283, opened 2018-01-27
aur-sync: Implement `--devel` option to also update VCS packages

maximbaz:
This is a rough version that adds support for VCS packages. 

I want to get your review first before I begin to improve the code, please leave comments on the general approach as well as implementation details. I'll also leave some comments inline.

Related: #65

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -6,6 +6,7 @@ readonly argv0=sync
 readonly XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
 readonly AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/$argv0}
 readonly AURDEST_SNAPSHOT=${AURDEST_SNAPSHOT:-$XDG_CACHE_HOME/aurutils/snapshot}
+readonly AURVCS=${AURVCS:-.*-(bzr|git|hg|svn)$}
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
 # default arguments
@@ -19,7 +20,7 @@ makepkg_args=(-cs)
 build=1 chkver=1 download=1 view=1
 
 # default options (disabled)
-chroot=0 list=0 rotate=0 snapshot=0 update=0 provides=0 repo_list=0
+chroot=0 list=0 rotate=0 snapshot=0 update=0 provides=0 repo_list=0 devel=0
 
 viewer() {
     if type -P vifm >/dev/null; then
@@ -107,7 +108,7 @@ fi
 longopts=allan,bind:,chroot,directory:,force,ignore:,ignorearch,ignore-arch,log
 longopts+=,noconfirm,no-confirm,nover,no-ver,noview,no-view,print,rmdeps,rm-deps
 longopts+=,sign,temp,tar,repo:,database:,root:,upgrades,nhehgvyf,continue,list,git
-longopts+=,makepkg-conf,pacman-conf,provides,repo-list
+longopts+=,makepkg-conf,pacman-conf,provides,repo-list,devel
 shortopts=d:D:AcfglLnprstTu
 
 if optstring=$(getopt -o "$shortopts" -l "$longopts" -n "$argv0" -- "$@"); then
@@ -178,6 +179,9 @@ while true; do
             shift ;;
         -u|--upgrades)
             update=1
+            shift ;;
+        --devel)
+            devel=1
             shift ;;
         --allan|--nhehgvyf)
             rotate=1
@@ -269,6 +273,10 @@ mkdir patch
 
 bsdcat "$root/$repo".db | db_namever | db_fill_empty >db_info
 
+if ((devel)); then
+  cut -f1 db_info | grep -E "$AURVCS" | xargs -I{} find "$aur_workdir" -maxdepth 1 -type d -name {} | xargs -r aur srcver >vcs_info || true
+fi
+
 if ((list)); then
     column -JN pkgname,pkgver -n "$repo" db_info
     exit
@@ -280,6 +288,12 @@ fi
 
   if ((update)); then
       aur vercmp <db_info | cut -d: -f1
+      if ((devel)); then
+        # At this point, mark **all** VCS packages for a potential update:
+        # - if `chkver` is used, all up-to-date packages will be filtered out, and user will only be presented with obsolete packages
+        # - if `chkver` is not used, we will not check versions and present all VCS packages for an update
+        cut -f1 vcs_info
+      fi
   fi
 } >argv
 
@@ -295,17 +309,32 @@ fi
       printf -- '%s\n' "${pkg_i[@]}"
   fi
 
-  if ((chkver)); then
-      cut -f1,3 pkginfo | aur vercmp -p db_info -c
-  fi
-
   if ((provides)); then
       # note: this uses pacman's copy of the repo (as used by makepkg -s)
       cut -f1 pkginfo | complement argv | aur rfilter -d "$repo"
   fi
 } >filter
 
-cut -f1 pkginfo | lib32 | complement filter >queue_0
+{ if ((chkver)); then
+      # 1. Unconditionally sync new packages (both regular and vcs) that are not in the repo yet (not in `db_info`)
+      cut -f1 pkginfo | grep -Fxvf <(cut -f1 db_info) || true
+
+      # 2. Sync packages (both regular and vcs) that are already in the repo, that are outdated according to AUR RPC
+      #    (whose current versions in `db_info` are older than versions in `pkginfo`)
+      #
+      #    vcs packages can be marked here as outdated if maintainer updates the static pkgver in PKGBUILD.
+      #    We expect that maintainers will do that for any major changes in PKGBUILD (such as changing source or build commands).
+      aur vercmp -p <(cut -f1,3 pkginfo) <db_info | cut -d: -f1
+
+      # 3. Sync vcs packages that are already in the repo, that are outdated according to aur-srcver
+      #    (whose current versions in `db_info` are older than versions in `vcs_info` - not in `pkginfo`)
+      ((devel)) && aur vercmp -p vcs_info <db_info | cut -d: -f1
+  else
+      cut -f1 pkginfo
+  fi
+} >pkgs_to_sync
+
+lib32 <pkgs_to_sync | complement filter >queue_0
 
 if [[ -s queue_0 ]]; then
     select_pkgbase queue_0 pkginfo | cut -f2 >queue


bin/aurcheck
```diff
@@ -76,6 +76,19 @@ namever_aur() {
     xargs -r aursearch -ir | jq -er '.results[] | [.Name, .PackageBase, .Version] | @tsv'
 }
 
+namever_vcs() {
+    for pkg in "$@"; do
+        base=$(aursearch -ir "$pkg" | jq -er '.results[] | .PackageBase')
```
maximbaz: This is needed for split packages, unless there is an easier way to get its folder name in `~/.cache`?

bin/aurcheck
```diff
@@ -76,6 +76,19 @@ namever_aur() {
...
+        cd "$HOME/.cache/aursync/$base" || continue   # TODO proper error handling, here and below
```
Foxboron: Can probably use `cd_safe` if  we move the `makepkg/util.sh` sourcing.

maximbaz: I'll replace hardcoded value with `AURDEST` from aursync, but I want your advice on handling errors, what do you want to see - just ignore error and continue to the next package, print something to stderr, ... ?

Foxboron: Refference: https://git.archlinux.org/pacman.git/tree/scripts/libmakepkg/util/util.sh.in#n79

bin/aurcheck
```diff
@@ -143,15 +158,33 @@ case $mode in
...
+                    aur+=("$pkg")
+                    echo "$line" >> "$tmp_aur"
+                    ;;
+            esac
+        done < "$tmp"
+
+        echo "${aur[@]}" | namever_aur | sort_join - "$tmp_aur" | \
+            sort_by "$pkg_key" - | cmp_all
+
+        namever_vcs "${vcs[@]}" | sort_join - "$tmp_vcs" | \
```
maximbaz: To describe the approach: split `$tmp` into AUR and VCS arrays, handle AUR ones just like before, but use the `namever_vcs` magic on VCS packages to get the new versions, then sort, compare and print exactly as for AUR ones.

bin/aurcheck
maximbaz: Ah I posted my question like 20 seconds later than you 😃 Just tested `cd_safe`, it prints "Aborting..." and exits the function early and does not let the for loop continue for other "valid" packages. I'm curious on the general approach though, I feel like it's better to ignore the package and continue the check for the rest of the vcs packages, what do you think?

maximbaz: This is `cd_safe`:

```
❯ aurcheck -d maximbaz-aur
/usr/share/makepkg/util/util.sh: line 80: cd: /home/maximbaz/.cache/aursync/alacritty-terminfo-git: No such file or directory
==> ERROR: Failed to change to directory /home/maximbaz/.cache/aursync/alacritty-terminfo-git
    Aborting...
alacritty-git 0.1.0.729.g59b561b-1 -> 0.1.0.730.gb82622e-1
alacritty-terminfo-git (not available)
antigen-git (not available)
aurutils-git (not available)
dragon-git (not available)
gtk-theme-arc-grey-git (not available)
i3ipc-python-git (not available)
libva-git (not available)
libva-intel-driver-git (not available)
lscolors-git (not available)
pam_u2f-git (not available)
py3status-git (not available)
python-azure-git (not available)
universal-ctags-git (not available)
```

This is `cd ... || continue`:

```
❯ aurcheck -d maximbaz-aur
/usr/bin/aurcheck: line 82: cd: /home/maximbaz/.cache/aursync/alacritty-terminfo-git: No such file or directory
alacritty-git 0.1.0.729.g59b561b-1 -> 0.1.0.730.gb82622e-1
alacritty-terminfo-git (not available)
aurutils-git 1.5.3.r266.gdc2c713-1 -> 1.5.3.r270.g199cc3b-1
universal-ctags-git 0.r5769.66c7915e-1 -> 0.r5771.5e4eca42-1
```

maximbaz:
I'm going to push an additional change to `aursync` which would assume "--no-ver" for VCS packages, because:

```
❯ aurcheck -d maximbaz-aur
alacritty-git 0.1.0.729.g59b561b-1 -> 0.1.0.730.gb82622e-1

❯ aursync -scu
  -> Using [maximbaz-aur] repository
==> Resolving dependencies...
    alacritty-git 0.1.0.729.g59b561b-1 is newer than 0.1.0.569.g4e9b1c5-1
    there is nothing to do
```

* `0.1.0.730.gb82622e-1` is the latest available version (👍)
* `0.1.0.729.g59b561b-1` is my currently installed version (👍)
* but `0.1.0.569.g4e9b1c5-1` is a complete garbage, it's `pkgver` reported by RPC (👎).

With the proposed change, `aursync -scu` correctly discovers all outdated VCS and regular AUR packages.

AladW:
I had a somewhat different approach in mind. In my view, checking VCS packages is logic that better belongs in aursync:

* Any required variables such as `AURDEST` are already defined there.
* Updating pkgver requires to source the PKGBUILD via `makepkg`, so the user should be given the chance to inspect PKGBUILDs and any new commits to them first.
* In the end what happens is that local repository versions are compared against whatever comes out from `pkgver()` in a PKGBUILD that's already on disk. So it's not really "checking the AUR".

So I would proceed as follows.

1. In the step where versions are compared in `aursync`, filter out any candidates that match the naming convention (or preliminary candidates).
2. Run git clone/pull on these candidates as with any other packages.
3. In a new loop, do the `makepkg -os` and `makepkg --printsrcinfo` dance for each package. It should not be necessary to loop `aursearch` (this would be expensive on many packages) since all metadata was already retrieved via `aurchain`.
4. Now repeat version comparison as in 1., using the new available information.

maximbaz:
My primary argument against your proposal is that I have a monitoring configured, which can run `aurcheck` in background and notify me when there are packages to update. For this purpose it is really useful to have `aurcheck` actually check for available updates, it tells me in the same run how many AUR and how many VCS packages are available, and then I can make a decision on when to actually upgrade my system.

Users will be able to inspect PKGBUILD when they run `aursync`, but `aurcheck` only runs `pkgver()`, there's nothing really to review at this point of time.

AladW:
You don't want it in `aursync` and I don't want it in `aurcheck`, so that leaves a simple but obvious compromise: keep it in a separate script and use the pipeline.

https://paste.xinu.at/85jr/

aurcheck would need some adjustments to print outdated repo versions with `aurcheck -c`, but the goal would be:
```
vcs ~/.cache/aursync/* | aurcheck -cd myrepo
```
The wildcard may be replaced by some pattern matching the naming convention.
```
find -regex (...) | xargs vcs | aurcheck -cd myrepo
```

maximbaz:
This is an interesting approach, but before we go into it, let me ask you to confirm, why do you feel against putting it all in `aurcheck`? 

> Any required variables such as `AURDEST` are already defined there.

True, but it would be just two lines of code to define it here, and if user configured custom `AURDEST`, both scripts will respect it.

> Updating pkgver requires to source the PKGBUILD via makepkg, so the user should be given the chance to inspect PKGBUILDs and any new commits to them first.

I agree that PKGBUILD needs to be examined during the time of `aursync`, but the code to update `pkgver()` does not and will not require reviewing PKGBUILD, whether we put it in `aurcheck` or `vcs` script.

> In the end what happens is that local repository versions are compared against whatever comes out from pkgver() in a PKGBUILD that's already on disk. So it's not really "checking the AUR".

Is it just a conceptual difference, or something more? It is checking the available updates for AUR packages, but does so without querying RPC. Most of the code in the `aurcheck` is doing versions comparison and printing outdated packages, which is identical for both types of AUR packages, the difference is only that some require querying RPC and some don't.

AladW:
Honestly, now that I've thought of a pipeline approach I prefer that over putting a lump of code in either `aursync` or `aurcheck`. It fits the modular approach to aurutils much better, and you can put such a pipeline wherever you want (whether it's in `aurcheck`, `aursync` or your own wrapper script).

maximbaz:
Hehe, I see 🙂 I like the modular approach too, just wanted to confirm.

I didn't see these before:

> That way people can make their own definition of "VCS package" by simply adjusting the pattern.

This I like very much!

> Note that I went back to -od, leaving people to diagnose on missing VCS clients on their own accord. The scope of -s is simply too wide for the purpose of running pkgver, and no sudo prompts occur with -d either.

This plays nicely with the idea of allowing users to define their own meaning of "VCS package", I like this too!

-------------

I saw you pushed `aursrcver`, I'm gonna play with it now.

maximbaz:
Now that I think about modularity, I've got the following idea, what do you think about it?

1. Extract RPC calls from `aurcheck` to `aurrpcver` - this script will host `namever_*` code and print the latest versions of AUR packages in the same way as `aursrcver` does.
2. Make `aurcheck` always accept the list of packages and their versions on stdin (basically the result of `aurrpcver` or `aursrcver`) - it's sole purpose becomes to compare the input against current repo and nicely print sorted list of packages that are outdated (maybe a rename to `aurcompare` is in order).

This way we gain three independent scripts with single purpose:
- `aurrpcver` finds the latest available versions of installed AUR packages
- `aursrcver` finds the latest available versions of installed VCS packages
- `aurcheck` tells you which of the given packages need an update

AladW:
Yes, I've always had the impression there was too much unrelated functionality in `aurcheck`. I usually can't say if splitting scripts makes sense until I try doing so myself, though.

maximbaz:
Awesome, let's split it between us then, I already tested that `aursrcver ~/.cache/aursync/*-git` works perfectly, I'll continue with this PR to get `aursrcver ~/.cache/aursync/*-git | aurcheck -cd myrepo` working (which should also allow piping regular AUR packages), and you can have a look at extraction unrelated functionality out of `aurcheck`.

This is the goal that I have in mind:

```bash
$ # check obsolete vcs packages (can be run by monitoring tools)
$ aursrcver ~/.cache/aursync/*-git | aurcheck -cd myrepo
$ # check obsolete aur packages (can be run by monitoring tools)
$ aurrpcver | aurcheck -cd myrepo
$ # upgrade the obsolete packages (executed by a person)
$ { aurrpcver; aursrcver ~/.cache/aursync/*-git } | aurcheck -cd myrepo | aursync -sc
```

Note that if we have the modular scripts `aursrcver` and `aurrpcver`, perhaps it doesn't make sense to have `-u` in `aursync` if we can just pipe the list of obsolete packages like that.

AladW:
There's two reasons `-u` in still in `aursync`:

* The repo selection mechanism, so you can add a single file repo to pacman.conf and have `aursync -u` automagically work.
* Piping stuff into aursync doesn't work well since it expects input from the user (to review files and confirm any aurbuild/makepkg prompts).

The rest should work, depending on what happens to an `aurrpcver` script. Note it was mentioned a while ago that it's unexpected that `aurcheck` waits for input from `stdin` if no arguments are specified, so perhaps an extra option could be added. (`-i`? unfortunately the standard `-` doesn't play nice with `getopts`)

maximbaz:
Got it, but in this case we've got to make `aursync -u` use both `aursrcver` and `aurrpcver`, I'd really appreciate upgrading all kinds of AUR packages in one go.

> Note it was mentioned a while ago that it's unexpected that aurcheck waits for input from stdin if no arguments are specified

Hehe I had exactly the same confusion until I looked at the script itself.

maximbaz:
Just realized that we don't have to pipe to `aursync` to make use of modular scripts, `xargs` could help us here, something like that:

```bash
$ { aurrpcver; aursrcver ~/.cache/aursync/*-git } | aurcheck -cdq myrepo | xargs | aursync -sc
```

This is just an idea and may or may not be an overkill, and as you said having `aursync` guess the name of the repo is useful too.

maximbaz:
I'm playing with this now, and I've got to say, if we can assume the list of packages with their NEW versions always comes via stdin, the `aurcheck` becomes very very simple. Could you have a look when you have a minute and share your thoughts? I've deleted a bunch of parameters because they become obsolete now (again, assuming `aurrpcver` and `aursrcver` exist and print out a list of packages and their new versions).

In meantime I'm going to look into making `aurrpcver` script now, because we need it to print the complete picture, to understand if this approach is desirable or not.

bin/aurcheck
```diff
@@ -118,12 +71,8 @@ fi
 unset database root
 while getopts :abcd:lqr: OPT; do
```
maximbaz: reminder: letters need cleanup

maximbaz:
☑ Monitoring works

```
$ aurrpcver -d myrepo | aurcheck -d myrepo                # RPC
$ aursrcver ~/.cache/aursync/*-git | aurcheck -d myrepo   # .SRCINFO
```

☑ Explicit upgrade works

```
$ { aurrpcver -d myrepo; aursrcver ~/.cache/aursync/*-git; } | aurcheck -qd myrepo | xargs aursync -sc
```

☑ Implicit upgrade works (as a prototype, this requires hardcoding the meaning of VCS packages)

```
$ aursync -scu
```

Waiting for your comments 🙂

bin/aursync
```diff
@@ -228,7 +229,7 @@ aurchain -t -- "$@" >pkginfo_0
   done
 
   if ((chkver)); then
-      cut -f1,3 pkginfo_0 | aurcheck -d "$repo" -r "$root" -c
+      cut -f1,3 pkginfo_0 | aurcheck -d "$repo" -r "$root"
```
maximbaz: Drawing your attention here because I can't come up with a good way to test this, but I think it should work™

AladW:
Thanks, I'm a bit busy right now so I'll take a look at it later.

Do note the second point on aursync though - any interactive questions (like interacting with vifm or y/N from pacman) will take input from `xargs` as well.

AladW:
See http://asciinema.org/a/nXDFkSPGiAwtr0kev37Rfl7Xp

maximbaz:
Thanks, I reproduce... but I'm completely puzzled, how is this even possible? How can `xargs` affect inners of the script, if its only job is to pass CLI arguments? Why my simple script above works? 😕

maximbaz:
Fascinating - [this](https://stackoverflow.com/a/30050792) explains it pretty well, the solution is to use `xargs -a <(echo flann) aursync -s` - then it will work 🙂

And just to be clear, I'm not arguing against `-u`, I'm just curious if this modularization enables just another cool approach to install and upgrade packages.

rafasc:
The merge (f430535) made `aurrpcver` output `[.Name, .PackageBase, .Version]` instead of the previous `[.Name, .Version]`. This breaks piping the output directly to `aurcheck` since the it expects "`pkgname\tversion`"

maximbaz:
Thank you very much, I fixed the merge issue. And thanks for testing too 🙂

rafasc:
You're welcome. Thanks for this PR!

Was something like `aurver --rpc <repo> --src <path>` considered? Having both `aurrpcver` and `aursrcver` combined since they're just version query scripts as in they output a package name and its version depending on the context.

Not sure how hard or if it's a good idea in the first place.

maximbaz:
This wasn't considered, because it's pretty easy to combine the output of `aursrcver` and `aurrpcver` with regular bash: 

```bash
{ aurrpcver -d myrepo; aursrcver ~/.cache/aursync/*-git; }
```

You can then process the combined result in any way you want, e.g. pipe it to `aurcheck`:

```bash
{ aurrpcver -d myrepo; aursrcver ~/.cache/aursync/*-git; } | aurcheck -d myrepo
```

Because it's so simple, it can be on a user to create an alias for this, if needed. 

Also, after this PR is completed, I'm planning to make all scripts automatically retrieve repo name if only one repo is configured (what `aursync` is doing today), that way for most people the commands will become even simpler:

```bash
{ aurrpcver; aursrcver ~/.cache/aursync/*-git; } | aurcheck
```

maximbaz:
@AladW do you think you'll get a chance to review this soon? 🙂

I'm personally happy with `aurcheck` and `aurrpcver` (want to know your opinion), but `aursync` needs some attention, it _has to know_ the definition of VCS packages and I don't really want to hardcode it - it needs this info not only for `-u`, but also to properly support `chkver` (e.g. for packages with `-git` dependencies, `aurchain` uses RPC so it will never mark such dependencies as required for an update).

AladW:
I'd like to focus on #288, since it directly affects which logic is placed where. As to the general idea, I'm not sure I'd like to drop most version checks to only have "`->`" - the other directions are useful too.

For `aursync` I think checking VCS packages should be behind some flag like `--vcs`, which is disabled by default. After all it behaves a lot differently than the usual `-u`.

maximbaz:
OK, let's return to this topic then after #288. For now, I can answer the comments:

> As to the general idea, I'm not sure I'd like to drop most version checks to only have "->" - the other directions are useful too.

Sure, makes sense. I think the new `cmp_newer` should stay the new default behavior, since is needed for `aursync` (and useful for monitoring of new package upgrades), the other two `cmp_equal_or_newer` and `cmp_all` can be hidden behind flags. I'll not make any changes now, I feel #288 will require a complete rewrite of this PR anyway 😞 

> For aursync I think checking VCS packages should be behind some flag like --vcs, which is disabled by default. After all it behaves a lot differently than the usual -u.

I don't think the implementation details should drive the need for a new flag, but rather users' understanding of the feature, and to me the "upgrade" means proper upgrade, regardless of the package type. On the other hand, both `yay` and `pacaur` have this flag too (called `--devel`), and I guess people find it useful, so... sure.

AladW:
>I'll not make any changes now, I feel #288 will require a complete rewrite of this PR anyway 😞

Most of the existing scripts are unchanged so far - just some functionality that has been shuffled around. The repo selection from `aursync` could be moved to `/bin/aur` so other scripts such as `/lib/build` can benefit. 

`aurrpcver` was done more in a general form via `/bin/aur-query`, which only outputs JSON (If [FS#49089](https://bugs.archlinux.org/task/49089?project=2) is fixed, it could be reduced to 5 lines) and is called from other scripts in `/lib`.

The next step is to see what to remove and add from `aurcheck` (now `/lib/updates`). Functionality to reliably handle split packages should remain in any case (the distinction between `pkgname` and `pkgbase`, as seen in `cmp_all` seems to have been removed in one of the commits). 

It has absorbed much over the years and in a similar vein to what's proposed here, it could be reduced to something that takes versions from `stdin` and compares against either the AUR or repository versions. (Note that the two last tasks were initially split, but then merged again to being similar in nature) 

Functions that go beyond these concepts (such as `namever_sync`) could be merged to the `/bin/aur` wrapper, which would be come a generalized `aursync` of sorts.

maximbaz:
> Functionality to reliably handle split packages should remain in any case - the distinction between pkgname and pkgbase, as seen in cmp_all seems to have been removed in one of the proposed commits.

Is that needed for manual use only? Because I do have split packages, both regular AUR and VCS, and the current PR seems to work just fine for both of them, unless I'm missing something.

AladW:
Well, `aursync` ends up using it (packages are piped to `aurcheck -q`, which calls `cmp_all`). Try with something like `python-virtualfish`.

maximbaz:
Using the code in this PR I ran `aursync -sc python-virtualfish`, changed `pkgver` to an older `1.0.4` and proceeded - this added `python-virtualfish` and `python2-virtualfish` to my repo. Then I installed with pacman only `python-virtualfish`. Then I ran `aurrpcver -d myrepo | aurcheck -d myrepo` (which is also exactly what `aursync -scu` will run in this PR), and I got the response:

```
python2-virtualfish 1.0.4-5 -> 1.0.5-5
python-virtualfish 1.0.4-5 -> 1.0.5-5
```

Then, to be absolutely sure, I also ran `aursync -scu`, changed `pkgver` back to `1.0.5` and proceeded with installation, and both `python-virtualfish` and `python2-virtualfish` were added to my repo.

Everything seems fine, isn't it?

AladW:
I'm not sure, it took a long time to get the split packages thing right, so I hope you won't blame me for being conservative here. In any event, the move to `/lib` has now been done. ;)

maximbaz:
No worries, have a closer look at the code, do some testing and let me know how it went 🙂 

I have an idea, since you want `--vcs` flag to aursync, how about we require a value for this flag that is the definition of VCS package, perhaps some kind of a regex? Because aursync needs this info, for `-u` and for detecting VCS dependencies, but I'm hesitant to hardcode one. We can still hardcode the default value, for example `-(bzr|git|hg|svn)$`, which is what [makepkg supports](https://www.archlinux.org/pacman/PKGBUILD.5.html#_using_vcs_sources_a_id_vcs_a):

> Currently makepkg supports the Bazaar, Git, Subversion, and Mercurial version control systems.

AladW:
Ah, the `pkgbase` logic was introduced for the `-b` flag. This allows to check AUR updates by using `pkgbase` as input instead of `pkgname`. I don't recall ever using this though.

For `aur-sync`, I'm thinking to move some of the functions from `aur-updates` there and rename latter to something else, e.g. `aur-vercmp`.

>We can still hardcode the default value, for example -(bzr|git|hg|svn)$, which is what makepkg supports

I guess you can do that. But we're still a while away from reaching that point...

maximbaz:
`pkgbase` querying - makes sense, and I see you've already removed the `-b` flag 👍 

`aur-vercmp` - if I understand your intentions correctly, that's similar to what I wanted to achieve for `aurcheck` in this PR too, a small and focused script that only checks versions.

AladW:
I've created the `vcs` branch with the following changes:

* pacman-related code was moved to `aur-vercmp-db` (unfinished)
* Use formatters instead of multiple `cmp` functions with duplicated logic

The idea is that `aur-vercmp` takes input from `stdin` and compares either against the AUR or a specified file (e.g. containing repo contents). What remains is to finish `aur-vercmp-db` and the formatters in `aur-vercmp`.

```
[archie@v22018013960760236 lib]$ cat vcs | ./aur-vercmp -F repo -a
aurutils-git 1.5.3.r356.gd98a4f4-1 = 1.5.3.r356.gd98a4f4-1
aws-cli-git (not available)
etl-git r8246.826849460-1 = r8246.826849460-1
nlohmann-json-git 3.1.1-1 = 3.1.1-1
openrct2-git (not available)
pikaur-git 0.5+11+ga8d428b-1 <- 0.5+9+gb3fd2d2-1
shaman-git (not available)
synfig-git (not available)

[archie@v22018013960760236 lib]$ cat repo | ./aur-vercmp -F vcs -a 2>/dev/null
aurutils-git 1.5.3.r356.gd98a4f4-1 = 1.5.3.r356.gd98a4f4-1
etl-git r8246.826849460-1 = r8246.826849460-1
nlohmann-json-git 3.1.1-1 = 3.1.1-1
pikaur-git 0.5+9+gb3fd2d2-1 -> 0.5+11+ga8d428b-1
```

maximbaz:
Awesome! Please create a PR when it's testable, I'm eager to try it out and leave some feedback 🙂

I'm especially interested in seeing how it all ends up being integrated, I'm using this branch for a couple of weeks now and I find it so useful that `aursync -u` updates both regular and vcs packages in one go, hope we can keep it like that (with `--devel` flag as you mentioned earlier).

AladW:
See #298. What remains is to either flip the arguments to `fmt_brief()`, or otherwise modify it to print "->" on "1" instead of "-1". (regular update functionality relies on `1`, `aur-srcver` on `-1`).

maximbaz:
Things are moving very quickly (which I like 😃), but I'm a little struggling to understand what's the current status of scripts... There was `aur-vercmp-db`, then there was `aur-checkupdates` and now neither exist.

Also, in your example above, when I want to compare vcs against repo, you seem to use `repo` file, not `-d myrepo`? Is the plan to have `cat vcs | ./aur-vercmp -d repo` instead of `cat vcs | ./aur-vercmp -F repo` long term in order to see which vcs packages are out of date?

AladW:
I think the following makes sense:
```
[archie@v22018013960760236 lib]$ cat repo | ./aur-vercmp -p vcs
aurutils-git: 1.5.3.r406.g455b319-1 -> 1.5.3.r407.gc64af87-1
aws-cli-git: - -> 1.14.41.r5559.ga42d6336-1
nlohmann-json-git: - -> 3.1.1-1
pikaur-git: - -> 0.5+22+g4c43ad4-1
shaman-git: - -> 2.0.0.α.r42.g73c048d-1
```
After all, the VCS versions are the "target" to compare to, much like the AUR. The source would be the contents of a repository as usual. With `-d` it looks like this:
```
[archie@v22018013960760236 lib]$ ./aur-vercmp -p vcs -d custom
aurutils-git: 1.5.3.r406.g455b319-1 -> 1.5.3.r407.gc64af87-1
aws-cli-git: - -> 1.14.41.r5559.ga42d6336-1
nlohmann-json-git: - -> 3.1.1-1
pikaur-git: - -> 0.5+22+g4c43ad4-1
shaman-git: - -> 2.0.0.α.r42.g73c048d-1
```

maximbaz:
This makes sense!

So one-liner would be something like this, except that it shows a weird warning:

```
$ aur vercmp -d myrepo -p <(aur srcver ~/.cache/aursync/*-git)
==> WARNING: vercmp: /proc/self/fd/11: empty file
alacritty-git: 0.1.0.731.g53a5a96-1 -> 0.1.0.736.g03f9e0c-1
aurutils-git: 1.5.3.r279.g2a0dbe0-1 -> 1.5.3.r409.g222005e-1
py3status-git: 3.7.r240.g3a2d0b2a-1 -> 3.7.r250.g03aee277-1
universal-ctags-git: 0.r5793.30ce6ee7-1 -> 0.r5807.66b4867b-1
```

The results are correct, and this already works on `master`.

What is missing is `-q` option in `aur-vercmp` (to print only package names without versions, so that we can use the output of `aur-vercmp` as input for `aur-sync`) and `--devel` option in addition to `aur-sync -u` (to update vcs packages together with regular ones) - do you agree? Do you want me to take a stab or do you want to do this yourself?

AladW:
On Mon, Feb 19, 2018 at 02:04:40AM -0800, Maxim Baz wrote:
> This makes sense!
> 
We should also update the man page since it took long enough to figure
it out... @Foxboron :-)

> So one-liner would be something like this, except that it shows a weird warning:
> 
> ```
> $ aur vercmp -d myrepo -p <(aur srcver ~/.cache/aursync/*-git)
> ==> WARNING: vercmp: /proc/self/fd/11: empty file
> alacritty-git: 0.1.0.731.g53a5a96-1 -> 0.1.0.736.g03f9e0c-1
> aurutils-git: 1.5.3.r279.g2a0dbe0-1 -> 1.5.3.r409.g222005e-1
> py3status-git: 3.7.r240.g3a2d0b2a-1 -> 3.7.r250.g03aee277-1
> universal-ctags-git: 0.r5793.30ce6ee7-1 -> 0.r5807.66b4867b-1
> ```
> 
> The results are correct, and this already works on `master`.
> 
The warning was fixed with 81e021f8584ecd8555fa25cf961f233a5199c339

> What is missing is `-q` option in `aur-vercmp` (to print only package names without versions, so that we can use the output of `aur-vercmp` as input for `aur-sync`) and `--devel` option in addition to `aur-sync -u` (to update vcs packages together with regular ones) - do you agree? Do you want me to take a stab or do you want to do this yourself?
> 
For the -q option I've added a colon in the output instead, so you can
easily do `aur vercmp | cut -d: -f1` - aur-sync does this already. Feel
free to propose something for `--devel`.

maximbaz:
Here's the initial proposal, it's basically a rebase on master, let me know how it looks. I still have the hardcoded `-git`, I'd like to make the definition of "vcs" configurable somehow, preferably with the default value `-(bzr|git|hg|svn)$`. What do you think about a yet another flag `--vcs '-(bzr|git|hg|svn)$'`?

maximbaz:
@AladW any chance to get this reviewed and merged? This became a relatively simple change now.

AladW:
My time for Arch stuff is non-existent right now, so any non-trivial changes ended up on the backlog. :frowning_face:

AladW:
Some issues:

* `aur srcver` is run both before any upstream changes (AUR) are merged, meaning output may be inaccurate, and before PKGBUILDs are inspected, meaning code is run before it may be viewed.
* `aur srcver` should not be run for all packages with VCS ending, but only for those present in the repo (`cut -f1 db_info | xargs -I{} find ... -name {}`)
* The `aur srcver` output is immediately lost after the argv step (due to process substitution) i.e cannot be used in later steps. A file like `vcs_info` may be suited.
* The "unless you use -f the packages will not be rebuilt anyway" is a hack that relies on how `aur build ` specifically handles existing packages. A different approach is needed here.
* Always assumes `AURDEST`, even if snapshots are used. This should be addressed by https://github.com/AladW/aurutils/commit/598871920721dc07810e782c42fdd1e11c6d46ca

As said I won't get to look further at this in a while. If you want to get your hands dirty, here's enough other things left in the 2.0.0 milestone though.

AladW:
@maximbaz https://github.com/AladW/aurutils/issues/302#issuecomment-375860725

maximbaz:
I know you are busy so I wasn't updating much here (except periodical rebases on `master`), however slowly but surely I'm addressing your comments 🙂

The current status:

**To do:**
- [ ] aur srcver is run both before any upstream changes (AUR) are merged, meaning output may be inaccurate, and before PKGBUILDs are inspected, meaning code is run before it may be viewed.

> This is true, but we need to update the versions to know if this PKGBUILD even needs to be presented to the user or not... So currently I don't see how this could be done differently, but I also don't experience any particular issue with this - after all, `aur srcver` only updates the version.

- [ ] The "unless you use -f the packages will not be rebuilt anyway" is a hack that relies on how aur build specifically handles existing packages. A different approach is needed here.

> I have a different approach in mind, will tackle next time 😉 

**Done:**
- [x] aur srcver should not be run for all packages with VCS ending, but only for those present in the repo (cut -f1 db_info | xargs -I{} find ... -name {}) to reduce performance issues (#286)
- [x] The aur srcver output is immediately lost after the argv step (due to process substitution) i.e cannot be used in later steps. A file like vcs_info may be suited.
- [x] Always assumes AURDEST, even if snapshots are used. This should be addressed by 5988719

AladW:
>This is true, but we need to update the versions to know if this PKGBUILD even needs to be presented to the user or not... So currently I don't see how this could be done differently, but I also don't experience any particular issue with this - after all, aur srcver only updates the version.

It only updates the version, but the PKGBUILD is executed in full nonetheless. Helpers like bauerbill and aurman use fancy mechanism to guess if a user already viewed a PKGBUILD, but I'm not fond of such approaches (instead, the viewing process itself should be made as convenient as possible).

What I had in mind was two "stages" - one for regular packages, and another for VCS packages. `vifm` has recently added tabs (https://github.com/vifm/vifm/issues/123#issuecomment-363430270), so we could use one tab for regular packages and one for VCS packages for better differentiation. (Also of note: https://github.com/vifm/vifm/issues/254#issuecomment-315124289)

maximbaz:
Could you explain what is the issue with PKGBUILD being executed in full? 

In my personal case, I have background process that is periodically running `aur srcver` and showing me an indicator on panel when there are packages that need an update. So in my case PKGBUILD was potentially executed tens of times before I got to `aur sync` it.

AladW:
A PKGBUILD is arbitrary code from arbitrary users, not sure what needs more explaining. In the past, even perfectly valid PKGBUILDs have been vandalized to e.g. point to malicious source URLs.

maximbaz:
I think I see the point of confusion 🙂 This PR is not running `aur-srcver` on _new_ PKGBUILDs, in only runs it on the PKGBUILDs of packages that are already in the database. If they are in database, user has reviewed them last time, and now we are running `aur-srcver` only to update the version.

AladW:
Yes, but as I pointed out above, you'd need to fetch new PKGBUILDs with `git pull` anyway. For example, some -git package may have changed the source to a newer updated fork after the original author abandoned the project.

maximbaz:
Where and when does the `git pull` happen? We `aur-srcver` only once, in the beginning, presumably on old PKGBUILDs yet.

AladW:
The point is it should happen, like I commented in https://github.com/AladW/aurutils/pull/283#issuecomment-373962475. 

>aur srcver is run both before any upstream changes (AUR) are merged, meaning output may be inaccurate

If you're using your own update script then that doesn't matter, but in aursync I'd like to have a complete implementation.

maximbaz:
Aaah, I think I see what you mean now - there is no `git pull` today, but you want to run `git pull` before executing `aur-srcver`.

But then I don't understand the suggestion with two stages. If you don't check the version, you don't know which packages have updates and which don't, so the only option is to show ALL packages in vifm? Even though they will be listed in a separate tab in vifm, it will be a lot of packages...

AladW:
>Even though they will be listed in a separate tab in vifm, it will be a lot of packages...

I don't think that's a big issue, since when the PKGBUILDs were already present on disk, any diffs will be displayed as separate files (both with git and tar approach) so you don't need to go over every directory. The ''ability'' to do latter is still present though.

https://github.com/AladW/aurutils/commit/46253a824453aa0bc3dfc60203dbee3af6e596c9 is interesting, that leaves only one place where AURVCS is actually used, i.e. in the `aur srcver` pipeline.

maximbaz:
I just think it's cleaner with the current approach, listen on RPC for potential major changes to PKGBUILD and otherwise do a quick srcver to just recalculate the version. Running `git pull` every time just to catch that one in a million case when a maintainer completely rewrote PKGBUILD and didn't bother running `makepkg` to regenerate pkgver... seems an overkill. And not worth sacrificing the nice and short list of packages that actually have updates, instead of showing a huge list of all VCS packages every time.

I don't see PKGBUILD diffs for vcs packages, by the way, presumably because of my periodic background job that runs `aur srcver` over and over again - not an "official" issue, but an issue for me 🙂

maximbaz:
@AladW I've seen some talk on the forum about this PR, so I thought, lets try again to clarify what exactly you dislike about the proposed approach?

1. PKGBUILD is being sourced by `aur-srcver` as part of `--upgrades`. 

Sure, but we only source _existing_ PKGBUILDs that user has reviewed in the past. Why running `aur-srcver` as part of `aur-sync` is considered harmful, but running `aur-srcver` on its own is not?

2. The need to run `git pull`, because `source` might have changed and maintainer didn't bother to update .SRCINFO when pushing the update to AUR.

First, why is it a problem of `aur-sync` and not `aur-srcver`? We can argue that `aur-srcver` can potentially give away incorrect versions because it checks the wrong source. However, running `git pull` would be a drastic slowdown, and I believe the gain is enormously tiny. If a package was unmaintained for a long time, and suddenly becomes maintained again, don't you think the maintainer will test out their PKGBUILD and thus update `pkgver`, so the package gets picked up by RPC version check?

maximbaz:
Do you agree that what you mention is related to `aur-srcver`, and not to `aur-sync`?

If user removed AURDEST, `aur-srcver` will not discover the package. If `aur-srcver` doesn't discover it, this PR will not update this package - user has to manually sync the package again, and review PKGBUILD when doing so.

This PR sources _only_ existing PKGBUILDs, it never sources something that is not reviewed.

With `git fetch`, let alone that I think again this is a problem of `aur-srcver` and not `aur-sync`, are you sure the problem you are trying to solve with... it is worth solving?

AladW:
>Do you agree that what you mention is related to aur-srcver, and not to aur-sync?

I mean, we're going in circles right now. Maybe that means `aur-sync` should have no `--devel` switch and this functionality is better implemented as a separate program that wraps `aur-sync`.

AladW:
As far as I understand things, this should be mostly equivalent:
```
#!/bin/bash
set -e
readonly XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
readonly AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/$argv0}
readonly AURVCS=${AURVCS:-.*-(bzr|git|hg|svn)$}
readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
tmp=$(mktemp -d)

db_tsv() {
    jq '.[] | [.pkgname, .pkgver] | @ tsv'
}

db_join() {
    grep -E "$1" | xargs -I{} find "$2" -maxdepth 1 -type d -name {}
}

ifne() {
    xargs -r "$@"
}

cd "$tmp"
# get repo contents
aur sync --list | db_tsv >db_info

# checkout latest revision for existing pkgbuilds
# this sources the pkgbuilds assuming it has been viewed priorly
cut -f1 db_info | db_join "$AURVCS" "$AURDEST" | ifne aur srcver >vcs_info

# sync VCS packages that are outdated according to aur-srcver
# RPC information may be outdated and is thus disabled
aur vercmp -p vcs_info < db_info | cut -d: -f1 | ifne aur sync --no-ver
```
Usage for a full upgrade:
```
aur sync -u
aur sync-devel
```

maximbaz:
That could work, I would only suggest naming this script `aur update-devel` to indicate that syncing _new_ VCS packages should still happen via `aur sync`. 

In fact, what do you think about extracting update functionality of regular packages also out of `aur sync`? That way, `aur sync` deals with syncing new packages, detecting dependency tree, etc, and `aur update` only handles updates of packages in a repo?

Whether have two scripts `update` and `update-devel` or one script for both, that doesn't matter much.

AladW:
Now that we have a sensible way to take package names from stdin, thanks to #368, the update functionality might as well be removed from `aur-sync`, yes.

maximbaz:
Would you like to implement the extraction? That way you can play with the code and decide whether you want one or two update scripts. The snippet you posted above indeed looks equivalent to this PR.

And I would be happy to test a PR 😛

AladW:
Looking over the code, the only statement is this:
```
  if ((update)); then
      aur vercmp <db_info | cut -d: -f1
  fi
```
So I'm not sure if it's worth moving out this single line (also considering `aur sync -u` is probably the most used `aurutils` command)

AladW:
Since I edited my post above, what do you think about an `examples` directory`? That might include other scripts like `aur-remove` and `aur-gc`. (Alternatively you could have a separate repo like `aurutils-contrib`)

maximbaz:
Nice! 

The name `examples` a bit implies that everyone likely needs to adapt the script to their needs, which is not the case for `update-devel` 🙂 

Just for my understanding, what meaning do you intend to put in the difference between having scripts in the `aurutils` vs `aurutils/examples`?

AladW:
My reasoning was as I put it in the commit message: scripts that are a simple combination of `aurutils` programs but don't fit inside a single pipeline would fall in the examples directory. The name is perhaps not ideal (is `contrib` better?), but if you didn't want to adapt the script you could symlink it somewhere in your path.

maximbaz:
I see, makes sense, `contrib` sounds better 🙂

AladW:
Great, is there anything left to discuss for `aur-sync` or can we continue this in #372?

maximbaz:
Nothing left, I'll close this PR, let's continue in #372.

PR #287, opened 2018-01-30
aursearch: allow sort by votes

### LINKED ISSUE #279: aursearch: configurable sort field (and maybe sort order)
`yay` does a very good job at sorting the results, by default it sorts by the number of votes and shows most popular packages in the bottom. I used it for a long time and it proved to be a good sorting approach.

That's why I think it would be cool if `aursearch` also accepted `sortby=votes|name`, and maybe even a `sortorder=topdown|bottomup`. As for the `sortorder`, initially I thought I could just reverse the entire output, but it won't be as neat, because every record consists of two lines.

What do you think about all of this?

maximbaz:
This adds `-V` to sort by votes.

It turns out I don't need reversed search, because I just realized that ascending sorting by votes already puts most voted packages in the bottom - that's exactly what I wanted 🙂

I'm fine to close #279 with this PR, and implement reverse order only if someone actually needs this.

### DIFF
--- a/bin/aursearch
+++ b/bin/aursearch
@@ -106,12 +106,12 @@ parse() { json_short; }
 by=name-desc
 sort_by=Name
 
-while getopts :imnVrRv OPT; do
+while getopts :imnk:rRv OPT; do
     case $OPT in
         i) by=none ;;
         m) by=maintainer ;;
         n) by=name ;;
-        V) sort_by=NumVotes ;;
+        k) sort_by=$OPTARG ;;
         r) parse() { tee; } ;;
         R) parse() { sed 's/^/[/; s/}{/},{/g; s/$/]/' ; } ;;
         v) parse() { json_long; } ;;
--- a/bin/aursearch
+++ b/bin/aursearch
@@ -34,10 +34,10 @@ dl_stdin() {
 }
 
 json_short() {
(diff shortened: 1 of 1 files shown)


AladW:
Thanks. I would probably make this more general and let the user specify what to search for; e.g. `aursearch -k NumVotes` or `aursearch -k Popularity` by using `$OPTARG`.

maximbaz:
I thought about that, but users would need to know the exact field names, make sure to respect capitalization, and the script itself should be error-resistant - are you sure you want all that? With `-V` everything is very simple and there is no way to make a mistake.

AladW:
I'm sure any user who uses `aursearch` is familiar enough with the RPC to know said names, or they can just be documented in the man page. If you specify the wrong field there's no error either - it just uses the default sort order. 
```
[archie@v22018013960760236 bin]$ ./aursearch -k Tacos aurutils
aur/aurutils-git 1.5.3.r234.g15ef2ab-1 (6)
    helper tools for the arch user repository
aur/aurutils 1.5.3-8 (60)
    helper tools for the arch user repository
```
I prefer the added flexibility in this case.

AladW:
Applied with some small changes. Cheers

maximbaz:
Awesome, thanks!

Issue #184, opened 2016-10-25
aurbuild: hook support


clawoflight:
I would like to propose adding support for some kind of hooks that allow users to define additional requirements that packages should fulfil before they are built.

The main idea would be to add pre-build hooks that have semantics similar to git pre-{commit,push,receive} hooks, i.e. the current action should be cancelled if a hook exits with non-zero.

I propose defining a folder in which hooks can be placed as executable files. `aursync` or `aurbuild` would execute them in the order returned by `find`.

Because of the strong coupling between aurutils and the system, I propose making this folder system-wide, such as `/etc/aurutils/hooks/pre-build.d/`.
The hooks could be passed the path to the untarred or cloned build files from AUR and maybe the pkgbase, name of the committer, etc.

Some applications I can think of:
- check hashes for specific versions against a trusted database (which is what I am working on)
- a trust management system (some maintainers are blindly trusted for specific packages, others require manual confirmation)
- impose custom restrictions on the system by making a hook write-protected and immutable and preventing other users from changing that
- turn your Philips Hue into a red strobe light whenever someone tries to build yaourt :wink:
- anything you can think of
  and generally allow loosely-coupled extensibility of aurutils, which is a great thing in itself.

I am ready to implement this, and it won't require much work to make it usable in a basic form.
I opened this issue to let people know what I plan on doing and to ask for feedback, especially what exactly others might want to use, and where the hooks should go.

AladW:
I've talked a bit about this on #archlinux, and they suggested that instead of hooks, you could expand on the existing idea of using pipelines. For example:
```
aurchain <pkgs> | check_hashes | check_trust | aurbuild
```
Furthermore these parts of the pipeline could have varying outputs depending on their use case. e.g.  you could have `aurbuild` return failed/succesful builds on stdout, or have `aurchain` print pkgbase instead of pkgname.

How does that sound?

clawoflight:
Sounds good for my own task, but I don't really see a way to make that loosely coupled.
The nice thing about hooks is that everyone could profit from them and no-one has to fork or patch aurutils for anything.

Earnestly:
A pipeline is already loosely coupled while still having tight cohesion.

Any filter you write can be inserted where ever you wish so long as each tool properly makes use of the (universal interface)[http://dl.acm.org/citation.cfm?id=191771] without needing to hardwire unrelated mechanisms such as hooks or triggers.

http://www.leancrew.com/all-this/2011/12/more-shell-less-egg/

[...]

PR #289, opened 2018-02-02
aur: allow execution of user defined scripts

rafasc:
Provide a similar mechanism that `git` has for user defined sub-commands.
`git` allows the use of any executable in `$PATH` with prefix `'git-'` to be used with the main `git` command.
E.g. `'git-name'` can be invoked with `$git name`.

In aa59551, I used executables on `$PATH` that were prefixed by `'aur-'`. But in retrospect that was probably a poor choice since [there are already some packages](https://aur.archlinux.org/packages/?K=aur-) which use executables with the `'aur-'` prefix that would end up there.
One alternative is to use another prefix, something like `'aurutils-'`.

Another would be having a dedicated directory for sub-commands scripts. E.g.
```bash
USER_CMDS_DIR=${XDG_CONFIG_HOME:-$HOME/.config}/aurutils/usrcommands
```
Besides giving the user free choice over the naming convention, not having unexpected results due to similar executables by other packages, this would also make listing available scripts (for use in completions or help messages) easier.

I think the second option is preferable, but i'll wait some feedback before pushing another commit.

### DIFF
--- a/bin/aur
+++ b/bin/aur
@@ -2,19 +2,28 @@
 readonly argv0=aur
 readonly cmd_dir=/usr/lib/aurutils
 readonly cmd=("$cmd_dir"/*)
-readonly custom_cmd_prefix='aur-'
+readonly usrcmd_prefix='aur-'
+readonly usrcmd=($(compgen -c $usrcmd_prefix))
 
 if [[ -z $1 ]]; then
     printf >&2 -- 'usage: %s [command]\n\n' "$argv0"
     printf >&2 -- 'available commands:\n'
-    printf >&2 -- '    %q\n' "${cmd[@]##*/}"
+    printf -- '%q\n' "${cmd[@]##*/}" | column 1>&2
+
+    if [[ ${#usrcmd[@]} -gt 0 ]]; then
+        printf >&2 -- '\navailable user commands:\n'
+        #ignore user commands that clash with official ones.
+        sort <(printf -- "%q\n" ${usrcmd[@]#aur-}) \
+             <(printf -- "%q\n" ${cmd[@]##*/}) \
+             <(printf -- "%q\n" ${cmd[@]##*/}) | uniq -u | column 1>&2
+    fi
     exit 1
 fi
 
 if [[ -x $cmd_dir/$1 ]]; then
     exec "$cmd_dir/$1" "${@:2}"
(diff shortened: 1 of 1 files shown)


AladW:
Well, from a purely selfish perspective the current commit would be useful, since there's now `/bin/aur-query` and `/bin/aur-sync`. I can see your points for using a different approach though.

maximbaz:
By the way, why do `/bin/aur-query` and `/bin/aur-sync` exist instead of `/lib/query` and `/lib/sync`?

rafasc:
Personally, I think we shouldn't do both but it's not technically impossible... The builtin `compgen -c 'aur-'` could be used for grabbing commands from the `$PATH` that match on cases listing is needed (e.g. `aur` called without arguments).

But in the end, I assume users would probably just create something like  `.config/aurutils/scripts` to house their scripts, which is a point in favor of the second approach.

'Config directory' version could look something like [this](https://github.com/rafasc/aurutils/blob/custom-scripts-support-xdg-config/bin/aur). Which I'll merge into this if we decide it's a better way to go forward.

Also interested in the motivation behind what @maximbaz mentioned.

AladW:
I've kept thinking about this but couldn't come up with a good reason to implement such PATH extensions. Essentially, it reminds me of the proposed support for hooks. (see #184)

So that goes back to the `aur-` prefix. The same scenario applies to `git-` - any users can upload their own programs to e.g AUR and have them be called by `git` (and in fact, many do). These programs would continue working as if they were called on their own.

To avoid confusion, `git` has a nice mechanism in the form of `git help -a`:
```
available git commands in '/usr/lib/git-core'
...
git commands available from elsewhere on your $PATH
...
```
As to `aur-sync`, I see it as something that wraps the commands, much in the vein of `aur`, rather than provide some core functionality of its own. Hence `/bin`. The same cannot be said of `aur-query`, though, so I'll think on whether to have `/lib` scripts call `/bin/aur` or simply use a fixed path.

AladW:
In mind with the above remark on `git help -a`, I've moved all scripts back to `/lib` to allow for easy separation from third-party programs in `/bin`.

rafasc:
In the future, if core `aur-` scripts are added to `/bin`, it should be enough to manually add them to `cmd` array.

AladW:
I'll merge this and merge the branch to master at some point. Thanks

rafasc:
Thank you!

PR #295, opened 2018-02-13
sync: AURURI variable

AladW:
Usually I would update the documentation, but there's none present for `aur-fetch` anyway. (@Foxboron)

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -1,67 +1,58 @@
 #!/bin/bash
 readonly argv0=fetch
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
-readonly aurweb='https://aur.archlinux.org'
-readonly urlpath='/cgit/aur.git/snapshot'
 set -o errexit
 
-declare    logdir=/dev/null
-declare -i snapshot=0 default_path=1
+declare logdir=/dev/null
+declare mode=git
 
 clone() {
-    declare -r remote=$1 logdir=$2
+    declare -r log_dir=$1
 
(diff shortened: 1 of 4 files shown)


lib/aur-fetch
```diff
@@ -1,67 +1,58 @@
...
-            git clone "$remote/$REPLY".git
+            git clone "$REPLY"
         fi
     done
 }
 readonly -f clone
 
 usage() {
-    printf '%s\n' "usage: $argv0 [-L logdir] [-t]" >&2
+    printf '%s\n' "usage: $argv0 [-L log_dir] [-t]" >&2
```
Foxboron: `printf '%s\n' "usage: $argv0 [-L log_dir] [-t mode]" >&2`

lib/aur-fetch
```diff
@@ -1,67 +1,58 @@
...
 
-while getopts :L:Pt OPT; do
+while getopts :L:t OPT; do
     case $OPT in
-        L) logdir=$OPTARG ;;
-        P) default_path=0 ;;
-        t) snapshot=1     ;;
-        *) usage          ;;
+        L) log_dir=$OPTARG ;;
+        t) mode=tar        ;;
```
Foxboron: This probably needs to be `$OPTARG`?

lib/aur-fetch
AladW: The idea was that the default "mode" is `git`, which is then set to `tar` if you specify the `-t` flag. Should be less typing and matches the old `aur-fetch` interface. I guess the initial mode declaration can be moved from the top of the script to getopt to make this more clear.

Foxboron: Right, this makes sense. I'll update the documentation accordingly

PR #308, opened 2018-02-20
Fix sudoers line in aur-build manpage

AlexNPavel:


AladW:
Thanks for the PR. It seems that adding `aur-build-nspawn` to sudoers doesn't work anymore, probably due to the use of `sudo aur build-nspawn` (i.e., `/bin/aur`). Not sure if whitelisting `/bin/aur` (and with that, all `aur-` scripts) is a good idea though. Happy to take suggestions.

maximbaz:
This seems to work for me:

```
NOPASSWD:SETENV: /usr/bin/aur build-nspawn *
```

AladW:
According to `sudoers(5)`:
```
     Wildcards in command line arguments should be used with care.
     Command line arguments are matched as a single, concatenated string.  This mean a wildcard character such as ‘?’ or ‘*’ will match across word boundaries, which may be unexpected.  For example, while a sudoers entry like:

         %operator ALL = /bin/cat /var/log/messages*

     will allow command like:

         $ sudo cat /var/log/messages.1

     It will also allow:

         $ sudo cat /var/log/messages /etc/shadow

     which is probably not what was intended.  In most cases it is better to do command line processing outside of the sudoers file in a scripting language.
```
Some other things you can do:

* Implement some check in `aur(1)` to see if called with sudo, and fail unless the command is `build-nspawn`
* Pass on sudo to the `exec`'d command to preserve `sudoers` rules
* Move the `sudo` call in `aur-build` to `aur`. I doubt this is a good idea since this moves the assumption that `aur-build-nspawn` is run as root to some other place.

AlexNPavel:
I must've accidentally ran a sudo command before testing that without realizing it. Updated commit uses @maximbaz's suggestion

AlexNPavel:
@AladW Is there much potential of abuse with that sudoers line though? We have `/usr/bin/aur build-nspawn *`, not `/usr/bin/aur build-nspawn*`, so it will only execute the `aur build-nspawn` command (with arguments) as sudo, correct? In what circumstance would this cause problems?

eli-schwartz:
This isn't really any more dangerous than having /usr/bin/extra-x86_64-build allowed, which is to say it isn't entirely safe... allowing any user to enter a chroot and build/install packages inside that chroot is hardly safe.

AladW:
Yeah, that's how I understand it as well. I'd still consider adding a note/warning on wildcards with sudoers (with a simple referral to the man page) if we add such an example.

AlexNPavel:
The sudoers line specifies `%build`, meaning that the user has to be in the build group already, meaning that you already trust the user to run builds.

That being said, we can add a warning about wildcards for this example

eli-schwartz:
Well, sure. Arch Linux uses a similar command to allow Devs and TUs to run extra-x86_64-build on the build server at https://pkgbuild.com

https://git.archlinux.org/infrastructure.git/tree/roles/archbuild/files/sudoers

It is, nevertheless, a security consideration. But no more than it was before this pull request.

AladW:
In the pacman 5.1 branch, I went back to whitelisting the respective devtools scripts. The initial reason for whitelisting the full script was commands such as `install(1)` that required root, however these are already run by `mkarchroot` where needed.

PR #311, opened 2018-02-25
preliminary bash completion

rafasc:
This is still in a very early state. Just opening the PR in order to get some feedback and/or help. 😊 
~~The main idea here is to get autocompletion from the output of `-h` pages. So as long as the help pages get updated, completions for short and long options come for free.~~

~~While this is convenient, in some cases we might want to define completions for other things, e.g repos for `-d` option etc. This will provide completions based of `-h` flags unless there's a function defining how the command should complete.~~ (moving away from parsing help text. see: https://github.com/AladW/aurutils/pull/311#issuecomment-401462680).

Stuff to consider
- [x] ~~Better and shorter help lines. Some need proper wrapping, some probably need to be rewritten. It's hard to be short without being _too_ short.~~ (moved away from parsing help text)
- [x] ~~Symlinks for standalone commands. Right now, standalone commands like `aur-build` only get completions after the completion for `aur` is called at least once. This can be fixed by adding symbolic links to the main completion. E.g. `aur-build -> aur` . (probably rename the completion file to _aurutils_ and do `'aur{,-*}' -> aurutils`~~
- [x] ~~Overrides for other scripts providing extra functionality. ( e.g. autocomplete repos for -d )~~ Override mechanism is there (call to `_aurutils_${words[1]}`) but no functions are implemented for now.
- [x] ~~Better code? This is my first autocompletion script. There is probably stuff done wrong all over.~~ 😁  Good enough.

### DIFF
--- a/completions/Makefile
+++ b/completions/Makefile
@@ -1,0 +1,8 @@
+PREFIX ?= /usr
+SHRDIR ?= $(PREFIX)/share
+
+bash/aur: command_opts.m4 bash/aurutils.in
+	m4 $^ >$@
+
+install: bash/aur
+	@install -Dm644 bash/aur $(DESTDIR)$(SHRDIR)/bash-completion/completions
--- a/completions/bash/aurutils.in
+++ b/completions/bash/aurutils.in
@@ -1,0 +1,32 @@
+_aur_completion()
+{
+    local cur prev words cword
+    _get_comp_words_by_ref cur prev words cword
+
+    SUBCOMMANDS
+
+    # complete subcommands
+    if [[ $cword -eq 1 ]]; then
+        COMPREPLY=( $(compgen -W "${subcommands[*]}" -- "$cur") )
+        return
+    fi
+
+    # If there's an override for subcommand, use it
+    if declare -F "_aurutils_${words[1]}" >/dev/null ; then
+        "_aurutils_${words[1]}"
+        return
+    fi
+
+    # Complete with the generated opts stored above, unless the previous option
+    # is stored with an : suffix, because the option requires an argument.
+    # Fallback to default (files) completion in such cases.
+
+    opts=($(_get_default_opts ${words[1]} ))
+    if [[ ${opts[*]} != *$prev:* ]]; then
+        COMPREPLY=($(compgen -W "${opts[*]%:}" -- "$cur"));
+    fi
+}
+
+DEFAULT_OPTS
+
+complete -o bashdefault -o default -F _aur_completion aur
--- a/completions/command_opts.m4
+++ b/completions/command_opts.m4
@@ -1,0 +1,41 @@
+divert(-1)dnl
+dnl List of commands to be completed
+dnl
+define(`CORECOMMANDS',
+HAVE_OPTDUMP(build,fetch-snapshot,search,fetch,repo-filter,
+             repo,chroot,pkglist,fetch-git,vercmp,sync,rpc)dnl
+NO_OPTDUMP(depends,jobs,graph,srcver))
+
+dnl recursively print all elements
+dnl How the element is ultimately printed depends on the
+dnl wrap_has_dump and wrap_no_dump macros
+dnl
+define(`HAVE_OPTDUMP',`ifelse(`$#',`0', ,`$#',`1',`wrap_has_dump(`$1') ',`wrap_has_dump($1)' `HAVE_OPTDUMP(shift($@))')')
+define(`NO_OPTDUMP',`ifelse(`$#',`0', ,`$#',`1',`wrap_no_dump(`$1') ',`wrap_no_dump($1)' `NO_OPTDUMP(shift($@))')')
+
+dnl override how elements are printed for SUBCOMMANDS macro
+dnl which defines an array with all avaliable subcommands
+dnl
+pushdef(`wrap_has_dump','$1')
+pushdef(`wrap_no_dump','$1')
+define(SUBCOMMANDS,subcommands=(CORECOMMANDS))
+
+dnl override how elements are printed for the DEFAULT_OPTS
+dnl which is a simple function that returns the options available for a
+dnl subcommand using case statement construct.
+dnl
+pushdef(`wrap_has_dump',`
+	$1) printf -- GET_OPTS($1) ;;')
+pushdef(`wrap_no_dump',`
+	$1) : nothing to print ;;')
+define(DEFAULT_OPTS,
+`_get_default_opts() {
+    case "`$'1" in CORECOMMANDS
+    esac
+}')
+
+dnl Helper macro to retrieves options from subcommand --dump-options
+dnl
+define(GET_OPTS,'`translit(esyscmd(aur $1 --dump-options),`
+',` ')'')
+divert(0)dnl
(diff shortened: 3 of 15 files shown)


Foxboron:
Hum. I can probably piggyback on this for a zsh completion!

rafasc:
zsh's `_gnu_generic` understands `--help` but unfortunately not `-h`. ( I believe it's the `--` option of `_arguments`).

If you could load those dynamically for the `aur` wrapper would be nice although that would also imply moving from `getopts` to `getopt` in every script.

AladW:
If there's no other solution I don't mind moving to getopt... but ideally there's a simpler alternative.

AladW:
@rafasc Note that with pacman 5.1 we can move to `parseopts.sh` which should allow easy availability of long options in all scripts

rafasc:
I have something that almost works for zsh along the same lines of this PR, i.e. allow custom completions but fallback to help based ones. 

`_arguments --` calls the first word in the command line with the `--help` flag and parses the output.

We could just fool the completion for `aur-<cmd>`  call `aur-help <cmd> --help` by changing the first word of the command line from `aur-<cmd>` to `aur-help <cmd>`. Where `aur-help <cmd>` wrapper would return the `aur-<cmd> -h` page.

https://github.com/rafasc/aurutils/blob/rafasc/WIPzsh_completion/completions/zsh/_aurutils#L33

This is "working", but relying on such hack will bite us at some point. In fact, there's already weird things going on:

E.g. `aur-deps-src -<tab>` doesn't hang (even though it should(?) because I'm not handling the fact it doesn't support `-h` or `--help` and it waits for stdin, maybe there's no stdin available?, who knows...).  
And without analysing what's really happening inside `_arguments`, it's hard to tell what's going wrong and _definitely_ not worth the hassle.

And since all this can be easily solved by supporting the long `--help` flag, moving to `getopt` or `parseopts.sh` looks like the better idea if we want to continue with this route.

AladW:
Perhaps it would be easier to consider "two-dash" commands such as `aur-deps-src` as internal and not consider them for tab-completion?

rafasc:
Sorry If I made things confusing by talking about zsh's case in this PR. The bash case that is covered. https://github.com/AladW/aurutils/pull/311/files?diff=unified#diff-debe164994a5f5a80b8205619460adefR117  
`_parse_help` can call any option you provide on the second parameter. i.e `_parse_help cmd --opt` (can also parse stdin if the first parameter is `-`)

But in zsh's case the `--help` option is not configurable because it is hardcoded inside `_arguments`:
https://github.com/zsh-users/zsh/blob/ed4e37e45c2f5761981cdc6027a5d6abc753176a/Completion/Base/Utility/_arguments#L91

The issues I mentioned [above](/AladW/aurutils/pull/311#issuecomment-371203855) are related to this limitation, but only concern the zsh completion which is not part of this PR (at least not yet). I guess it can wait until `pacman 5.1`

rafasc:
If we move to `getopt` it doesn't really matter if we have an `aur help` or not. That was just a way to avoid forcing scripts to implement long options just to `--help` available.

AladW:
I think I'll implement `getopt` soon, waiting for pacman 5.1 is like waiting for godot and the syntax for `getopt` and `parseopts` look closely related anyway. That said, I want to stabilize the interface first, i.e. be satisfied on the nomenclature of the `aur` scripts. 

In particular, I renamed `aur-deps-src` to `aur-arrow` (since that's exactly what the output is) and removed `aur-deps` (since it's functionality is trivial and now described in `aur-arrow(1)`. 

That leaves `aur-deps-rpc`. Perhaps `aur-buildorder` is more suitable (in vein of `auracle buildorder`), but it completes in an ugly manner with `aur-build` and `aur-build-nspawn`.

rafasc:
I've been busy with other stuff and this is the thing being constantly postponed :( 

I have a `getopt` conversion branch that needs some conflict fixing. I'll try to make a PR out of it later this week. This PR would be rebased and rebuilt on top of that.

---

Regarding the rename of new scripts, not sure if `aur-arrow` or `aur-buildorder` are the correct choices but I cannot come up with better ones, so 🤷‍♂️

But that probably deserves to be discussed as a separate issue.

AladW:
@rafasc with pacman 5.1 released, `parseopts` should be almost identically to use as `getopt`. You just don't need the exit code == 4 check (as I believe seeing in your getopt branch)

rafasc:
Now that we have #406, `parseopts` related topics should follow from there.

jrudess:
I'm trying to take advantage of this tree for tcsh completions.  If it is useful for anyone else, I've [merged](https://github.com/jrudess/aurutils/tree/jrudess/bash_completion) this against master and hopefully didn't introduce too many bugs.

rafasc:
@jrudess I would not use this for anything.

I think the original idea of parsing help text is too fragile and a bit of a pain to maintain. 
Writing help text while leaving it compatible with both zsh's and bash mechanisms to parse
help text is obnoxious. Sometimes a simple comma trips one or the other.

I think these help parsers mechanisms are great when you don't have control of the software you're writing the completions for. But this isn't the case here.

@AladW do you think something like this would be ok, instead of the current work?
``` diff
+    --dump-options) printf -- '--%q\n' ${opt_long[@]%:} ;
+                    printf -- '%q' "${opt_short//:}" | while read -n 1 sopt; do
+                        printf -- '-%q\n' $sopt;
+                    done
+                    exit ;;
+    --)             shift; break ;;
```

I have an alternative, incomplete and a bit ugly branch based on this idea:
https://github.com/AladW/aurutils/compare/master...rafasc:simplerbashcompletion?expand=1

I understand this is one of the topics holding the release, but I'll be somewhat time constrained in the near future, so I am perfectly ok if someone else wants to take this out of my hands.

AladW:
For the while loop, the shortest (no pun intended) I can think of is
```
printf '%q' "${opt_short//:}" | fold -w1 | sed 's/^/-/g'
```
(Why is %q required?)

>I understand this is one of the topics holding the release, but I'll be somewhat time constrained in the near future, so I am perfectly ok if someone else wants to take this out of my hands.

Same here, take all the time you want/need.

AladW:
Also does the `--dump-options` need to include some indication if an option takes an argument?

rafasc:
Using `sed`, The shorter I could come up is:
`printf -- '%s' "${opt_short//:}" | sed 's/./-&\n/g' `

>(Why is %q required?)

It isn't. I'll replace it with `%s`, thanks.

>Also does the --dump-options need to include some indication if an option takes an argument?

I didn't consider this, but it can make sense to leave that indication in order to have a different default completion for options with required arguments. (Maybe by leaving the `:` suffix, at the expense of having to peel them off on the other end).

On the other hand, anything smarter than autocompleting to files, would require writing override functions, so I think the benefit is limited.

AladW:
On an unrelated matter, it probably makes no sense to have all the `--nofoo` aliases (instead of `--no-foo`), since all it does it clutter the completion

AladW:
>On the other hand, anything smarter than autocompleting to files, would require writing override functions, so I think the benefit is limited.

Sure, but autocompleting files should already cover most cases (off-hand, I only know of `aur sync -d` and `aur-sync --ignore` where this would not apply)

rafasc:
I think it's the right choice. (As I slightly suggested via IRC). I'll try to find some time to update this PR.

rafasc:
I am officially abandoning the previous idea in favor of the approach mentioned in https://github.com/AladW/aurutils/pull/311#issuecomment-401462680 with 1e68ddd

PR #322, opened 2018-03-18
man formatting

FichteFoll:
While reading the man pages *a lot* yesterday, it bothered me that usage of bold and italics was rather inconsistent. This PR aims to:

- Unify bold and italic usage
- Repeat option arguments in the OPTIONS sections

And while working on those, I noticed that a couple (read: many) hyphens needed escapation, because otherwise they become subject to line wrapping.

The only non-formatting changes were changing one "workarounds" to "works around" and changing a "suffix" to "prefix", both in aur-sync(1).


Foxboron:
@FichteFoll <3<3

AladW:
Nice work! :clinking_glasses:

PR #326, opened 2018-03-20
aur-search: Add indicator of orphaned and out-of-date packages

maximbaz:
Inspired by yay 🙂

I realized what I'm missing from yay - indicators "orphaned" and "out-of-date". I don't think any configuration options are needed to disable this behavior.

image

### DIFF
--- a/lib/aur-search
+++ b/lib/aur-search
@@ -26,8 +26,8 @@ json_short() {
         read -r OutOfDate
         read -r Description
     }; do
-        [[ "$OutOfDate" ]] && OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
-        [[ "$Maintainer" == "" ]] && Maintainer="(Orphaned)" || Maintainer=""
+        [[ -n "$OutOfDate" ]] && OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
+        [[ -z "$Maintainer" ]] && Maintainer="(Orphaned)" || Maintainer=""
         printf "${BLUE}aur/${ALL_OFF}${BOLD}%s ${GREEN}%s ${ALL_OFF}(%s) ${RED}%s %s${ALL_OFF}\\n    %s\\n" \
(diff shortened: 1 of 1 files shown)


AladW:
Looks good to me. You should also declare OutOfDate/Maintainer local: https://github.com/AladW/aurutils/blob/master/lib/aur-search#L13

This made me look a bit into `jq` options, and `//` looks useful:
```
.OutOfDate // ""
[[ -z $OutOfDate ]] && OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
```
Not sure if there's something similar for the negation `[[ "$Maintainer" != "null" ]]`

maximbaz:
Nice 👍 

I decided to make use of `//` in both places, to avoid a case if maintainer's username is "null" 🙂

lib/aur-search
```diff
@@ -26,8 +26,8 @@ json_short() {
         read -r OutOfDate
         read -r Description
     }; do
-        [[ "$OutOfDate" == "null" ]] && OutOfDate="" || OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
-        [[ "$Maintainer" != "null" ]] && Maintainer="" || Maintainer="(Orphaned)"
+        [[ "$OutOfDate" ]] && OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
+        [[ "$Maintainer" == "" ]] && Maintainer="(Orphaned)" || Maintainer=""
```
AladW: You could remove this line by using `${Maintainer:-(Orphaned)}` in the `printf` command below. :blush:

maximbaz: But if maintainer is not empty, we need to clear its value (not to print their name)

AladW:
Yeah, there's even maintainers called `orphan`...

lib/aur-search
AladW: Ah, right.

AladW: This works though: https://github.com/AladW/aurutils/commit/05cb9e36c15e9972e547a7cde0919226ac2498dd

maximbaz: Nice trick!

Issue #313, opened 2018-02-25
aur-search: intersection of results


pfactum:
It would be nice to have an option for aursearch to return an intersection of results for given terms instead of a union so that the search results could be more relevant.

AladW:
A naive way might be to do one request for each argument, and form the intersection with `join(1)`. Ideally the RPC also supports such functionality and a manual implementation is not required.

maximbaz:
I've always considered `aur-search` to be the `pacman -Ss` for aur packages. However `pacman -Ss` does intersection by default, not union, so I would even argue that the default behavior for `aur-search` must change to be intersection too.

AladW:
@pfactum please see if the above commit works as intended.

pfactum:
@AladW It looks so, thanks.

AladW:
Should be fixed with 167b0fcf6d4248f3e592cf4d30d4625c8e73c1e0

PR #328, opened 2018-03-21
aur-search: use intersection of results for multiple terms

maximbaz:
This is a prototype for #313.

Now the behavior of `aur search` matches `pacman -Ss`, when multiple search terms are provided, only packages that match ALL of the terms are returned.

I didn't create an option to keep the old behavior because I'm not sure if it is actually needed.

Feedback is welcome 🙂

image

### DIFF
--- a/lib/aur-search
+++ b/lib/aur-search
@@ -9,63 +9,51 @@ query_type=search
 search_by=name-desc
 sort_key=Name
 
-json_short() {
+parse_short() {
+    jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | [
+        .Name         // "-",
+        .Version      // "-",
+        .NumVotes     // "-",
+        .Maintainer   // "-",
+        .OutOfDate    // "-",
+        .Description  // "-"
+        ] | @tsv'
+}
+
+output_short() {
     local Name Version NumVotes Maintainer OutOfDate Description
 
-    jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | .Name,
-        .Version,
-        .NumVotes,
-        .Maintainer // "",
-        .OutOfDate  // "",
-        .Description' | while
-    {
-        read -r Name
-        read -r Version
-        read -r NumVotes
-        read -r Maintainer
-        read -r OutOfDate
-        read -r Description
-    }; do
-        [[ "$OutOfDate"  ]] && OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
-        [[ "$Maintainer" ]] && unset Maintainer
-
+    while IFS=$'\t' read -r Name Version NumVotes Maintainer OutOfDate Description ; do
+        [[ "$OutOfDate"  == "-" ]] && OutOfDate=""  || OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
+        [[ "$Maintainer" != "-" ]] && Maintainer="" || Maintainer="(Orphaned)"
         printf "${BLUE}aur/${ALL_OFF}${BOLD}%s ${GREEN}%s ${ALL_OFF}(%s) ${RED}%s %s${ALL_OFF}\\n    %s\\n" \
-               "$Name" "$Version" "$NumVotes" "${Maintainer+(Orphaned)}" "$OutOfDate" "$Description"
+               "$Name" "$Version" "$NumVotes" "$Maintainer" "$OutOfDate" "$Description"
     done
 }
 
-json_long() {
+parse_long() {
+    jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | [
+        .Name            // "-",
+        .PackageBase     // "-",
+        .Version         // "-",
+        .Description     // "-",
+        .URL             // "-",
+        (.Keywords       // ["-"] | join(" ")),
+        (.License        // ["-"] | join(" ")),
+        .NumVotes        // "-",
+        .Popularity      // "-",
+        .OutOfDate       // "-",
+        .Maintainer      // "-",
+        .FirstSubmitted  // "-",
+        .LastModified    // "-"
+        ] | @tsv'
+}
+
+output_long() {
     local Name PackageBase Version Description URL Keywords License
     local NumVotes Popularity OutOfDate Maintainer FirstSubmitted LastModified
 
-    jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | .Name,
-        .PackageBase,
-        .Version,
-        .Description,
-        .URL,
-        (.Keywords // ["(null)"] | join(" ")),
-        (.License  // ["(null)"] | join(" ")),
-        .NumVotes,
-        .Popularity,
-        .OutOfDate,
-        .Maintainer,
-        .FirstSubmitted,
-        .LastModified' | while
-    {
-        read -r Name
-        read -r PackageBase
-        read -r Version
-        read -r Description
-        read -r URL
-        read -r Keywords
-        read -r License
-        read -r NumVotes
-        read -r Popularity
-        read -r OutOfDate
-        read -r Maintainer
-        read -r FirstSubmitted
-        read -r LastModified
-    }; do
+    while IFS=$'\t' read -r Name PackageBase Version Description URL Keywords License NumVotes Popularity OutOfDate Maintainer FirstSubmitted LastModified ; do
         printf -- 'Name:          %s\n' "$Name"
         printf -- 'Base:          %s\n' "$PackageBase"
         printf -- 'Version:       %s\n' "$Version"
@@ -121,13 +109,43 @@ case $query_type in
     search) format=${format-short} ;;
 esac
 
-# set filters
+# set parsers
 case $format in
-     long) parse() { json_long  "$sort_key"; } ;;
-    short) parse() { json_short "$sort_key"; } ;;
+     long) parse() { parse_long  "$sort_key"; } ;;
+    short) parse() { parse_short "$sort_key"; } ;;
 esac
 
+# set output
+case $format in
+     long) output() { output_long;  } ;;
+    short) output() { output_short; } ;;
+esac
+
+# prepare for execution
+
+trap_exit() {
+    if [[ ! -o xtrace ]]; then
+        rm -rf "$tmp"
+    fi
+}
+
+tmp=$(mktemp -dt "$argv0".XXXXXXXX)
+trap 'trap_exit' EXIT
+
+source /usr/share/makepkg/util/util.sh
+cd_safe "$tmp"
+
 # pipeline
-printf -- '%s\n' "$@" | aur rpc -t "$query_type" -b "$search_by" | parse
+for query in "$@"; do
+    printf -- '%s\n' "$query" | aur rpc -t "$query_type" -b "$search_by" | parse > query_results
+    if [[ -f query_results_combined ]]; then
+        mv query_results_combined query_results_prev
+        grep -Fxf query_results query_results_prev > query_results_combined
+    else
+        mv query_results query_results_combined
+    fi
+done
+
+output <query_results_combined
 
 # vim: set et sw=4 sts=4 ft=sh:
--- a/lib/aur-search
+++ b/lib/aur-search
@@ -11,12 +11,12 @@ sort_key=Name
 
 parse_short() {
     jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | [
-        .Name         // "-",
-        .Version      // "-",
-        .NumVotes     // "-",
+        .Name,
+        .Version,
+        .NumVotes,
         .Maintainer   // "-",
         .OutOfDate    // "-",
-        .Description  // "-"
+        .Description
         ] | @tsv'
 }
 
@@ -33,19 +33,19 @@ output_short() {
 
 parse_long() {
     jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | [
-        .Name            // "-",
-        .PackageBase     // "-",
-        .Version         // "-",
-        .Description     // "-",
-        .URL             // "-",
+        .Name,
+        .PackageBase,
+        .Version,
+        .Description,
+        .URL,
         (.Keywords       // ["-"] | join(" ")),
         (.License        // ["-"] | join(" ")),
-        .NumVotes        // "-",
-        .Popularity      // "-",
+        .NumVotes,
+        .Popularity,
         .OutOfDate       // "-",
         .Maintainer      // "-",
-        .FirstSubmitted  // "-",
-        .LastModified    // "-"
+        .FirstSubmitted,
+        .LastModified
         ] | @tsv'
 }
 


AladW:
I'm working on a "fold" function for aur search, so that with long arrays like make/depends the output is properly wrapped like with `cower`/`pacman`. Once I'm done with that I'll take a look. :+1:

maximbaz:
Rebased to include new fields License and Keywords, as well as renamed "First Submitted" to "Submitted". Hope I didn't miss anything else.

image

maximbaz:
Although to be fair I don't see the License and Keywords because they are not returned by aur-rpc...

image

AladW:
Yeah, Keywords and License are not available in the search output. (Hence why I set an array with a (null) field in case those fields are absent, so json_long works with both search and info output)

maximbaz:
Aaaah... Then I'm happy to report that I did not break that functionality either 😄 

image

Morganamilo:
This still queries the RPC for each term right? If you only want the results that match all the terms then can't you just search the RPC for the first term. Then filter the results against the other terms. That should be that same and take 1 RPC request instead of N.

maximbaz:
It is possible, but I want to keep the change simple and not reimplement the filtering, don't want to bother with accounting for the field that it is being searched by, etc. Also, it sounds scary when you say N calls, but in reality 99.9% of usages will issue not more than... 3 RPC calls, I would say 🙂

AladW:
Actually, it shouldn't be complicated thanks to the power of `jq`...

https://rosettacode.org/wiki/Set#Finite_Sets_of_JSON_Entities

Morganamilo:
Yeah it might be a bit of over optimization but I do like to minimize the use of the RPC.

maximbaz:
Oh, and I thought `bash` syntax is weird, but now I've seen some `jq`...

```
# If A and B are sets, then intersection(A;B) emits their intersection:
def intersection(A;B):
  (A|length) as $al
  | (B|length) as $bl
  | if $al == 0 or $bl == 0 then []
    else 
      reduce range(0; $al + $bl) as $k
        ( [0, 0, []];
          .[0] as $i | .[1] as $j
          | if $i < $al and $j < $bl then
              if A[$i] == B[$j] then [ $i+1 , $j+1,  .[2] + [A[$i]]]
              elif  A[$i] < B[$j] then [ $i+1 , $j,  .[2] ]
              else [ $i , $j+1,  .[2] ]
              end
            else .
            end
         ) | .[2]
    end ;
```

maximbaz:
By looking a bit more I don't feel comfortable going this way, how about merging the current approach and improving in a separate iteration if needed? The split between parsing and printing in this PR will be useful for every approach.

AladW:
There seem to be some unrelated changes in the commits so it'll take some time to review

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
 search_by=name-desc
 sort_key=Name
 
-json_short() {
-    local Name Version NumVotes Maintainer OutOfDate Description
+parse_short() {
```
AladW: unrelated name change

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
...
+    jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | [
+        .Name         // "-",
```
AladW: unrelated change (and `//` only makes sense on fields that might be null or empty, never the case for name, version e.a.)

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
...
+        .Version      // "-",
+        .NumVotes     // "-",
+        .Maintainer   // "-",
+        .OutOfDate    // "-",
+        .Description  // "-"
+        ] | @tsv'
```
AladW: unrelated change, no benefit from going tsv instead of newline separated (more compact notation but makes tools like awk and read that use newlines by default harder to use)

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
...
-        read -r Name
-        read -r Version
-        read -r NumVotes
-        read -r Maintainer
-        read -r OutOfDate
-        read -r Description
-    }; do
-        [[ "$OutOfDate"  ]] && OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
-        [[ "$Maintainer" ]] && unset Maintainer
+output_short() {
```
AladW: unrelated name change

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
...
+    local Name Version NumVotes Maintainer OutOfDate Description
 
+    while IFS=$'\t' read -r Name Version NumVotes Maintainer OutOfDate Description ; do
+        [[ "$OutOfDate"  == "-" ]] && OutOfDate=""  || OutOfDate="(Out-of-date: $(date -d @"$OutOfDate" '+%d %B %Y'))"
```
AladW: unrelated change (old checkout?)

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
...
+        [[ "$Maintainer" != "-" ]] && Maintainer="" || Maintainer="(Orphaned)"
         printf "${BLUE}aur/${ALL_OFF}${BOLD}%s ${GREEN}%s ${ALL_OFF}(%s) ${RED}%s %s${ALL_OFF}\\n    %s\\n" \
-               "$Name" "$Version" "$NumVotes" "${Maintainer+(Orphaned)}" "$OutOfDate" "$Description"
+               "$Name" "$Version" "$NumVotes" "$Maintainer" "$OutOfDate" "$Description"
```
AladW: as L27

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
...
     done
 }
 
-json_long() {
+parse_long() {
```
AladW: As L14

lib/aur-search
```diff
@@ -9,63 +9,51 @@ query_type=search
...
+    jq -r --arg key "$1" '[.results[]] | sort_by(.[$key])[] | [
```
AladW: As L20

lib/aur-search
```diff
@@ -121,13 +109,43 @@ case $query_type in
     search) format=${format-short} ;;
 esac
 
-# set filters
+# set parsers
```
AladW: unrelated change, see also https://en.wikipedia.org/wiki/Filter_(software)

maximbaz:
I'll leave some comments to explain the changes and hopefully make review easier:

1. Functions `json_short` and `json_long` are split in two stages - parsing PRC results and printing. 

1. Because parsing is separate now, previously the output of `jq` parsing was immediately read, now it is converted via `@tsv` and being read later from this new format - this introduced problem with fields that are missing a value, `read` cannot handle `\t\t`. That's why I made `-` to be a default value for every single field that `jq` is trying to parse, to avoid any potential for future issues with `read`.

1. Because missing value is marked as `-` instead of empty string, detecting `OutOfDate` and `Orphaned` changed a bit too.

1. `Keywords` and `License` changed their default value from `(null)` to `-` for consistency with the rest of the fields.

1. Intersection of results is done with grep and temporary files, most of that code was inspired by similar pieces in `aur-sync`.

lib/aur-search
```diff
@@ -121,13 +109,43 @@ case $query_type in
...
+case $format in
+     long) parse() { parse_long  "$sort_key"; } ;;
+    short) parse() { parse_short "$sort_key"; } ;;
+esac
+
+# set output
 case $format in
-     long) parse() { json_long  "$sort_key"; } ;;
-    short) parse() { json_short "$sort_key"; } ;;
+     long) output() { output_long;  } ;;
```
AladW: As L12

lib/aur-search
```diff
@@ -121,13 +109,43 @@ case $query_type in
...
+    short) output() { output_short; } ;;
```
AladW: As L14

lib/aur-search
```diff
@@ -121,13 +109,43 @@ case $query_type in
...
+source /usr/share/makepkg/util/util.sh
+cd_safe "$tmp"
+
 # pipeline
-printf -- '%s\n' "$@" | aur rpc -t "$query_type" -b "$search_by" | parse
+for query in "$@"; do
+    printf -- '%s\n' "$query" | aur rpc -t "$query_type" -b "$search_by" | parse > query_results
+    if [[ -f query_results_combined ]]; then
+        mv query_results_combined query_results_prev
+        grep -Fxf query_results query_results_prev > query_results_combined
```
AladW: Does this preserve the validity of the JSON arrays? E.g. things like resultcount will definitely be wrong

AladW:
I started the review before viewing your comment. I can't say I'm very fond of all these changes, since they affect large parts of that code without this being actually required for intersection functionality.

lib/aur-search
AladW: Or I guess it parses the results first to TSV format, and then runs grep on it line-wise? Seems problematic if you want to expand functionality later

maximbaz:
Well yes, changes look large, but do you see any actual issues with them? Splitting monolithic "parse-and-print" into two components sounds like a good idea anyway, having `-` as a default value for all fields is just for consistency and possible issues with RPC when suddenly some fields will become missing.

lib/aur-search
AladW: https://github.com/AladW/aurutils/pull/328#discussion_r176923571

AladW:
Well, with the old code you would immediately see which fields would be potentially null/empty in the RPC output, now you can't. I think that's more valuable than a potential scenario where suddenly the AUR starts behaving in erratic ways (haven't seen that in 2 years anyway)

lib/aur-search
maximbaz: Exactly - why is it problematic though? When a more fancier functionality is available, it can read the TSV values in variables and do whatever it needs to do, it will have to split the current "parse-and-print" function anyway.

maximbaz:
I don't mind removing the default `-` from the fields that cannot ever be empty 🙂

lib/aur-search
maximbaz: Is it documented somewhere which exactly fields are optional and which are required? I was hoping it would be documented [here](https://wiki.archlinux.org/index.php/AurJson), but it doesn't seem to be the case.

AladW: I don't think so, unless you go investigate the PHP code! :grimacing:  What was there before is purely empirical

maximbaz: Then I see even less reasons to remove the default that saves us from potentially missing values... But I pushed this nevertheless, used the empirical data from `master` 🙂

AladW:
All the other things in the review need fixing too.

Note that the whole grep/file approach should not be required since you can just use `uniq -d`.
```
printf -- '%s\n' "$@" | aur rpc ... | jq ... @tsv | sort | uniq -d | while read ...
```

lib/aur-search
maximbaz: Functions `json_short` and `json_long` are split in two stages - parsing PRC results and printing. This is parsing, thus named `parse.

lib/aur-search
maximbaz: This is done

lib/aur-search
maximbaz: This is needed for removing duplicates.

lib/aur-search
maximbaz: Functions `json_short` and `json_long` are split in two stages - parsing PRC results and printing. This is the printing the output part, thus named `output`.

lib/aur-search
maximbaz: Because missing value is marked as `-` instead of empty string, detecting `OutOfDate` and `Orphaned` changed a bit too.

lib/aur-search
maximbaz: Replied there

lib/aur-search
maximbaz: Functions `json_short` and `json_long` are split in two stages - parsing PRC results and printing. This is parsing, thus named `parse`.

lib/aur-search
maximbaz: Removed default values from the field, this should be done.

lib/aur-search
maximbaz: if the functions are called `parse` and they are doing parsing, I think the comment should also say that

lib/aur-search
maximbaz: answered there

lib/aur-search
maximbaz: answered there

maximbaz:
Could you please highlight which exactly things need fixing? I replied to every comment.

Thanks for the `uniq -d` hint, will check it out!

AladW:
The changes are still to invasive for me to merge in a single PR. I'll probably give a shot at it myself with your `@tsv` idea and see what turns up

maximbaz:
Okay! 👍

AladW:
This is what I have in mind:
* call the while/read functions `fmt_long` and `fmt_short`, in line with `aur-fetch`, or `parse_long` and `parse_short` in line with the used pipeline
* call the `jq` functions `tabulate_long` and `tabulate_short`, to emphasize on `@tsv` and in line with `aur-deps-rpc`
* use `read -a` in `fmt_long` to save some space and since the fields are already clear from the `printf` lines
* leave out changes in format (hypens et al) for a separate commit or PR
* add an `-a` option which returns the union (`uniq -u`) instead of the intersection (`uniq -d`) of matched packages.

The end pipeline should look as follows:
```
printf -- '%s\n' "$'" | aur rpc -t "$query_type" -b "$search_by" | tabulate "$sort_key" | combine | parse
```

maximbaz:
Sounds very good! Just make sure that `read` doesn't break on tabulated string that is missing some values, that was the whole point of me adding `-` and changing formats (so that instead of `\t\t` I have `\t-\t` to parse) - maybe there is a way to achieve the same without changing the format that I didn't find 🙂

AladW:
https://lists.gnu.org/archive/html/help-bash/2016-05/msg00041.html

Any drawback in using `@csv` instead of `@tsv`?
```
user@v22016103960739491:~$ string=a,b,,d
user@v22016103960739491:~$ IFS=, read -a array <<< "$string"
user@v22016103960739491:~$ echo ${array[@]}
a b d
user@v22016103960739491:~$ echo ${array[0]}
a
user@v22016103960739491:~$ echo ${array[1]}
b
user@v22016103960739491:~$ echo ${array[2]}

user@v22016103960739491:~$ echo ${array[3]}
d
```

maximbaz:
Don't think so, I'm sure jq is smart enough to handle possible commas in the received data itself, but worth checking anyway.

AladW:
`jq` does the right thing but `read` has no knowledge about quoting:
```
user@v22016103960739491:~$ tabulate_short Name < out
"python2-octaviaclient","1.4.0-1",0,"andybz","","OpenStack Client plugin for Octavia, an OpenStack Load Balancing project"
user@v22016103960739491:~$ echo ${foo[5]}
"OpenStack Client plugin for Octavia
user@v22016103960739491:~$ echo ${foo[6]}
an OpenStack Load Balancing project"
```

AladW:
At the risk of angering the bash gods, we could use `@sh` and `eval`:

https://github.com/stedolan/jq/wiki/Cookbook#extract-data-and-set-shell-variables

maximbaz:
In my opinion `@tsv` and default value `-` is better, but perhaps I just don't see something bad about this approach that you do.

AladW:
Yeah, I can't think of anything better either (the alternatives above are hardly satisfactory). I do something similar in `aur-vercmp` for `join(1)` use anyway.

AladW:
Actually we don't need two jq functions, one suffices when you can ignore what's not required:
```
while IFS=$'\t' read -r Name _ Version Description _ _ _ NumVotes _ Maintainer OutOfDate _ _; do
```

maximbaz:
Nice finding!

AladW:
I ended up rewriting most of the script: https://github.com/AladW/aurutils/commit/c07e9571bdb4f7b0644b2700ac6069bcfc56df2e

Count me surprised if it still works.

AladW:
1. uniq -d prints all duplicate lines, if there are none to begin (in the case of one argument) with it prints none
2. I forgot to set the separator to tab for column(1), should be `column -s $'\t'`
3. on info searches, the union of results is already done by the AUR (and the intersection makes no sense to begin with), so it should be disabled in that case.

I've also changed the code so that column is only run once for every search instead of once for every result. An issue that remains is that `column` ignores empty lines. On Debian there's a `column -e` option, but in the actual non-patched-to-death version `-e` seems to mean something else entirely...

maximbaz:
1. I see, so basically we need to skip the `combine` step if there is only one entry in the `$@`?
2. Yep, that fixes the issue except for the missing empty lines. If there is no workaround, we can still run column once per result, and print `\n` after that, doesn't sound too bad...
3. So `combine` step should not only be skipped if there is only one entry in `$@`, but also if `-i` is provided?

AladW:
Running column once per result is fine if there are few results, but will slow things down drastically if there are a few thousands...

I've used `sed` for now but I'm hoping to find some better solution. All in all, while I think the move to `@tsv` is generally useful, I think the union and intersection of `search` results should be handled on the AUR side. Compare the ongoing discussion on aur-dev:

https://lists.archlinux.org/pipermail/aur-dev/2018-March/004495.html

AladW:
dcab37a added some reminders to the code, so I think this can be closed. It would still be a good idea to engage in the above aur-dev discussion though.

maximbaz:
I confirm that all the issues are fixed, thanks 🙂

Agree that it should be implemented on the AUR side.

> \# XXX add option for intersection or only leave -a?

I'd say intersection should stay default and without an option, because that's how `pacman -Ss` works. Consistency 🙂

AladW:
Works for me. Now we just need a man page update...

AladW:
One issue remains, and that is whatever I do, I can't use both `join(" ")` and `//` for Keywords/License. The output is always empty.

As powerful as jq is, I swear malbolge has a more pleasant UI at times...

PR #331, opened 2018-03-21
additional documentation fixes (follow-up to #322)

rafasc:
Here are some documentation improvements that I failed to get ready in time to piggyback #322.

Align options with descriptions if there is space for it.
Add some missing section numbers.
Add mention that sudo wildcarding aur might run aur- prefixed scripts from $PATH
Add some missed escapes for -
Fix inconsistency between manpage formatting of arguments inside square brackets


AladW:
Thanks!

PR #340, opened 2018-03-26
Make sure the is no extra space when package is not orphaned

maximbaz:
Just to be done with aur-search bugs 🙂

This fix kinda depends on the fact that `(Oprhaned)` is always before `Out Of Date`, not sure if you want something more sophisticated that this...

image

P.S. I found `-a` useful for showing relevant screenshots, it allows to gather all important cases in one picture 🙂

Fixes #337.

PR #341, opened 2018-03-26
Fix jq parsing

maximbaz:
https://github.com/AladW/aurutils/pull/340#issuecomment-376271509

/cc @Morganamilo


Morganamilo:
Yeah that seemed to fix it. nice one.

PR #343, opened 2018-03-26
aur-search: Fix order of fields so that votes are correctly shown

maximbaz:
One more bug:

image


AladW:
Right, whenever I change the jq part the short options need changing too... maybe some more robust approach (another array with `read -a`) would be useful. Until then, thanks for the fix. :+1:

PR #345, opened 2018-03-30
Add documentation for GPGKEY variable

null-git:
Added mention to the GPGKEY variable to use another key for signing. This variable doesn't seem to be documented at all.


man1/aur-build.1
```diff
@@ -61,6 +61,8 @@ database (\fBrepo\-add \-R\fR).
 .TP
 .B \-s
 Sign built packages and the database (\fBrepo\-add \-s\fR) with gpg.
+To use another key than the default, the GPGKEY environment variable can be
```
AladW: Ideally, `GPGKEY` would be put in bold (`\fBGPGKEY\fR`)

Issue #286, opened 2018-01-30
aur-srcver: improve performance


maximbaz:
On my laptop it currently takes 18 seconds on average to check 13 `*-git` folders with `aursrcver`, let's discuss ideas how we can speed it up.

maximbaz:
One obvious idea is to detect Github sources and instead of running `makepkg` use Github API to query the latest commits - this is what `yay` is doing, and you have already mentioned this in another issue.

```bash
$ curl https://api.github.com/repos/zsh-users/antigen/branches
```

I think Github is used a lot, so it would definitely improve the average case. For example, all of my 13 currently installed `-git` packages are hosted on Github.

However, this is not as simple as it sounds, and `yay` does not handle all cases correctly. Caveats:

1. You cannot assume that the branch is called `master`:
    - User could have changed branch in PKGBUILD by appending `#branch=custom` to the `source` parameter.
    - The repo itself could have a custom default branch - see [zsh-users/antigen](https://github.com/zsh-users/antigen) for an example, open the link and you are looking at the `develop` branch.

        This is how to detect default branch on Github:
        ```bash
        $ curl -s https://api.github.com/repos/zsh-users/antigen | jq .default_branch
        "develop"
        ```
2. You cannot easily tell if the commit corresponds to `pkgver`, because `pkgver()` can be defined by a PKGBUILD creator in various ways. Even the "common" approach is not obvious:
    * package: `antigen-git-v2.2.3.r6.gc91f77c-1-any.pkg.tar`
    * commit: `c91f77c8e9d96da43ae6dcaca4f2c823532302dc`

    Yay solves this by hosting its own file with latest installed commits, but this approach is also error-prone.

3. Github limits number of anonymous API calls, and at some point starts throwing 50x errors. Yay struggles a lot from this. The solution is to make the `aursrcver` optionally accept user token, or better yet a file path to where the token is stored.

AladW:
On Tue, Jan 30, 2018 at 02:21:05AM -0800, Maxim Baz wrote:
> One obvious idea is to detect Github sources and instead of running `makepkg` use Github API to query the latest commits - this is what `yay` is doing, and you have already mentioned this in another issue.
> 
I like the idea but as layed out it's complex to implement. I would try
some simpler approaches first:

* makepkg --printsrcinfo is slow as all hell. Remove it and the "mangle"
  function and simply source the PKGBUILD in a subshell.
* Use parallel(1) to run makepkg -od multiple times depending on
  available cores.

[...]

PR #346, opened 2018-03-30
Optimize aur-srcver using git ls-remote

maximbaz:
This is how `aur-srcver` would look like when using `git ls-remote`.

As I shared in #286, `aur-out-of-date` doesn't produce good results for me, that's why I went for looking into this approach.

For my packages, the runtime of `aur-srcver` decreases from `18s` to `8s`.

How this works:

- Read `source` out of PKGBUILD (e.g. `source = alacritty::git+https://github.com/jwilm/alacritty.git#branch=master`)
- Remove custom name (e.g. `alacritty::`)
- Remove all but the last protocol (e.g. `git+`) - do we need to keep it?
- Detect branch name, if it is specified
- Run `git ls-remote` on the source to get their last commit hash

#### Big problems:

1. This approach doesn't work for non-vcs packages, and I'm not sure how to make it work. Should we somehow detect non-vcs packages in the input and for them fallback to aur-rpc?

2. This now reports version as `<pkg-name>    <commit-hash>`, so it's not compatible with `aur-vercmp` anymore. This breaks #283 😞

Do you have some ideas how to overcome these issues?

### DIFF
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -9,13 +9,14 @@ get_pkgbuild_info() {
     env -C "$1" -i bash -c '
         PATH= source PKGBUILD
 
-        if [[ -v epoch ]]; then
-            fullver=$epoch:$pkgver-$pkgrel
-        else
-            fullver=$pkgver-$pkgrel
-        fi
+        source="${source#*::}"                           # remove custom name of a source
+        source="${source#*+}"                            # keep only one protocol (TODO: do we need many?)
 
-        printf %s\\t%s\\n "${pkgbase:-$pkgname}" "$fullver"
+        branch="${source#*#branch=}"                     # detect custom branch if it is specified
+        source="${source%#*}"                            # remove branch from the source string
+        [[ "$branch" == "$source" ]] && branch="HEAD"    # use HEAD if custom branch is not specified
+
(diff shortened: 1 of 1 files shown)


AladW:
So why are you removing `makepkg` completely? As it stands this would _only_ work with `git` packages, and not with the plethora of other VCS packages.

>This now reports version as <pkg-name> <commit-hash>, so it's not compatible with aur-vercmp anymore

I was about to ask what you'd do with the commit hash in the first place. `yay` stores the commit hash for built packages in some configuration file and compares with it later, which I suppose is functional but doesn't work outside `yay` or whatever program decides to implement such mechanism.

maximbaz:
> So why are you removing makepkg completely?

This is more of a prototype anyway, to test the performance and general approach, but I realized too late that there isn't really much I could do with git hashes. Because I don't see any way what to do with hashes (besides doing what yay does and store them in a file, but I don't really like that approach), I think it is safe to conclude that this approach will not work for us.

With `aur-out-of-date` approach we would keep makepkg call for all packages that are giving the UNKNOWN state, I only wish there were fewer UNKNOWN results... I'll make an issue in their upstream, will see what the author says about that.

PR #347, opened 2018-04-04
Use parallel to speed up aur-srcver

### LINKED ISSUE #286: aur-srcver: improve performance

maximbaz:
How it works:

1. `--halt soon,fail=1` is **not used**, I want to update as many packages as possible, if one keeps failing it shouldn't block the rest
2. `--results '{#}_makepkg'` saves stdout and stderr of every makepkg run, `{#}` is the sequence number, used later to determine which run failed
3. `--joblog makepkg_log` saves the `tsv` table of which command finished and which failed, in there I can see a command, its exit code and its sequence number.

`parallel` exits with a code that corresponds to the number of jobs that failed. If any job failed, look in `makepkg_log` which jobs failed, find corresponding stdout and stderr files using the sequence number and print all of this to stderr.

In the end, print out the latest known versions. Even if one or two updates failed, the rest have been updated and can be used later in the pipeline. 

image

Fixes #286

### DIFF
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -28,6 +28,12 @@ usage() {
     exit 1
 }
 
+trap_exit() {
+    if ! [[ -o xtrace ]]; then
+        rm -rf "$tmp"
+    fi
+}
+
 if ((!$#)); then
     usage
 fi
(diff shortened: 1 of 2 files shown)


AladW:
Nice!

PR #348, opened 2018-04-08
aur-build: Sync pacman database after building everything in chroot

maximbaz:
I build everything in chroot, and I noticed that I _always_ need to sync pacman db afterwards, every single time. I have not seen a case yet when I would not want to sync pacman db after the build.

When building _not_ in chroot, I believe pacman db is being synced after building every package. When building in chroot, it is enough to sync it once after all packages are built.

P.S. I think it's pretty safe to add `/usr/bin/pacman -Sy` in `/etc/sudoers` to not require sudo password for the sync - because there is no `*` in the end, commands like `/usr/bin/pacman -Sy evil` will still require password. I made this change to `man` page in a separate commit - if you disagree, I can remove that commit.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -182,6 +182,10 @@ while read -r -u "$fd" pkg _; do
     fi
 done
 
+if ((chroot)) && ! ((no_sync)); then
+    sudo pacman -Sy
+fi
+
 exec {fd}<&-
 
 # vim: set et sw=4 sts=4 ft=sh:
--- a/man1/aur-build.1
+++ b/man1/aur-build.1
@@ -93,7 +93,7 @@ prompts, \fBsudoers\fR(5) can be used instead.
 For example:
 .EX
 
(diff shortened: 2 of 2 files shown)


AladW:
That does a [partial upgrade](https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported) on the host system. As packages in a chroot are built against potentially newer dependencies, the only correct approach is `pacman -Syu` (when building on the host, an `-Syu` is sufficient on the local repo only), however I don't want to impose a full system upgrade in aurutils.

maximbaz:
Interesting, so the issue is not with with `pacman -Sy` per-se (because I'm not updating anything with this command), but with people incorrectly using `pacman -S newpkg` afterwards instead of upgrading their system before installing a new package?

AladW:
Unlike other distributions Arch packages contain very little information on what packages they depend on (in particular, what .so versions). So it's easy to hose your system if you e.g. have an old openssl available but packages in a new database (retrieved from -Sy) assume a new one.

PR #361, opened 2018-04-18
aur-deps-src: rewrite in awk

AladW:
I went ahead and rewrote `aur-deps-src` in gawk. Note that I've only tested this for a small amount of packages. @wheatdog

---
P.S. I'm considering to rename it back to `aur-queue` or such, since the script works quite differently from `aur-deps-rpc`. (for one it outputs pkgbase instead of pkgname)


Earnestly:
AFAIR in awk you would use two arrays keyed on the same values.  Best to check http://awk.freeshell.org/ which is maintained by @e36freak who also hangs around on #archlinux.  They're the best I know

AladW:
Thanks! Feel free to submit a patch if you like.

e36freak:
I'm not feeling fighting with git on figuring out how to make this work, so here's your sauce:

http://ix.io/18fv

Enjoy

e36freak:
Totally forgot I wrote a thing for this, see "ssplit()" in https://github.com/e36freak/awk-libs/blob/master/strings.awk though it's probably still not as fast as gawk's native thingy

AladW:
>It may not be as slow as you expect

Yeah, it's probably not too relevant. I just checked on 70000 packages (AUR 3 mirror) and the real bottleneck is calling `vercmp` through the shell. Might as well use a more portable construct then. Thanks for the input.

PR #368, opened 2018-04-22
 aur-sync: Allow reading package names from standard input 

CyberShadow:
For when you want the latest commit of *everything*.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -88,7 +88,7 @@ trap_exit() {
 }
 
 usage() {
-    plain "usage: $argv0 [-ABcDfglLprstTu] [long options] [--] pkgname..."
+    plain "usage: $argv0 [-ABcDfglLprstTu] [long options] [--] pkgname... [-]"
     exit 1
 }
 
@@ -274,9 +274,12 @@ if ((list)); then
     exit
 fi
 
-{ if (($#)); then
-      printf -- '%s\n' "$@"
-  fi
+{ for arg in "$@"; do
+      case $arg in
+          -) cat ;;
+          *) printf -- '%s\n' "$arg" ;;
+      esac
+  done
 
   if ((update)); then
       aur vercmp <db_info | cut -d: -f1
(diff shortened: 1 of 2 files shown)


maximbaz:
Maybe you are looking for #283?

AladW:
Since `pkgname` cannot contain spaces or tabs, you could use `aur-sync --list`
```
aur sync $(aur sync --list | jq .pkgname)
```
I guess you could add something like this to take packages from stdin (newline separated, as `pacman`)
```
for arg in "$@"; do
  case $arg in
    -) stdin=1 ;;
  esac
done

if ((stdin)); then
  mapfile -t pkg_stdin
  set -- "$@" "${pkg_stdin[@]}"
fi
```
so the above transforms to:
```
aur sync --repo-list | jq .name | aur sync -
```

CyberShadow:
Ahh, thanks! It did seem too easy.

> That said, you can already achieve the above with command substitution

That works! Though I think you meant ` | jq -r '.[][].pkgname'`.

> I guess you could add something like this to take packages from stdin

OK, I turned this PR into a change that implements it. How's this?

AladW:
Nice, I like the `cat` trick. :+1:

AladW:
Hmm, I wonder what happens if you specify `-` twice?

CyberShadow:
The second cat hits EOF immediately, so it's a no-op.

```shell
$ echo test | ( cat ; cat )
test
```

AladW:
All right, thanks for clarifying.

AladW:
Not sure it works completely as intended: https://github.com/jrudess/aurutils-plugins/blob/master/lib/aur-update-devel#L20-L23

Earnestly:
@AladW fwiw, you can `declare -l` the `reponse` variable to avoid needing to do any case matching.

```
help declare
...
      -l	to convert NAMEs to lower case on assignment
```

CyberShadow:
One possibility is that aurutils could read interactive responses from the TTY directly, like `sudo`. That would however mean that `yes` and similar programs become unusable (at least not without being combined with `script` or such).

Earnestly:
Indeed, please don't ever do that.

AladW:
I guess a better approach is to instead add `-a` to aur-sync. Then you could do
```
aur sync --no-ver-shallow -a <(cut -d: -f1 updates)
```
Not sure if there's still much benefit in `-` then, though programs like `xargs` support both `-a` and `-`.

Issue #317, opened 2018-03-01
aur-sync: how to manage cache? (remove sources, clean logs, etc.)


maximbaz:
I'm sure this has to be possible, but I can't find the way in the documentation.

The size of my `~/.cache/aurutils/sync` is `3 GB` right now, and I don't want it to take that much. I'm happy to keep PKGBUILD and .SRCINFO cached, but in the cache I also see the sources (downloaded tar archives, or `src/` folder for `-git` packages), lots of log files (package build logs, namcap logs, pkgver logs), and I don't really want all those in my cache.

I am normally using `aur sync -sc <pkgname>`.

AladW:
https://bbs.archlinux.org/viewtopic.php?pid=1753269#p1753269

maximbaz:
Would you consider adding a flag for `aur-sync` to cleanup the cache of just built packages? Seems like a pretty useful feature, unless I'm missing the benefits of having sources kept forever in the cache?

maximbaz:
It seems it's best to keep `src/` and `<reponame>/` directories in `~/.cache/aurutils/sync/*-git` folders.

First of all, `aur srcver` will take longer due to re-cloning the repo (well, until we get `aur out-of-date` ready). But perhaps having the repo available will help `aur-sync`ing VCS repos faster too.

For regular AUR packages however, that download new tarballs with every version update, I don't see any benefit of keeping the sources in cache.

AladW:
Tasks that involve removing files in the user's home directory, even if in `.cache` by default are something I'd rather leave to the user. If something goes wrong they only have themselves to blame, and not `aursync`. :wink:

I'm not sure what you mean by your second comment. Note that if you don't want to keep things cached in the first place, you can use `makepkg -C`. (this could be an `aur-sync` parameter, with `aur-build` you just pass it after `--` as usual)

maximbaz:
Hehe, I see 🙂

Choosing not to cache things in the first place would be acceptable, and yes it works for `aur-build` (I only have to provide `--` twice, because I build in chroot). Let's add this flag to `aur-sync` and that's all (I think you meant lowercase `-c`, or `--clean`, by the way).

AladW:
`makepkg` has both a `-c` and a `-C` flag, from `makepkg(8)`:
```
 -c, --clean

    Clean up leftover work files and directories after a successful build.

 -C, --cleanbuild

    Remove the $srcdir before building the package.
```
[...]

Issue #305, opened 2018-02-20
Add "aur remove"


terencode:
Description: Remove a package from a repository database.
Usage: <code>aur remove <i>package</i> [--repository=all]</code>

This would have 3 flags:
- <code>--repo=<i>repo</i></code>: The repository you want the package to be removed from.
- `--uninstall`: Also run <code>sudo pacman -R <i>package</i></code>
- `--no-cache`: Also remove all the built packages versions of the package.

maximbaz:
IMHO this feels "too simple" to be part of aurutils, it's just 3 calls with no extra logic, you can make your own function/alias:

```
repo-remove $repo $package
pacman -R $package
paccache -r
```

Earnestly:
@maximbaz Instead they can encapsulate it in a script which is then placed in their `PATH` as `aur-remove` to gain the functionality they want without having to involve upstream at all.  Just like how git works.

maximbaz:
Oh yes, precisely, I forgot about this neat feature 🙂

terencode:
@Earnestly Would it be possible to have an example/ link to documentation?

terencode:
@maximbaz I agree, but what if you want to remove the package in all the the databases and repository?
This makes it repetitive even with an alias/function and thus should be automated.

terencode:
@AladW Ok thanks, I'll use that once I know how to add custom scripts.

maximbaz:
https://github.com/AladW/aurutils/pull/289

```
mkdir ~/bin
export PATH=~/bin:$PATH
echo '#!/usr/bin/env bash\n\necho "I''m called from aur!"' > ~/bin/aur-bla
chmod +x ~/bin/aur-bla

aur bla
```

terencode:
@maximbaz my version: 

``
#!/usr/bin/env bash

package="$1";
find /home/terence/.cache/pacman/ -name '*.db.tar' -execdir repo-remove {} "$package" \;
sudo pacman -R "$package";
``

maximbaz:
> Is it really not worth adding?

That's just my opinion, decide together with @AladW 🙂 

Do you have some specifics, which part will not work all the time? Maybe we can give you a quick hint on how to solve this.

Foxboron:
@AladW Should we add this to the documentation? Under what manpage?

AladW:
[...]

PR #372, opened 2018-04-24
Create update-devel

AladW:
This example script is based on `aur-sync` changes proposed in #283. 

It is a simple reuse of `aurutils` components, similar to the examples in `aur(1)`. As such it is part of a new `examples` directory. This directory could include other scripts such as `aur-remove` (#305) and `aur-gc` (#317).

To use a script from the examples directory, it may be either copied to or symlinked in a directory in `PATH`. A symlink should be preferable unless local changes to the file are prefered.
```
$ cp -s /usr/share/aurutils/examples/aur-update-devel /usr/local/bin
$ aur update-devel
```

### DIFF
--- a/contrib/update-devel
+++ b/contrib/update-devel
@@ -4,10 +4,9 @@ readonly XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
 readonly AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/sync}
 readonly AURVCS=${AURVCS:-.*-(bzr|git|hg|svn)$}
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
-tmp=$(mktemp -d)
 
 db_tsv() {
-    jq '.[][] | [.pkgname, .pkgver] | @tsv'
+    jq -r '.[][] | [.pkgname, .pkgver] | @tsv'
 }
 
 db_join() {
@@ -18,8 +17,17 @@ ifne() {
     xargs -r "$@"
 }
 
-cd "$tmp"
+trap_exit() {
+    if [[ ! -o xtrace ]]; then
+        rm -rf "$tmp"
+    fi
+}
+
+tmp=$(mktemp -d)
+trap 'trap_exit' EXIT
+
 # get repo contents
+cd "$tmp"
 aur sync --list | db_tsv >db_info
 
 # checkout latest revision for existing pkgbuilds


maximbaz:
I have an idea, with this script I would need to run update twice (first `aur sync -u`, then `aur update-devel`), should we maybe make this script update everything? Or rename to `update` and have a `--devel` flag? What do you think?

AladW:
I thought it was simpler to run `aur sync -u` instead of adding a flag:  64955449d12313b4987289632738f55f3ff1c1af

AladW:
The `Makefile` should also be updated

maximbaz:
I wanted to avoid two calls to `aur sync`, so that I can review everything once, and then let it proceed...

maximbaz:
I'm also getting some weird `jq` parsing errors:

```
jq: error: syntax error, unexpected INVALID_CHARACTER (Unix shell quoting issues?) at <top-level>, line 1:
.[] | [.pkgname, .pkgver] | @ tsv                            
jq: 1 compile error
```

AladW:
Yeah, `aur-sync --upgrades --no-ver` should suffice. You just won't get pretty-printed versions.

Regarding `jq`, I feel about that software like [some people feel about `tar`](https://xkcd.com/1168/). I guess there's a very slight benefit in having `aur sync --list` output to JSON because repo names can contain arbitrary characters, but meh.

contrib/update-devel
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+set -e
+readonly XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
+readonly AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/$argv0}
```
maximbaz: `$argv0` -> `aursync`

contrib/update-devel
```diff
@@ -0,0 +1,31 @@
...
+readonly AURVCS=${AURVCS:-.*-(bzr|git|hg|svn)$}
+readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+tmp=$(mktemp -d)
+
+db_tsv() {
+    jq '.[][] | [.pkgname, .pkgver] | @ tsv'
```
maximbaz: `@tsv` should not have space in between

AladW:
I've thought on adding `$@`, yeah

AladW:
Hm, the question is whether to add "$@" to `aur-sync --list` as well, e.g. to specify `--repo`. I guess something like `aur sync --list -sc --provides` would be the same as `aur sync --list`...

maximbaz:
Yeah, I guess it's safe - I checked and `aur sync --list -sc --provides` works fine, and `--list` is always the first action that is being executed, so it should be safe to add `"$@"` to both places.

AladW:
Done. Left a code comment just in case

Issue #241, opened 2017-05-10
Should `aursync -f` include `--no-ver` by default?


monotykamary:
Using `aursync -f` on an existing package does nothing since the package still has to go through `chkver` before executing, and thus requires `aursync -f --no-ver`. All cases I can think of in `aursync` in which that a target build package already exists on the system and requires `--no-ver`, must also require `--force` - or I might be missing something.

AladW:
I've thought of that, but then you would also ignore the full dependency chain even if you only want to rebuild a single package... perhaps you could implement some "shallow" variant of `--no-ver`. Or simply use aurbuild directly (it can be called directly from a directory containing a PKGBUILD)

AladW:
Can anyone think of a use case where `-f` without `--nover` would also be useful? I guess `--rebuild` (`-f` and `chkver` for dependencies only, not command-line arguments) and `--rebuildtree` (`-f` and `--nover`) are descriptive, but they further add to option bloat in `aur-sync`. 

Maybe you could have `-f` have a side-effect on version checks and add a `--ver` flag to enable version checking again, should so be required.

PR #373, opened 2018-04-24
Add contrib/update to update both regular and VCS packages

maximbaz:
@AladW this is an alternative to #372 with some bugs fixed and `--devel` flag.

Detects everything first, then fires up `aur sync` and updates everything in one go.

### DIFF
--- a/contrib/update
+++ b/contrib/update
@@ -39,7 +39,7 @@ trap 'trap_exit' EXIT
 cd "$tmp"
 
 # get repo contents
-/usr/bin/aur sync --list | db_tsv >db_info
+aur sync --list | db_tsv >db_info
 
 # detect outdated packages according to AurJson
 aur vercmp <db_info | cut -d: -f1 >to_update
--- a/contrib/update
+++ b/contrib/update
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -e
 readonly XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
-readonly AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/$argv0}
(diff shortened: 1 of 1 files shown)


maximbaz:
This is still not as good as #283, because `aur-sync` detects dependency tree, and `--nover` updates them all. So for example if an outdated VCS package has an up-to-date VCS package as a dependency, even though `contrib/update` detects only _one_ package to update, `aur sync` will pull its dependency tree and offer to update _two_ packages.

> Because aur-sync will complain about an ambiguous repo choice.

Good point, we can add a flag for the repo name. However I want first to see if @AladW has some ideas how to solve the issue I described above, because otherwise this script is not as useful for me...

AladW:
Solving https://github.com/AladW/aurutils/issues/241 should solve the first issue

maximbaz:
You mean the proposed `--rebuild` flag, that will treat the provided packages with `nover`, but will check versions of their dependencies?

AladW:
At the risk of maximum unclarity, the following should work:
```
complement_if() {
    if (($1)); then
        complement "$2"
    else
        tee
    fi    
}

(...)

if ((chkver)); then
    # packages with equal or newer versions are taken as complement for
    # the queue. if chkver_shallow is enabled, packages on the command-line
    # are excluded from this complement.
    cut -f1,3 pkginfo | aur vercmp -p db_info -c | complement_if chkver_shallow argv
fi
```

maximbaz:
With which approach would you like to proceed, this or #372? I don't have a preference anymore, and there is less code to support in #372 🙂

AladW:
Well I like #372 better since it doesn't have an added switch. :wink: Anyway let me know if #374 works as intended.

maximbaz:
I'll close this one, test #374 and post there

PR #374, opened 2018-04-24
aur-sync: --nover-shallow

AladW:
#373 #241

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -19,7 +19,7 @@ makepkg_args=(-cs)
 build=1 chkver=1 download=1 view=1
 
 # default options (disabled)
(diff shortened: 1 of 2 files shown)


lib/aur-sync
```diff
@@ -192,6 +201,10 @@ while true; do
         --nover|--no-ver)
             chkver=0
             shift ;;
+        --nover-shallow|--no-ver-shallow)
+            chkver=1
```
maximbaz: `chkver` is `1` by default anyway

AladW: Yeah, but it leaves out any ambiguity this way. 

In general, I've been wondering if we should do like `git` and have positive flags, even for options that are enabled by default...

maximbaz:
#372 together with this seems to work fine, I'll be testing some more over time 👍

PR #375, opened 2018-04-24
contrib: Extract aur-vercmp-devel

maximbaz:
I realized that I would actually benefit more from having a script that _detects_ outdated VCS packages, and extracted it into `aur-vercmp-devel`. The `aur-update-devel` becomes a one-liner, I still find it useful, but I'm not sure if you are OK with keeping it in the contrib 🙂


AladW:
I like `aur-vercmp-devel` since it has one strictly defined task. I'd remove `aur-update-devel` from contrib then and maybe add something to the README (see https://github.com/AladW/aurutils/commit/95bcd0bbd606a858fafdb5c35823202c1bcf5861)

maximbaz:
Done 👍

PR #377, opened 2018-04-25
Remove -r flag to allow updating when no VCS pkgs are outdated

maximbaz:
Just a small typo, `-r` prevents running normal `aur sync -u` if all VCS packages are up-to-date.

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -14,5 +14,5 @@ You can build upon these scripts to fit your own workflow. For example, you can 
 
 # sync VCS packages that are outdated according to aur-vercmp-devel
 # sync all packages that are outdated according to AurJson (--upgrades)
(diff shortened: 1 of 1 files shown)


AladW:
I did a longer writeup in the new README: https://github.com/AladW/aurutils/tree/master/contrib#aur-vercmp-devel

maximbaz:
That's fair, you may want to add the example with `--upgrades` to showcase that both regular and VCS packages can be updated in one go, because it somehow wasn't obvious to me that I can combine `-u` with packages from stdin. Otherwise looks good 👍

AladW:
I have some doubts on the second pipeline I've added, since that will offer all (possibly non-git) dependencies for inspection as well, which is irrelevant when `aur srcver` is not run on those packages.

Also perhaps these longer comments better belong in the script itself

PR #381, opened 2018-04-25
aur-srcver: Handle errors with running parallel

### LINKED ISSUE #380: aur-srcver: show reasonable error message when parallel is missing
I symlinked contrib/aur-vercmp-devel to appear on my PATH. When I run it with `AUR_REPO=aur` set (so it checks for the aur repo), I get the following output:

```sh
$ aur vercmp-devel
  -> Using [aur] repository
srcver: failed to update 127 packages
/usr/lib/aurutils/aur-srcver: line 66: makepkg_log: No such file or directory
mpv-build-git: v0.28.0.360.g013a8f75f3-1 -> v0.28.0.487.g65f0825315-1
shaderc-git: r439.ad86c6c.glslang_r2511-1 -> r465.5fd1b25.glslang_r2706-1
```

I would expect it to check any of my devel packages on whether their upstream source has an update, which is definitely the case.

(Ignore the weird version numbers for now, they are an artifact of my merging of the two repos earlier.)

(It also ignores `-d reponame`, but that doesn't matter to me anymore since I merged my aur and aur-devel repos earlier.)

maximbaz:
Simple way to fix #380: print a message and exit if parallel exit code is >101.

Also adds a special check if `parallel` is not available, because it is a required tool and we want to be as clear about this as possible.

```
❯ lib/aur-srcver ~/.cache/aurutils/sync/*-git
srcver: "parallel" is not found in $PATH
```

### DIFF
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -38,10 +38,7 @@ if ((!$#)); then
     usage
 fi
 
-if ! type -P parallel >/dev/null; then
-    printf >&2 '%s: "parallel" is not found in $PATH\n' "$argv0"
-    exit 1
-fi
+command -V parallel >/dev/null || exit
 
 # XXX trickery for hyphen and absolute path arguments
 mapfile -t arg_path < <(readlink -e -- "$@")
@@ -56,14 +53,14 @@ exitcode=$?
 
 if ((exitcode > 101)); then
     printf >&2 '%s: error running "parallel", exit code %d\n' "$argv0" "$exitcode"
-    exit $exitcode
+    exit "$exitcode"
 fi
 
 if ((exitcode > 0 )); then
     printf >&2 '%s: failed to update %s packages\n' "$argv0" "$exitcode"
 
-    while IFS=$'\t' read -r seq _ _ _ _ _ exitcode _ command; do
-        if ((!exitcode)); then
+    while IFS=$'\t' read -r seq _ _ _ _ _ jobexitcode _ command; do
+        if ((!jobexitcode)); then
             continue
         else
             printf >&2 '8<----\n'
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -70,6 +70,8 @@ if ((exitcode > 0 )); then
             cat "${seq}_makepkg.err" >&2
         fi
     done < makepkg_log
+
+    exit "$exitcode"
 fi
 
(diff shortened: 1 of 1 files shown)


lib/aur-srcver
```diff
@@ -38,6 +38,11 @@ if ((!$#)); then
     usage
 fi
 
+if ! type -P parallel >/dev/null; then
```
AladW: Easier ist to use `command -V parallel >/dev/null || exit`:
```
$ ~/aurutils/lib$ ./aur-srcver *
./aur-srcver: line 42: command: parallel: not found
```

lib/aur-srcver
```diff
@@ -47,10 +52,15 @@ trap 'trap_exit' EXIT
 cd "$tmp" || exit
 parallel --will-cite --nice 10 -j +2 --joblog makepkg_log --results '{#}_makepkg' \
   'cd {}; makepkg --skipinteg --noprepare -od' >/dev/null 2>&1 ::: "${arg_path[@]}"
-num_failed=$?
+exitcode=$?
+
+if ((exitcode > 101)); then
+    printf >&2 '%s: error running "parallel", exit code %d\n' "$argv0" "$exitcode"
+    exit $exitcode
```
AladW: quoted in line above, but not here (doesn't really matter for integers, but might as well be uniform)

AladW:
:+1:

AladW:
In hindsight, what was the reasoning for https://github.com/AladW/aurutils/pull/381/commits/532699d4f9e6d918cd61c6cca53a5f38bd330951? If a job was successful, you might as well print the new version, and if `parallel` itself fails, we have already bailed.

maximbaz:
Half of the jobs could fail, and previously we printed lots of errors, and then full list of versions, half of which were up-to-date and another half is obsolete. So I thought, maybe let's print versions only if there are no failed jobs, to have a consistent output... Now if versions are printed, you can be sure that these are truly the latest versions.

We could have also made the script to print versions of only successful jobs and skip failed jobs, but I'm not sure if we need that complexity at all - we would likely want people to solve their issues with `srcver` first anyway, right?

AladW:
>We could have also made the script to print versions of only successful jobs and skip failed jobs, but I'm not sure if we need that complexity at all - we would likely want people to solve their issues with srcver first anyway, right?

Yeah, you're probably right. Even it this were desired, I don't see a straighforward way to get the job arguments from `parallel`

PR #383, opened 2018-04-30
aur-depends: remove unrelated functionality

AladW:
To better reflect the name (and more generally, to reduce the scope of the tool), this
commit removes most processing of the retrieved dependency information from the AUR.
In particular, if a total order is required, this must be done manually through cut(1)
and tsort(1).

Usage in aur-fetch remains similar, apart that pkgbase must be used when retrieving
build files (this was an oversight when recursion was introduced with commit
3bd2fcbcacdb939cbc782a99702580db35cdc24b).

Adapting aur-sync(1) is a more challenging task, and is not fully covered by this
commit.

### DIFF
--- a/lib/aur-depends
+++ b/lib/aur-depends
@@ -1,27 +1,11 @@
 #!/bin/bash
 # aur-depends - retrieve package dependencies using AurJson
-set -o noclobber
 readonly argv0=depends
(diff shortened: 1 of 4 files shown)


AladW:
https://lists.archlinux.org/pipermail/pacman-dev/2018-April/022450.html :tada:

AladW:
With https://github.com/AladW/aurutils/pull/383/commits/b2ad51b6fadc4e38b166c74dc49b5ab74650472f this should be ready for testing

Issue #379, opened 2018-04-25
aur-sync: differentiate between new PKGBUILDs and those without changes


FichteFoll:
I'm using `ranger` as my AUR_PAGER, though I'm not sure whether this might affect vifm too. When I sync a package that exists in the sync cache, vifm used to show a patch for any commits to the aur git repo since the last sync, in addition to the entire folder. When there were no changes, the patch was just empty.

Now, however, empty patches don't show up at all and I can't tell whether I have reviewed a PKGBUILD already or not when it didn't have any changes.

AladW:
Isn't this the same behavior as in the release version (1.5)? https://github.com/AladW/aurutils/blob/master/lib/aur-fetch-git#L14

FichteFoll:
I've never used the release version, so I wouldn't know. I do recall that I was seeing empty patches with size 0 with vifm before AUR_PAGER was implemented, though.

In the end, the only thing I want is that it doesn't look the same to me for PKGBUILDs I have already viewed and those that didn't have any changes. You could also omit unchanged PKGBUILDs from the review entirely.

AladW:
Since I prefer to view non-empty patches, I'll see what I can do on omitting unchanged PKGBUILDs. It's similar to what I'm trying to achieve here: https://github.com/AladW/aurutils/blame/master/contrib/README.md#L57-L59

rafasc:
In my opinion the patches are a aurutils courtesy. They're nice to have but aren't part of the build per se. So I do not have a strong opinion if they're present or not. but I disagree with omitting build files, even if they're unchanged.

That being said, they are a good indicator of what needs review. One option we could use is to always produce a "patch".

But what should happen when a new package is built? A diff against the empty tree (diff against 4b825dc642cb6eb9a060e54bf8d69288fbee4904) could work, but would be a bit redundant. 

Maybe having a file explicitly saying "this package is new to aurutils, please review, etc"?

AladW:
I guess the sledgehammer approach would be `git-subtree`. "New" packages would be recorded as commits, and these could be saved to some patch file.

rafasc:
A much simpler option would be to store something inside the repo to mark it as seen.

Like adding a ref named AURUTILS which points to exact commit the repo was added.

``` sh
# create ref
$ git update-ref AURUTILS $(git rev-parse HEAD) 
# check if exists
$ git rev-parse --verify -q AURUTILS >/dev/null && echo seen || echo new 
```

[...]

PR #390, opened 2018-05-13
aur-sync: Show top-level patches first in vifm

maximbaz:
A while ago patches were shown on top, now they are mixed with the rest of the files. It is useful to have the patches shown on top.

If we have markers for new packages, we should use similar technique to show them on top (vifm is not sorting files alphabetically, so instead of naming it `!new.pkgname` we should just find and print these markers first.

`vifm` is smart enough to ignore duplicate entries, that's why the last `find` call can be left as it is.

------------

`master`:

aurutils-7

This PR:

aurutils-8


AladW:
Some experimenting with `bfs` and `tree`:

https://asciinema.org/a/f9FMNkllU90BGvGVidAJookgp

rafasc:
I thought `find -L . -maxdepth 2 -mindepth 1 -depth | tac | vifm - '+view!'` could be a solution, but apparently we cannot rely on `find`'s output order because the order of files may change depending on the order files are stored in the filesystem.

In our case, since we always generate the .patch files after the symbolic links, it "works" but it's not guaranteed. In short, it seems it's hard to show the .patch first consistently with a simple call to `find`. 😞

AladW:
On the other hand, in the old 1.5 implementation `find` was called once for every package in the queue, so ~250 times for stuff like ROS! I think my main objection is to hardcode a `.patch` or `.diff` extension rather than use some pre-defined directory to store them.

As an aside, I wonder if `vifm` has means to "save" a current view. Then use of `tmp_view` would not be required if you want to change directories and return to the build files for the current queue.

AladW:
Also note there's nothing preventing you from calling a package `foo.patch` or `foo.diff` :disappointed:

maximbaz:
> Note there's nothing preventing you from calling a package foo.patch or foo.diff 😞

`-type f` solves that 🙂 

Still hardcoded extensions and still two `find` calls, but it works:

```
(find -L . -maxdepth 1 -type f -name '*.patch' -o -name '*.diff'; find -L . -maxdepth 2 -mindepth 1) | vifm - '+view!'
```

rafasc:
> I wonder if vifm has means to "save" a current view.

>Custom list exists only while it's visible, once left one can't return to it, so there is no appearances of it in any history.

 From what's said on the vifm manpage, I think there isn't a way to save custom views.


> Also note there's nothing preventing you from calling a package foo.patch or foo.diff

This opens a new can of worms. Because having .patch files on `$tmp_dir` can technically conflict with other packages. 

Installing packages: `foo` and `foo.diff` would result in a clash between the diff from `foo` and the directory from `foo.diff`.

AladW:
Maybe we can create some subdirectory with a marker like `!patch` (cf. the proposed `!new`)? Then you'd have:
```
find -P '!patch'
find -L . -maxdepth 2 -mindepth 1 ! -ipath './!*'
```

maximbaz:
Looks fine to me. `-P` is the default behavior according to `man`, we can omit it, and there's no need to complicate the second `find` with `! -ipath` because `vifm` is smart to ignore duplicates.

So this works well:

```
(find '!patch'; find -L . -maxdepth 2 -mindepth 1) | vifm - '+view!'
```

rafasc:
If we're using the `!` prefix (which is not allowed on package names) we can also keep the patches in the root of `$tmp_dir` if we write the patches with that prefix.

Files on the root of `$tmp_dir`with `!` prefix are shown first. And while `vifm` deals with duplicates, this would be a fine feature to have in the pager fallback. 
So I propose doing something like this:


```diff
diff --git a/lib/aur-fetch-git b/lib/aur-fetch-git
index 543523a..a3c1bda 100755
--- a/lib/aur-fetch-git
+++ b/lib/aur-fetch-git
@@ -44 +44 @@ case $log in
-    directory) merge() { merge_across_fs "$1" "$log_dir/$1".patch; } ;;
+    directory) merge() { merge_across_fs "$1" "$log_dir/!$1".patch; } ;;
diff --git a/lib/aur-fetch-snapshot b/lib/aur-fetch-snapshot
index d950524..d6c3723 100755
--- a/lib/aur-fetch-snapshot
+++ b/lib/aur-fetch-snapshot
@@ -53 +53 @@ case $log in
-    directory) diff_log() { tee -a "$log_dir/$1".diff | diffstat -Ckq -f3; } ;;
+    directory) diff_log() { tee -a "$log_dir/!$1".diff | diffstat -Ckq -f3; } ;;
diff --git a/lib/aur-sync b/lib/aur-sync
index 1ebf207..8ab1c77 100755
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -81,0 +82,5 @@ complement() {
+list_bang_first() {
+    find -L "$1" -maxdepth 2 -mindepth 1 -ipath "$1/!*"
+    find -L "$1" -maxdepth 2 -mindepth 1 ! -ipath "$1/!*"
+}
+
@@ -348 +353 @@ if ((view)); then
-        (find -L . -maxdepth 1 -name '*.patch'; find -L . -maxdepth 2 -mindepth 1) | vifm - '+view!'
+        list_bang_first "." | vifm - '+view!'
@@ -350 +355 @@ if ((view)); then
-        find -L "$tmp_view" -maxdepth 2 | command -- ${PAGER:-less}
+        list_bang_first "$tmp_view" | command -- ${PAGER:-less}
```

lib/aur-sync
```diff
@@ -79,6 +79,12 @@ complement() {
     grep -Fxvf "$@" || return $(($?-1))
 }
 
+list_markers_then_files() {
+    find -L "$1" -maxdepth 1 -mindepth 1 -name '!*.new'
+    find -L "$1" -maxdepth 1 -mindepth 1 -name '!*.patch' -o -name '!*.diff'
+    find -L "$1" -maxdepth 2 -mindepth 1 ! -ipath "$1/!*"
```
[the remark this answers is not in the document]
maximbaz: Valid point, let's play nice and exclude marker files by extension too.

lib/aur-sync
```diff
@@ -82,7 +82,7 @@ complement() {
 list_markers_then_files() {
     find -L "$1" -maxdepth 1 -mindepth 1 -name '!*.new'
     find -L "$1" -maxdepth 1 -mindepth 1 -name '!*.patch' -o -name '!*.diff'
-    find -L "$1" -maxdepth 2 -mindepth 1 ! \( -ipath "$1/!*.new" -o -ipath "$1/!*.patch" -o -ipath "$1/!*.diff" \)
+    find -L "$1" -maxdepth 2 -mindepth 1 -not \( -ipath "$1/\!*.new" -o -ipath "$1/\!*.patch" -o -ipath "$1/\!*.diff" \)
```
rafasc: another small note, `-not` and `-ipath` are not POSIX.
They could be replaced by `!` and `-path`. 
@AladW what's your position on this?

maximbaz:
This is nice too! Is `-u` there just in case, or it is needed for some cases?

Also, do you want to put all patches in `!patch` folder, or keep them on the root level? Either is fine for me.

AladW:
Yeah, you can leave out `-u` - it's a remainder from a previous experiment.

I'm mostly undecided on whether to have a separate dir or separate files - though I'm not convinced the (admittedly rather unintuitive) `!` convention in "lower" programs like `aur-fetch-git` where it has less relevance.

maximbaz:
Just as another idea, we could make a more explicit folder structure, and it won't require using special characters like `!`. Example:

```
./new
./new/a
./new/b
./patches
./patches/c.patch
./patches/d.diff
./packages
./packages/a
./packages/a/PKGBUILD
./packages/b
./packages/b/PKGBUILD
./packages/c
./packages/c/PKGBUILD
./packages/d
./packages/d/PKGBUILD
```

We could keep `*.diff` and `*.patch` files in different folders if needed, but I think they have the same purpose and won't ever exist at the same time anyway...

AladW:
Sounds like a much better approach than relying on clever tricks like LC_ALL=C. I guess it means more navigation for non-`vifm` users but that's their concern, not mine. :slightly_smiling_face:

AladW:
One thing, I'd perhaps use shorter names like `pkg` for better viewing on small screens/terminals

maximbaz:
Done, short folder names are `new`, `diff` and `pkg`. Currently they are created beforehand, not on demand, as I was not sure how and where to nicely plug such logic in and whether it is actually needed.

maximbaz:
An example of how this currently looks in real world with `aur sync -u`:

image

FichteFoll:
I'm indifferent on empty `new` and `diff` folders as well. I can configure `ranger` to flatten the tree while reviewing, otherwise it will just show that the directories have zero entries.

I like this direction a lot. :+1: 

@maximbaz you can create multiple directories with a single `mkdir` call.

AladW:
I was wondering whether `new` could be a file that contains differences for directories in `AURDEST`. When cloning a new dependency tree, a lot of duplicates would appear in both `new` and `pkg`.

@rafsc this approach might also solve what you mentioned in https://github.com/AladW/aurutils/issues/379#issuecomment-385984987. We could write some empty file `OK` or `CONFIRM` or such, and check for its existence after the editor exists. If it is not present (i.e. deleted by the user), we assume the viewing was not "approved" and abort the build process.

AladW:
>I didn't fully understand the idea with new being a file that contains differences for directories in AURDEST.

You could do `find -type d` on `AURDEST` before and after using `aur-fetch`, then save the differences to the `new` file. Whether this is an effective approach is a different matter... maybe instead an additional flag to `aur-fetch` could be added where new files are moved to. That could fit the suggestion of separating packages to `new` and `pkg`.

>Selfish me prefers relying on exit code

One doesn't exclude the other, it's just an additional check if the editor exited with code 0.

maximbaz:
Got it, thanks 👍

We should probably tackle filling up the `new` folder in a different PR, to keep changes simple, what do you think? And how do you feel about the current approach so far for keeping patches on top, ready for testing and merging? 😉

FichteFoll:
>someone could make a feature request for `ranger` to implement the same

I certainly will, if I don't implement it myself. Should be much easier than adding support for stdin listing.
Checking for existance of key files seems like a decent compatibility layer for file managers that don't support that, however.

>I was wondering whether new could be a file that contains differences for directories in AURDEST. When cloning a new dependency tree, a lot of duplicates would appear in both new and pkg.

>One way we could go with new is just to create (empty?) files, whose names indicate which packages in pkg are new.

That would make reviewing new packages more complex, since you need to check for the names of new packages first and then search for them in the pkg list. Symlinking would definitely be preferrable.

lib/aur-sync
```diff
@@ -118,6 +118,8 @@ tmp=$(mktemp -dt "$argv0".XXXXXXXX)
 tmp_view=$(mktemp -dt view.XXXXXXXX)
 trap 'trap_exit' EXIT
 
+mkdir -p "$tmp_view/diff" "$tmp_view/new" "$tmp_view/pkg"
```
AladW: More compact: `mkdir -p "$tmp_view"/{diff,new,pkg}`

I've thought on some other names: 
* diff__s__ (compare `refs` in .git, and indeed multiple `.diff` files are stored there)
* queue (the files symlinked there reflect the build queue)
  * possible confusion when reading the code with `"$tmp"/queue` which is a regular file...
 
Since `new` is not yet implemented perhaps we can leave the reference (and according name) for a different PR.

lib/aur-sync
```diff
@@ -337,17 +339,17 @@ if ((download)); then
...
     cd_safe "$tmp_view"
 
     if [[ -v AUR_PAGER ]]; then
         command -- $AUR_PAGER
     elif type -P vifm >/dev/null; then
-        find -L . -maxdepth 2 -mindepth 1 | vifm - '+view!'
+        find -L diff new pkg -maxdepth 2 | vifm - '+view!'
     else
-        find -L "$tmp_view" -maxdepth 2 | command -- ${PAGER:-less}
+        find -L "$tmp_view/diff" "$tmp_view/new" "$tmp_view/pkg" -maxdepth 2 | command -- ${PAGER:-less}
```
AladW: as above: `"$tmp_view"/{...}`

lib/aur-sync
maximbaz: I like the `diffs`, not so keep on `queue`, maybe `pkgs` (also plural?), will remove `new`

AladW:
I did a basic review. However I have bunch more of ideas (such as using `xe` for cleaner pipelines and `git-ls-files` to not view irrelevant files) so I'd like to delay this a bit longer.

maximbaz:
Sure 👍 And if that's easier, feel free to pick pieces of this PR and commit them yourself as you go through your ideas, as long as the functionality eventually ends up in `master` I am happy 😉

AladW:
For example `xe` makes listing the build directories with a single `find` invocation trivial, without hardcoding any extensions or other tricks:
```
# unlike xargs, -I{} expands to multiple arguments!
xe -f "$tmp"/queue -I{} find -L {} -maxdepth 1
```

maximbaz:
a friendly ping on this bug @AladW 😋

AladW:
It's definitely still on the program. I'll be away for the next few days, I'll take a look after

AladW:
There's a very simple solution that keeps the patches in the top-level directory, if we go against the taboo of parsing `ls`:
```
ls -1r --escape --group-directories-first | tac | vifm -
```

AladW:
Nevermind, that would have no way to recurse with a given depth.

AladW:
For now, how about we go back to the 1.5 approach? The chance of packages ending in .patch is pretty slim, and we can always decide on the subdirectories when solving #379.

maximbaz:
I'm fine with this

maximbaz:
@AladW because of the tree view in `vifm`, patch files are now no longer at the top again 😉

Issue #391, opened 2018-05-19
Permeate parallel usage


AladW:
Right now `parallel` is an optional dependency for `aur-srcver`, but I think there is much to gain in using it in other scripts as well. In particular it is vastly more flexible than `xargs`, e.g. `{}` can be expanded to several arguments, and can easily be used in combination with `wget -i -` for threaded downloads.

I had imported `xe` as a submodule with this in mind, but `parallel` already has all its features, a longer trackrecord and does not require additional compiling. It is also in the community repo (only depending on `perl` and `procps-ng` which are installed on 100% of Arch systems according to pkgstats).

The only (practical) downside compared to `xe` I can think of is that by default, a `.parallel` dotfile is created. This can however be moved elsewhere with `PARALLEL_HOME` introduced with https://git.savannah.gnu.org/cgit/parallel.git/commit/?id=685018f532f4e2d24b84eb28d5de3d759f0d1af1

----
In summary:
* **xe** - Pros
  * Has basic `parallel` abilities including line buffering of output
  * Better interface than `xargs`, such as `-I{}` with multiple arguments and default new-line separation
  * Good performance
* **xe** - Cons
  * Not in the repos, either dependency on AUR package or import as submodule
  * Relatively new project, will probably find bugs during usage
* **parallel** - Pros
  * Highly flexible with many options for transforming command lines
  * Long-established project
* **parallel** - Cons
  * Requires use of `--will-cite` to not spam an ad on donating 10k EUR to the author (alternative: single use of `printf 'will cite\n' | parallel --bibtex`)
  * Minimal startup time of 0.10 seconds
  * Writes ~/.parallel unless `PARALLEL_HOME` is set

rafasc:
Based that alone, I don't see a strong argument to use `xe`.

AladW:
There's one other small thing against parallel, in that it takes 0.10 seconds to start. I guess this is negligible.

AladW:
Also having to use `--will-cite` in every invocation gets pretty annoying. I guess you could instead mention users to run:
```
printf 'will cite\n' | parallel --bibtex
```

AladW:
I updated the original post with a summary of pro/cons. I tend to go towards `xe`, but it's annoying for it to be an AUR package (and it would hardly justified bringing it to the repos with 0 votes and `parallel` as competitor).

maximbaz:
Personally I don't see either mentioned cons of parallel being an issue, to be honest.

[...]

PR #394, opened 2018-05-19
Use parallel instead of xargs (#391)

AladW:
Solves #391

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -336,11 +336,11 @@ else
 fi
 
 if ((download)); then
-    xargs -a "$tmp"/queue aur fetch "${fetch_args[@]}"
+    parallel -X aur fetch "${fetch_args[@]}" :::: "$tmp"/queue
 fi
 
 # link build files in the queue (absolute links)
-xargs -a "$tmp"/queue printf "$(pwd -P)/%s\\0" | xargs -0 ln -st "$tmp_view"
+parallel -Xj1 ln -s "$(pwd -P)"/{} "$tmp_view" :::: "$tmp"/queue
 
 if ((view)); then
     cd_safe "$tmp_view"
@@ -360,7 +360,7 @@ if ((build)); then
     build_args+=(-d "$repo" -r "$root" -a "$tmp"/queue)
 
     # check if dependency graph is valid
-    printf '%s\0' "$tmp_view"/*/.SRCINFO | xargs -0 aur graph >/dev/null
+    aur graph "$tmp_view"/*/.SRCINFO >/dev/null
 
     if ((chroot)); then
         aur build "${build_args[@]}" -- "${makechrootpkg_args[@]}" \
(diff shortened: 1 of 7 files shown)


lib/aur-sync
```diff
@@ -333,11 +333,11 @@ fi
 cd_safe "$aur_workdir"
 
 if ((download)); then
-    xargs -a "$tmp"/queue aur fetch "${fetch_args[@]}"
+    parallel -X aur fetch "${fetch_args[@]}" :::: "$tmp"/queue
```
AladW: I'm not sure why, but in the past I've seen that multiple jobs when cloning over SSH was problematic. Needs testing

lib/aur-sync
```diff
@@ -357,7 +357,7 @@ if ((build)); then
     build_args+=(-d "$repo" -r "$root" -a "$tmp"/queue)
 
     # check if dependency graph is valid
-    printf '%s\0' "$tmp_view"/*/.SRCINFO | xargs -0 aur graph >/dev/null
+    aur graph "$tmp_view"/*/.SRCINFO >/dev/null
```
AladW: I went with a simple glob here to avoid an additional parallel invocation... I guess we're pretty unlikely to reach ARG_MAX here anyway.

lib/aur-vercmp
```diff
@@ -58,18 +58,12 @@ my_vercmp() {
...
-    }'
-}
-
 parse_aur() {
     aur rpc -t info | jq -r '.results[] | [.Name, .Version] | @tsv'
 }
 
 parse_sync() {
-    pacman -Slq "$1" | prefix "$1" | xargs -r pacman -Sddp --print-format '%n %v'
+    pacman -Slq "$1" | parallel -rXj1 pacman -Sddp --print-format '%n %v' "$1"/{}
```
AladW: Note that `parallel` conveniently copies the "$1"/ prefix with `-X`, removing the need for a separate function.

makepkg/PKGBUILD
```diff
@@ -10,13 +10,11 @@ source=('git+https://github.com/AladW/aurutils')
 sha256sums=('SKIP')
 conflicts=('aurutils')
 provides=('aurutils')
-depends=('pacman>=5.0' 'git' 'jq' 'pacutils' 'wget')
+depends=('pacman>=5.0' 'git' 'jq' 'pacutils' 'wget' 'parallel')
```
AladW: A possible place to add the `will cite` notice could be an `.install` file. That or a forum post when 2.0.0 is announced. It seems out of scope to be include this in `aur(1)` or a different manual.

rafasc: Are you referring something like just a warning or actually doing 
`printf 'will cite\n' | parallel --citation` as part of the `.install`?

I am not a lawyer, but:
Based on how the aur works, the commands in `.install` are technically run by the user, so I guess it could be an option. Especially if we use a custom `$PARALLEL_HOME` for aurutils, in order to isolate aurutils use of the tool.

AladW: In this it would be a warning. If we wanted to run this automatically, it should be part of the prepare() function or such (since it needs to be run as the user running the PKGBUILD), though I guess this entails an unattended modification of the home directory and would thus be frowned upon.

I like the idea of using a custom `$PARALLEL_HOME`. Perhaps we could add a small wrapper to `lib` that sets it both to `~/.config/aurutils/parallel` and wraps any `parallel` commands with `--will-cite` (leaving a comment as to the licensing problematic)

AladW:
I added `aur-jobs` which does what https://github.com/AladW/aurutils/pull/394#discussion_r189447439 describes. Unless there's any weird bugs I don't know I think this can be closed.

PR #397, opened 2018-05-20
aur-chroot: allow using repos from specified pacman.conf

### LINKED ISSUE #396: aur-build: allow to use (local) repos from specified pacman.conf
https://bbs.archlinux.org/viewtopic.php?pid=1786410#p1786410

AladW:
Fixes #396


AladW:
I suspect this will break a whole bunch of stuff. *crosses fingers*

PR #400, opened 2018-05-22
aur-fetch: show diff to PKGBUILD first

### LINKED ISSUE #399: aur-sync: having .SRCINFO complicates previewing diffs in vifm
When I'm reviewing a diff, I'm mostly interested in looking at `PKGBUILD`, that there are no changes besides version bump.

`vifm` has a preview window that shows head of a selected file. Ideally I would be able to scroll through the `diffs` folder with `j`, without opening each patch in editor. But that requires that the relevant contents is shown in the preview section of `vifm` without making extra actions.

Currently `.SRCINFO` files are always shown on top of `*.patch` files, meaning that it is `.SRCINFO` file that is shown in the preview section of `vifm`. Here's an example of a `aur sync` in full screen on my laptop:

image

I can't be sure that `rm -rf /` is not added in the bottom of PKGBUILD unless I open this (and every other) patch in editor, or focus preview pane, scroll it down and focus back to files pane.

Who else is using `vifm`, what do you think about this issue? Basically I want to review 3 patches with `j`, `j`, `j` instead of `j<Enter>G:q<Enter>`, `j<Enter>G:q<Enter>`, `j<Enter>G:q<Enter>` or `j<C-w,l>G<C-w,h>`, `j<C-w,l>G<C-w,h>`, `j<C-w,l>G<C-w,h>`.

Solutions could be somehow removing `.SRCINFO` from patches, or making sure `PKGBUILD` diff is always on top of the `*.patch` file.

maximbaz:
This ended up being much easier than I expected.

Fixes #399

### DIFF
--- a/lib/aur-fetch-git
+++ b/lib/aur-fetch-git
@@ -12,7 +12,7 @@ merge_upstream() {
 
     if [[ $(git rev-parse HEAD) != $(git rev-parse '@{upstream}') ]]; then
         # only print log on upstream changes
-        git --no-pager log --patch --stat '..@{upstream}' >> "${1:-/dev/stdout}"
+        git --no-pager log -O <(echo 'PKGBUILD') --patch --stat '..@{upstream}' >> "${1:-/dev/stdout}"
 
         # discard any local changes (#349)
         git reset --hard HEAD@{upstream} >&2
--- a/lib/aur-fetch-git
+++ b/lib/aur-fetch-git
@@ -10,9 +10,15 @@ log=stdout
 merge_upstream() {
     git fetch -v >&2
 
+    # default to showing PKGBUILD diffs first (#399)
+    if [[ ! -s ${GIT_WORK_TREE:-$PWD}/git.orderfile ]]; then
+        echo PKGBUILD > "${GIT_WORK_TREE:-$PWD}"/git.orderfile
+    fi
+
     if [[ $(git rev-parse HEAD) != $(git rev-parse '@{upstream}') ]]; then
         # only print log on upstream changes
-        git --no-pager log -O <(echo 'PKGBUILD') --patch --stat '..@{upstream}' >> "${1:-/dev/stdout}"
+        git --no-pager log -O "${GIT_WORK_TREE:-$PWD}"/git.orderfile \
+            --patch --stat '..@{upstream}' >> "${1:-/dev/stdout}"
 
         # discard any local changes (#349)
         git reset --hard HEAD@{upstream} >&2
--- a/lib/aur-fetch-snapshot
+++ b/lib/aur-fetch-snapshot
@@ -67,6 +67,7 @@ for a in ./*.tar.gz; do
     x=${a%%.tar.gz}
 
     if tar_no_mode_diff "$a" "$startdir"; then
+        # FIXME show diff to PKGBUILD first, '-S' is weird (skips files), see #400
         diff -ur "$startdir/$x" "$x" | diff_log "$x"
     fi
 


AladW:
Nice! I do wonder why `-O` only takes files as argument...

maximbaz:
Indeed...

AladW:
It might also be a good idea to add this to `aur-fetch-snapshot`, see `diff -S`:
```
       -S, --starting-file=FILE
              start with FILE when comparing directories
```

maximbaz:
Nice 👍 However, there is an undocumented catch (or simply a bug) that I found in my tests: when I provide `-S`, the diff fails to find differences in hidden files and directories. Given the potential risk of missing important parts of diff, do you want to use `-S`?

AladW:
Wow, I can confirm this :disappointed: Sounds like a good candidate for [bug-diffutils](https://lists.gnu.org/mailman/listinfo/bug-diffutils)

maximbaz:
Actually, it looks like it is not about hidden files at all, regular files can be skipped too - it's all about an order. Apparently `diff` creates an ordered list of files to compare, and then `-S` just removes the head of the list, instead of placing the "starting file" to the top of the list.

I will still send the report, but I have a feeling that it might be "by design".

Example: all 4 files have different contents.

```
❯ tree -a
.
├── one
│   ├── dir
│   │   └── file
│   ├── .hidden-dir
│   │   └── file
│   ├── .hidden-file
│   └── zfile
└── two
    ├── dir
    │   └── file
    ├── .hidden-dir
    │   └── file
    ├── .hidden-file
    └── zfile

6 directories, 8 files

❯ diff -ur one two | grep '+++' | wc -l
4

❯ diff -ur -S zfile one two | grep '+++' | wc -l
1

❯ diff -ur -S zfile one two                     
diff --color --unified -ur -S zfile one/zfile two/zfile
--- one/zfile   2018-05-24 01:23:49.237899164 +0200
+++ two/zfile   2018-05-24 01:24:58.260920662 +0200
@@ -1 +1 @@
-one
+two
```

AladW:
Interesting, in that case a FIXME comment sounds appropriate enough.

maximbaz:
I wanted to add the link to my bug report to the FIXME comment, but the mailing list archive did not update for almost a month now, so no link.

maximbaz:
https://lists.gnu.org/archive/html/bug-diffutils/2018-05/msg00003.html

AladW:
Great, thanks for reporting this upstream!

I wonder on a slightly different approach, being to write `PKGBUILD` to a `git.orderfile` inside the package dir. That way, calling `git diff` or `git log` outside of `aur-fetch` would yield the same results. (If the user already has a git.orderfile, that would be overwritten; I suppose it could be checked and reused)

AladW:
Something like:
```
merge_upstream() {
    git fetch -v >&2

    # default to showing PKGBUILD diffs first
    if [[ ! -s git.orderfile ]]; then
        echo PKGBUILD > git.orderfile
    fi

    if [[ $(git rev-parse HEAD) != $(git rev-parse '@{upstream}') ]]; then
        # only print log on upstream changes
        git --no-pager log -O git.orderfile \
            --patch --stat '..@{upstream}' >> "${1:-/dev/stdout}"

        # discard any local changes (#349)
        git reset --hard HEAD@{upstream} >&2
    fi
}
```

maximbaz:
Sure 👍 

I only changed one thing in your snippet, `git ... -O ${GIT_WORK_TREE:-$PWD}/git.orderfile` (full path to `git.orderfile`, not assuming it is in the current folder).

AladW:
That's fine, though -O should looks at the top git dir anyway:
```
If diff.orderFile is a relative pathname, it is treated as relative to the top of the working tree.
```

maximbaz:
Up to you, I don't care much, this seemed more intuitive but I can change :)

AladW:
:+1: 

Probably need to document this at some point...

maximbaz:
Just updated 5 AUR packages, and had 5 `git.orderfile` files in `vifm`, the number grows linearly. These files take up vertical space and don't bring any value (diffs are already generated after all, there's no point in me reviewing or modifying `git.orderfile`). Are you sure you don't want to reconsider for approach in https://github.com/AladW/aurutils/pull/400/commits/99fb2b7487868e86454a35f112a75217b630959c that doesn't generate the noise?

AladW:
I just stuffed it in `.git`

PR #406, opened 2018-05-29
scripts: use util/parseopts.sh for option parsing

rafasc:
Pacman 5.1 provides a util script for option parsing that supports short
and long options.

In addition to provide easier to read options, this will also benefit
completion scripts that rely on parsing the output of `--help`.

--- 8< ---

I didn't have time to test this today, but wanted to put this out so we can have a dedicated place to discuss this.

134264b just converts existing use of `getopts` to `parseopts` without adding any extra functionality. Thus, `aur-{srcver,jobs,graph,depends}`<sup>1</sup> scripts remain without long option support.

I tried to follow the loop style used upstream where it uses a common `shift` for all options and the same line for the `pattern)` and `list ;;`. Saves a bunch of vertical space which makes it a bit nicer to read. But long `patterns)` kind mess up with the style<sup>2</sup>. (more evident on aur-sync for now)

Also kept just one space before the last command and `;;`. I always had mixed feelings about aligned `;;`, it looks prettier but adds noise to the history every time we need to adjust alignment.

Writing this I also noticed I have small inconsistencies on style. How strongly do you feel about this?

resolve before merging:
- [x] ~inconsistent quoting of `$1`~
- [x] ~add parseopts to scripts not covered by 134264b [1] (?)~
- [x] ~split long cases over multiple lines. [2] (?)~
  - ~split over multiple conditions and use cascading`;&`~
  - ~other or leave it as-is~

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -101,124 +101,70 @@ usage() {
 
 source /usr/share/makepkg/util/util.sh
 source /usr/share/makepkg/util/message.sh
+source /usr/share/makepkg/util/parseopts.sh
 
 if [[ -t 2 && ! -o xtrace ]]; then
     colorize
 fi
 
-if getopt -T || (($? != 4)); then
-    error "$argv0: util-linux getopt required"
-    exit 22
-fi
+opt_short='d:D:AcfglLnprstTu'
+opt_long=('allan' 'bind:' 'chroot' 'directory:' 'force' 'ignore:' 'ignorearch'
+          'ignore-arch' 'log' 'noconfirm' 'no-confirm' 'nover' 'no-ver' 'noview'
+          'no-view' 'print' 'rmdeps' 'rm-deps' 'sign' 'temp' 'tar' 'repo:'
+          'database:' 'root:' 'upgrades' 'nhehgvyf' 'continue' 'list' 'git'
+          'makepkg-conf:' 'pacman-conf:' 'provides' 'repo-list' 'nover-shallow'
+          'no-ver-shallow')
 
-longopts=allan,bind:,chroot,directory:,force,ignore:,ignorearch,ignore-arch,log
-longopts+=,noconfirm,no-confirm,nover,no-ver,noview,no-view,print,rmdeps,rm-deps
-longopts+=,sign,temp,tar,repo:,database:,root:,upgrades,nhehgvyf,continue,list,git
-longopts+=,makepkg-conf:,pacman-conf:,provides,repo-list,nover-shallow,no-ver-shallow
-shortopts=d:D:AcfglLnprstTu
-
-if optstring=$(getopt -o "$shortopts" -l "$longopts" -n "$argv0" -- "$@"); then
-    eval set -- "$optstring"
-else
+if ! parseopts "$opt_short" "${opt_long[@]}" -- "$@"; then
     usage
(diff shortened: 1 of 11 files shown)


AladW:
I agree that aligned ;; isn't worth it. About the loop style, it's fine by me as long as we stick to lower-case for variables that are not exported. I don't feel too strongly about the vertical style, it used to be like this then I changed it for some reason.

Note that aur-jobs has the same options as parallel, so `aur jobs --help` should have the desired effect. The other candidates have no options to begin with I believe.

lib/aur-sync
```diff
@@ -101,124 +101,70 @@ usage() {
...
-    eval set -- "$optstring"
-else
+if ! parseopts "$opt_short" "${opt_long[@]}" -- "$@"; then
     usage
 fi
-
-tmp=$(mktemp -dt "$argv0".XXXXXXXX)
-tmp_view=$(mktemp -dt view.XXXXXXXX)
-trap 'trap_exit' EXIT
+set -- "${optret[@]}"
```
AladW: I guess `optret` is the one exception here, in that `parseopts` sets the upper-cased variable `OPTRET` (the other variables are user-defined)

rafasc: Good catch, I was mindless switching the case. I'll fix that real quick.
Thanks.

AladW:
About the inconsistency of quoting; I typically don't quote variables when quotes would make no differences (such as assignments to other variables). If this is confusing I don't mind if these quotes are added back.

rafasc:
If you're ok with it, I don't mind it either way.

lib/aur-sync
```diff
@@ -101,124 +101,70 @@ usage() {
...
-        --provides)
-            provides=1
-            shift ;;
-        -d|--database|--repo)
-            repo=$2
-            shift 2 ;;
-        --root)
-            root=$2
-            shift 2 ;;
-        *)
```
AladW: Since I'm not yet familiar with `parseopts` - is `--` always guaranteed to be present in the option string, so that a glob is not required?

rafasc: I believe that's the case. c.f [parseopts.sh.in#n169][2] and [makepkg.sh.in#n1290][1]

[1]:https://git.archlinux.org/pacman.git/tree/scripts/libmakepkg/util/parseopts.sh.in?h=release/5.1.x#n169
[2]: https://git.archlinux.org/pacman.git/tree/scripts/makepkg.sh.in?h=release/5.1.x#n1289

Although, seems like makepkg.sh has a leftover `OPT_IND=0` from the previous option parser.

AladW: Thanks. For OPT_IND, I'll blame @eli-schwartz. If the parser breaks, I'll blame him for it as well. :smile:

AladW:
Looks like the points on style were solved, so I'll go ahead and merge this. Thanks for your work!

If more scripts need option support this can be covered in a separate PR.

rafasc:
Thanks.

lib/aur-sync
eli-schwartz: Oh yes, blame poor Eli for everything. I see how this goes these days...

And after I gifted --packagelist to the world!

*goes off, sulks, breaks getopt to teach the world a lesson*

PR #427, opened 2018-07-13
Preserve GNUPGHOME in aur-chroot

### LINKED ISSUE #426: aur-chroot does not preserve GNUPGHOME
As per title: `--preserve-env=GNUPGHOME` should be added to the arguments of `sudo makechrootpkg` in `aur-chroot`:[116][1] if we are to build packages with PGP signatures. For its part, `makechrootpkg` takes particular care to preserve GNUPGHOME for the parts executing outside of the chroot. (I have encountered this issue with 1.5.3, not with latest git, but the `aur-chroot` code concerned appears to mirror that in former `aurbuild` exactly.)

[1]: https://github.com/AladW/aurutils/blob/master/lib/aur-chroot#L116

alexshpilkin:
Closes #426. I added `env` inside `sudo` because that's what makechrootpkg does, not sure if it’s necessary.

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -113,7 +113,7 @@ if ((prepare)); then
 fi
 
 if ((build)); then
-    sudo PKGDEST="${PKGDEST:-$PWD}" makechrootpkg -r "$directory" \
+    sudo --preserve-env=GNUPGHOME,PKGDEST makechrootpkg -r "$directory" \
          "${makechrootpkg_bindmounts_ro[@]}" "${makechrootpkg_args[@]}"
 else
     printf >&2 'container path: %q\n' "$directory"


lib/aur-chroot
```diff
@@ -113,8 +113,9 @@ if ((prepare)); then
 fi
 
 if ((build)); then
-    sudo PKGDEST="${PKGDEST:-$PWD}" makechrootpkg -r "$directory" \
-         "${makechrootpkg_bindmounts_ro[@]}" "${makechrootpkg_args[@]}"
+    sudo --preserve-env=GNUPGHOME env PKGDEST="${PKGDEST:-$PWD}" \
```
AladW: The man page mentions that preserve-env takes a comma-separated list (seems a recent introduction), so
```
sudo --preserve-env=GNUPGHOME,PKGDEST
```
should work since makechrootpkg [defaults to PWD](https://git.archlinux.org/devtools.git/tree/makechrootpkg.in#n386) if PKGDEST is not set.

lib/aur-chroot
```diff
@@ -113,8 +113,9 @@ if ((prepare)); then
...
+         makechrootpkg -r "$directory" "${makechrootpkg_bindmounts_ro[@]}" \
+         "${makechrootpkg_args[@]}"
```
AladW: A third break is probably too much (I'm aware devtools adds a break for every argument in some places, which is quite strange)

Earnestly:
(Silly github "review" system.)

lib/aur-chroot
Earnestly: This is the best solution then

AladW:
To this day I have no idea why we have both "Single comment" and "Review". I guess with latter you can "approve" if it's somehow not clear from the comments... :roll_eyes:

maximbaz:
10 "single comment" generate 10 email notifications, 1 "review" with 10 comments generates 1 big email notification with all the comments.

AladW:
Ah, good point. :slightly_smiling_face:

lib/aur-chroot
alexshpilkin: I overlooked this, thanks. Now this is truly a one-line change =)

alexshpilkin:
Thanks for the suggestions, should be better now. I rewrote the commit and rebased on top of AladW:master.

AladW:
:+1:

Issue #334, opened 2018-03-25
aur-build: ccache support


cyrinux:
I'm not sure if ccache is supported.
I think from chroot build cache is lost?
Is it planned to support it?
https://wiki.archlinux.org/index.php/Ccache

Regards

AladW:
https://wiki.archlinux.org/index.php/Ccache#makechrootpkg

Which brings us to more end-of-options abuse:
```
aur build -c -- -d /path/to/your/cachedir:/ccache -cu -- CCACHE_DIR=/path/to/your/cachedir
```
With `aur-sync` you'd use `--bind` but the CCACHE_DIR env isn't passed. (The above command line depends on makepkg's ability to take environment variables from the command line, see https://git.archlinux.org/pacman.git/tree/scripts/makepkg.sh.in#n1287)

AladW:
#370 should cover this issue and #369

bartoszek:
Here's my current invocation for ccache build with aurutils.
```
aur chroot -B -D /tmp/$(uname -m) -M /etc/makepkg.conf.ccache -d aur-local-repo; arch-nspawn /tmp/$(uname -m)/root pacman --noconfirm -S ccache; aur vercmp-devel|vipe|cut -d: -f1|aur sync -c -D /tmp/$(uname -m) -T --bind-rw=/home/_ccache:/build/.ccache --makepkg-conf=/etc/makepkg.conf.ccache --no-ver-shallow --no-view -; sudo rm -rf /tmp/x86_64/
```
One things i would consider to include in aurutils to make it nicer
:arrow_double_up: detecting `ccache/distcc` in `makepkag.conf@BUILDENV` and insert those to deps.
- Could be done in `aur-sync` with querying `makepkg.conf` and appending `ccache` to deps array
- Alternatively could be included in `aur-chroot` by also querying `makepkg.conf` and appending `cache` to `mkarchroot` invocation.
- Eventulay i could be added as extra ops to `aur-sync`, something like `-(I)nstall: ${pkgname}` for user to pass additional packages that is needed for build procedure.
- Lastly, as most dirty/simplest resort: just append `ccacke distcc` to  https://github.com/AladW/aurutils/blob/3e89f9eb829f6d031ab19d4b92054d7c89aed26b/lib/aur-chroot#L111

Btw. the last solution seems most resonable to me :smirk:

Earnestly:
[...]

PR #428, opened 2018-07-18
add bind-rw switch to aur-sync for ccache handling

bartoszek:
address discrepancy in #334


AladW:
Should probably be included in `aur-sync(1)` too: https://github.com/AladW/aurutils/blob/master/man1/aur-sync.1#L160

PR #430, opened 2018-07-28
Follow proper man punctuation standards

bb010g:
Reformat man page sources to follow end of sentence, end of line and spaces between punctuation and words on macro lines. Lines have also been reflowed to be at or under 72 characters per line.

---
This change is [Reviewable](https://reviewable.io/reviews/aladw/aurutils/430)

### DIFF
--- a/man1/aur-build.1
+++ b/man1/aur-build.1
@@ -25,8 +25,8 @@ All arguments after \-\- are passed to \fBmakepkg\fR(8), or
 
 .TP
 .BI "\-a " queue
-A text file with directories containing a PKGBUILD. If none is
-specified, the current directory is assumed.
+A text file with directories containing a PKGBUILD.
+If none is specified, the current directory is assumed.
 
 .TP
 .B \-c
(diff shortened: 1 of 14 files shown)


man1/aur-build.1
```diff
@@ -104,19 +107,19 @@ For example:
...
 .SH SEE ALSO
-.BR aur (1),
-.BR aur\-chroot (1),
-.BR pacconf (1),
-.BR pacman (1),
-.BR makepkg.conf (5),
-.BR pacman.conf (5),
-.BR makepkg (8),
-.BR repo-add (8),
+.BR aur (1) ,
```
Earnestly: This is incorrect, there should be no space between `(1)` and `,`.  This would lead to the `,` being rendered as bold which is not the intent.

FichteFoll: Afaicr, this behavior with having punctuation as separate arguments only works in the mdoc extension, not man.

FichteFoll:
Considering you added newlines between sentences, you should consider doing this for subsentences or building blocks in a sentence as well, as that would improve diffs significantly. This is called semantic line feeds.

AladW:
@bb010g any comments on the above?

AladW:
Since I've removed all `\f` macros, any required remaining changes should be discussed in a separate issue or PR, should theres till be interest.

PR #431, opened 2018-07-30
Update man of aur repo-filter

Lucki:
There are some differences in the documentation and the actual behavior of `aur repo-filter`.


AladW:
:+1:

PR #433, opened 2018-08-08
aur-chroot: auto distcc/ccache

### LINKED ISSUE #369: aur-build: distcc support
Hi,
Would it be nice to have a `distcc` support added to `aurbuild`. Currently I've created `/usr/share/devtools/makepkg-x86_64_distcc.conf` with according distcc configuration and add `--distcc` switch to `aurbuild` to select this configuraiont, and append `distcc` alongside `base-devel` to `mkarchroot`.
I think a solution could be to infect `options=(distcc)` to PKBUILDs and export `DISTCC_HOSTS` environment variable to build chroot.
cheers, Bartus.

bartoszek:
### Autodetect ccache/distcc in aur-chroot
Slick an simple, using `check_buildenv` function from option.sh@pacman. 
- enter subshell
- - source `makepkg.conf`
- - source `option.sh@mapkepkg`
- - call `check_buildenv` for ccache and distcc
- store package names in `makepkg_conf_buildenv_deps` variable
- append `makepkg_conf_buildenv_deps` to `pacman` invocation in `prepare` section.
resolves #369 #334

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -92,6 +92,15 @@ makepkg_conf=${makepkg_conf-/usr/share/devtools/makepkg-$machine.conf}
 
 # CacheDir will be replaced by devtools (#416)
 pacman-conf | host_with_cache "${host_repo[@]}" >"$tmp"/custom.conf
+# parse the makepkg.conf for buildenv dependencies
+readonly makepkg_conf_buildenv_deps=$(
+    deps_array=()
+    source ${makepkg_conf}
+    source /usr/share/makepkg/util/option.sh;
+    check_buildenv "ccache" "y" && deps_array+=("ccache")
+    check_buildenv "distcc" "y" && deps_array+=("distcc")
+    echo ${deps_array[@]}
+)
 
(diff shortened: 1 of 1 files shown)


bartoszek:
This could be expanded further, for the users of [makepkg-optimize](https://aur.archlinux.org/packages/makepkg-optimize2/) :thinking:

AladW:
Since you only have to install distcc or ccache once on creation of the container, is there much point in the added code? (Especially since we have a dedicated section for ccache in `aur-chroot(1)`.)

AladW:
Apparently nothing speaks to just installing ccache and distcc unconditionally (next to base-devel), so let's do that. The man page would then also need updating

bartoszek:
I would keep this PR, perhaps in the future, we could have support for extra `makepkg optimizations` 🤔

AladW:
I can't say I've ever used `makepkg-optimize` (and it is unlikely to be merged to `makepkg` upstream), so I'm content with the expanded `man` page for now. Thanks.

PR #439, opened 2018-08-22
aur-repo-filter: pass --repo=... to pacsift

DamienRobert:
If --repo=reponame is not passed to pacsift, then it won't filter to match
$reposet and will output 'local/pkgname' which then yields an expac
'package not found' error.


AladW:
:+1:

PR #440, opened 2018-08-22
aur-build: fix a bug in find_pkg_dupes()

DamienRobert:
Indeed `makepkg --packagelist` returns the absolute name of the package,
so the find afterwards won't find it even if it exists. Add a
`| xargs basename -a` missing between the `makepakage
--packagelist` and the `find`.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -36,7 +36,7 @@ find_pkg_dupes() {
     # This function uses the extended --packagelist introduced with
     # pacman 5.1. It assumes that PKGEXT does not change between
     # invoking --packagelist and makepkg in the later build process.
-    makepkg --packagelist | xargs basename -a | xargs -I{} find "$1" -maxdepth 1 -name '{}'
+    makepkg --packagelist | xargs -I{} basename -a '{}' | xargs -I{} find "$1" -maxdepth 1 -name '{}'
 }
 
 print_quoted() {
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -36,7 +36,7 @@ find_pkg_dupes() {
     # This function uses the extended --packagelist introduced with
     # pacman 5.1. It assumes that PKGEXT does not change between
     # invoking --packagelist and makepkg in the later build process.
-    makepkg --packagelist | xargs -I{} find "$1" -maxdepth 1 -name '{}'
+    makepkg --packagelist | xargs basename -a | xargs -I{} find "$1" -maxdepth 1 -name '{}'
 }
 
 print_quoted() {


DamienRobert:
Note that these two commits should probably be squashed together...

AladW:
Nice catch. Wouldn't it suffice to leave out `-name`? i.e.
```
makepkg --packagelist | xargs -I{} find '{}'
```

AladW:
Addition, the first argument could be passed to PKGDEST.
```
[vagrant@archlinux makepkg]$ PKGDEST=/var/cache/pacman/custom makepkg --packagelist | xargs -I{} find {}               
/var/cache/pacman/custom/aurutils-git-1.5.3.r790.g2bfa37d-1-any.pkg.tar.xz
[vagrant@archlinux makepkg]$ PKGDEST=/var/cache/pacman/custom makepkg --packagelist
/var/cache/pacman/custom/aurutils-git-1.5.3.r790.g2bfa37d-1-any.pkg.tar.xz
```

eli-schwartz:
That's a lot of subprocesses, why did it ever do that instead of:

```
while read -r pkgfile; do
    if [[ -f "${pkgfile##*/} ]]; then
        printf '%s\n' "${pkgfile##*/}"
    fi
done < <(makepkg --packagelist)
```

No need to munge $PKGDEST, no need to use the find executable as your bash builtin `test -f`

DamienRobert:
Yes leaving out '-name' and passing PKGDEST=$1 to makepkg would indeed be a cleaner solution, probably with 2>/dev/null to find to silence the files not found messages.
(If you don't set PKGDEST, then the $1 argument (= the repo location) becomes useless :-))

AladW:
Thanks. I dropped the function and did something similar: https://github.com/AladW/aurutils/commit/7b3df954fb8301debc229f5c1d5a890183d4c690

That way, there's no need to silence the "not found" messages. PKGDEST is required either way, since build and repo location differ:
```
$ makepkg --packagelist
/vagrant/aurutils/makepkg/aurutils-git-1.5.3.r790.g2bfa37d-1-any.pkg.tar.xz
$ PKGDEST=/var/cache/pacman/custom/ makepkg --packagelist
/var/cache/pacman/custom/aurutils-git-1.5.3.r790.g2bfa37d-1-any.pkg.tar.xz
```

DamienRobert:
Ok, I agree with your changes, so I'll close this pr!

I know that the build and repo locations differ, that's why my first hacky commit was to use `xargs basename -a` rather than simply replacing '-name' by '-wholename' in `find` :)

Issue #29, opened 2016-03-07
Use a common.bash library


AladW:
To reduce code duplication.

AladW:
On second thought, having a library to dump stuff to makes "quick fixes" too tempting.

PR #441, opened 2018-08-23
aur-chroot: A local repo without a file:// entry is not an error

DamienRobert:
`aur sync -c -d foo -r foo_db_path` will call `aur chroot -d foo`, and
if the repo 'foo' does not have a corresponding 'file://' entry in
pacman.conf, then the grep -Em1 fails which exit the script.

But this is not an error, it just means that the corresponding directory
won't be bind_mounted. So desactivate 'errexit' for this case.

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -80,7 +80,8 @@ done > "$tmp"/custom.conf
 
 for repo in "${host_repo[@]}"; do
     # FIXME: custom.conf may contain multiple file:// entries per local repo
-    server=$(get_file_server "$repo" "$tmp"/custom.conf | grep -Em1 '^file://')
+    # the '|| true' is to not exit when no file servers are available
+    server=$(get_file_server "$repo" "$tmp"/custom.conf | grep -Em1 '^file://' || true)
     server=${server#file://}
 
     # arch-nspawn only supports specifying a single CacheDir via -c


DamienRobert:
By the way the way $server is autodetected is not uniform:
- if $database is specified, then aur-build and aur-sync use
~~~ sh
    server=$(pacconf --single --repo="$database" Server)                        
    server=${server#*//}                                                        
~~~
This works only if the first Server argument refer to a file:// and not an http://

- if $database is not specified, then aur-sync (and not aur-build) does autodetection via conf_file_repo and this works whenever there is only one repo with (only) one file:// Server

- arch-chroot uses grep -Em1 '^file://'

What about replacing `server=$(pacconf --single --repo="$database" Server)`
by `server=$(pacconf --repo="$database" Server | grep -Em1 '^file://')`?
(with the same set +e as for arch-chroot). This would autodetect the first file:// configured Server.

DamienRobert:
Going back to `aur chroot`, for my usage my custom repo is hosted by a remote server, and I configured pacman to retrieve it via https://. For aurutils operations I sshfs the remote directory to my local machine, and pass '-r path_to_the_sshfs' to aurutils. This works well, and this is not a problem for `aur chroot` because since the repo is available over the web, there is no need to bind mount any local directory.

AladW:
Thanks for the analysis. `|| true` should be enough in this case; I'll take a closer look later.

DamienRobert:
Ok I updated the pr with a `|| true` instead.

AladW:
About server detection not being uniform, I wonder if such detection is needed in all 3 scripts (`aur-build`, `aur-sync` and `aur-chroot`). 

In particular, we could keep everything (including `--repo-list`) in `aur-build` and specify bind mounts to `aur-chroot` manually. What do you think?

DamienRobert:
`aur-chroot` definitively don't need to know about AUR_REPO, it could just simply bind mount all file:// servers. (This could even be a feature to ask upstram for `makechrootpkg`.

`aur-sync` needs to know about the db for `-u`, we could move the `--list` to `aur-build` but this would involve passing the options to it correctly.

In https://github.com/DamienRobert/aurutils/tree/wip you can see my current implementation of unifying the logic. (Obviously the fixup commits still there show this is a wip :)) In particular I added a new binary
`aur-repolist` that can list the db, I find this more modular than stuffing everything in `aur-sync`.

AladW:
I'm not sure about having `lib` files that are not executables, see #29. However I do think that `aur-repolist` is a good starting point for a more general repository management tool akin to `repoctl`, see #444.

About `aursync -u`, the main reason it exists is that I didn't know how to use `xargs` while preserving `stdin`. Since then I found out about `xargs -a`, so I'd say it's not mandatory `-u` stays, if there are benefits in leaving it out.

AladW:
I'll merge this now and take a look at #448 later. Thanks.

PR #442, opened 2018-08-23
aur.1: fix typos

DamienRobert:
Print Perl modules that are both in the AUR and official repositories:
=> using -v will show Perl modules that are in official repositories but
not in AUR

Check foreign packages for AUR updates:
=> without -m then this checks all packages


AladW:
Thanks. About `mksrcinfo`, the main reason I kept it around was that it's significantly faster than `makepkg --printsrcinfo` (at least in the 2015 release version), which is relevant with many PKGBUILDs. I don't mind discussing this in a separate PR though.

PR #443, opened 2018-08-23
Add $AUR_DBROOT

DamienRobert:
This allows to specify the root via an env var rather than the -r argument.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -97,7 +97,7 @@ else
     usage
 fi
 
-db_root=$(realpath -e -- "${db_root-$server}")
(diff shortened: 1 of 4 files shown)


AladW:
If we add this it should probably be documented in the `ENVIRONMENT` sections of `aur-build(1)` and `aur-sync(1)`

DamienRobert:
Done for `aur-sync(1)`, but `aur-build(1)` does not have an `ENVIRONMENT` section. Should I add it?

AladW:
Sure, I believe aur-build supports AUR_REPO as well, I just forgot to
include such a section in aur-build(1)

DamienRobert:
Done

AladW:
:+1:

PR #445, opened 2018-08-24
aur-sync: --list should not clear AUR_VIEWDIR

maximbaz:
I'm still not 100% sure that having a global state was a good idea (e.g. what if you run `aur sync somepkg` in parallel in two terminals?), but at least `aur sync --list` should not clear this shared folder. 

I use `aur sync --list` in a script that runs periodically to be notified of new updates, and I was very surprised when out of the blue vifm got cleared while I was reviewing PKGBUILD 😄

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -224,14 +224,6 @@ if ! [[ -w $root/$repo.db && -r $root/$repo.db ]]; then
     exit 13
 fi
 
-# reset state of view directory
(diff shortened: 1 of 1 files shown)


AladW:
Thanks. I admit not thinking about concurrency at all when introducing AUR_VIEWDIR. I reverted the commit: https://github.com/AladW/aurutils/commit/3359668a039f43206c7923deaa17652d1a03e37d

PR #448, opened 2018-08-26
[wip] factorise db handling and add `aur repolist`

DamienRobert:
I am opening this new pr to continue the discussion of
    https://github.com/AladW/aurutils/pull/441
(which is out of scope of that pr).

So this pr add a library helper to handle the detection of the database file in a uniform manner, and add 'aur repolist'.

You don't have "having lib files that are not executables", and I can agree
with that, but personally I hate even more code duplication, which get too easily unsynchronised :)

In particular I agree that lib helpers should not be used for 'plumbing'
commands, but I think they are ok for 'porcelain' commands. However I was
operating under the assumption that 'aur sync' was a porcelain command, but
since you talk about removing '-u' maybe you consider it as a 'plumbing'
command?

If you prefer the 'plumbing' route, then we could do the following things:

- remove '-u' from `aur sync` (eventually make a 'aur checkupdate' for
this), and --list, --repo-list (which could go to aur repolist).

- In `aur build` remove all autodetection of the repository path from the
repository name, and request to be given the direct path to the db file
(something like $AUR_DB=$AUR_DBROOT/$AUR_REPO.db). This has the advantage of
only having to pass one variable rather than two, and $AUR_DBROOT and
$AUR_REPO are easily recovered from $AUR_dB). Eventually add 'aur config'
for db autodetection.

- `aur vercmp` has two purposes: compare packages between stdin and the aur
or compare packages between stdin and a file (+ the -c option which is
essentially reversing the comparison order with a different output format).
Change it to only compare the versions of two files, and move the aur
comparison to 'aur checkupdate'.

So `aur sync -u` would be replaced by `xargs -a <(aur repolist $AUR_DB |
aur checkupdate | cut -f 1) aur sync`.

Would you prefer such an approach?

Also what about `aur repolist`? Since you talk about replacing `repoctl` do you have in mind something more like `aur repo update ..., aur repo list ...`?

### DIFF
--- a/aur.in
+++ b/aur.in
@@ -23,6 +23,7 @@ if [[ -z $1 ]]; then
 fi
 
 readonly PATH=$lib_dir:$PATH
+export AURUTILS_LIB=$lib_dir
 
 if type -P "aur-$1" >/dev/null; then
     exec "aur-$1" "${@:2}"
--- a/contrib/vercmp-devel
+++ b/contrib/vercmp-devel
@@ -21,7 +21,7 @@ trap 'rm -rf "$tmp"' EXIT
 cd_safe "$tmp"
 
 # get repo contents
-aur sync --list "$@" | cut -f2,3 >db_info
+aur repolist "$@" >db_info
 
 # checkout latest revision for existing pkgbuilds
 # this sources the pkgbuilds assuming they have been viewed priorly
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -46,8 +46,8 @@ usage() {
 }
 
 source /usr/share/makepkg/util/util.sh
-source /usr/share/makepkg/util/message.sh
 source /usr/share/makepkg/util/parseopts.sh
+source $AURUTILS_LIB/db_helper.sh
 
 if [[ -t 2 && ! -o xtrace ]]; then
     colorize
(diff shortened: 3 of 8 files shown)


maximbaz:
I support the plumbing route, it makes the code simpler and personally I anyway have aliases for all aurutils operations that I use.

AladW:
I generally agree with the direction of this PR. Some notes, also on #450:

### aur-repo
* Instead of using subcommands like `repoctl`, I've went with `aur-repo` and modes enabled by short or long options.
  * With 07bfe40, this includes `update` (similar to `cower -u`), lists the repo both as represented in the sync db and directly, and prints the database path for later reuse.
    * I prefer using command-line options and stdin/stdout for IPC due to exported variables being [historically problematic](https://lists.archlinux.org/pipermail/aur-general/2015-August/031314.html) for AUR helpers. This is unlikely to cause issues with "our" `AUR` namespace, but better safe than sorry. (Right now, the only variable we export outside of user interaction is `PKGDEST`.)
* Not all repo functionality must necessarily be merged to this script, e.g. the specialized functions in `aur-build`.
* Configuration of `pacman.conf` is likely best left to the user, as the steps involved are small and up to personal preference/use case.
* Some common requested functionality like `aur repo --remove` may be added. 

### aur-build
* I see `aur-build` both as plumbing and porcelain. It contains all the build/repo-add logic, but I often use it directly from the command line.
  * Ideally, `aur-build` would do nothing if the full path to the database is passed as argument, and use `aur-repo` to automatically detect a `file://` repo if not. Compare 65eea31.
  * Still figuring out the details on `AUR_REPO` and `AUR_DBROOT` vs. `AUR_DBPATH`.

### aur-checkupdates / aur-buildorder
* I've tried this approach before, but besides resulting in many user complaints, there's little point to create separate executables for a few lines of code. 
  * The point in https://github.com/AladW/aurutils/pull/448/commits/349f89ae288aa1e1f156d83b672871074b15f6ee however remains valid. In 07bfe40, I'me moved this functionality to `aur-repo`.
  * With repo functionality from `aur-sync` moved to `aur-repo`, the script has a reasonable size again.

AladW:
On `aur-vercmp`: https://github.com/AladW/aurutils/pull/450/commits/62271b6f3fc1e3c70a0aec896854ab24bc01851e sets `upair` for both all and equal, which should be avoided for performance reasons. See https://github.com/AladW/aurutils/commit/dd4c302997c5eaee752a54c974c237a2b952d5c8. Considering the amount of testing that went to `aur-vercmp`, I'd probably avoid too drastic changes.

AladW:
Thanks for your work!

Issue #404, opened 2018-05-28
aur-repo-filter: support versioned packages


AladW:
See cdee9ceaa109234b7941947113faeb693b150e9a

AladW:
Related https://github.com/falconindy/expac/pull/35

AladW:
As it seems, `pacsift` doesn't support this: https://github.com/andrewgregory/pacutils/issues/24

maximbaz:
I feel that in this case explicitly using --no-provides is okay, if you already have a package that provides X and you want a different package that provides the same functionality, it's okay to ask user to confirm this with an extra flag.

AladW:
That makes sense, but the current UI doesn't really support it. When I see the above output, I think something went wrong: an upgrade is announced, followed by "there is nothing to do". Something like "Warning: foo: bar -> baz" could work.

Until then I've added a short option for `--provides`, so now you can do `aur sync -Punk` for the classic `pacaur` behavior. :slightly_smiling_face:

AladW:
I see no way to solve this apart from writing a C program interfacing `libalpm` (and perhaps even then it might now work), so removing this from the 2.0 milestone.

AladW:
Some progress: https://github.com/AladW/aurutils/commit/ad8111496ca56d689905fbe57fb52d05451279fa

PR #452, opened 2018-09-27
devel: implement aur-repo

AladW:
The main change is the implementation of `aur-repo`, see #448. This replaces `aur-vercmp -d <repo>` by `aur-repo -u`, introduces the `AUR_DBPATH` environment variable, and makes `aur-sync`'s repositoy selection available to `aur-build`.

Other changes include:
* Add `-P` (#404)
* Add long options for all scripts
* Remove dependency on `ed`
* Remove tree mode for `vifm`, display directories in build order (#379)

cc @DamienRobert


AladW:
@rafasc This probably isn't functional in the slightest, whenever you have time to take a further look. At least it's on a separate branch this time. :slightly_smiling_face:

rafasc:
@AladW good call on making a separate branch :) Especially because 4b454ae actually broke it.

Don't consider this a full-blown review, didn't have much time and didn't gave it much thought. But here are the places I had to amend to get it somewhat running.

Issues were caused by changing the output of `conf_file_repo` to use null chars instead of newline separators. Not sure if you want to revert to newlines, o keep the safer nulls since we're dealing with paths.

``` diff
diff --git a/lib/aur-repo b/lib/aur-repo
index 660a249..7c37ed7 100755
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -12,12 +12,12 @@ conf_file_repo() {
             repo = substr($1, 2, length($1)-2)
         }
         $1 ~ /^Server/ && $2 ~ /^file:/ {
-            printf("%s\0%s\n", repo, $2)
+            printf("%s\0%s\0\n", repo, $2)
         }'
 }
 
 db_namever() {
-    awk -v '
+    awk '
     /%NAME%/ {
         getline
         printf("%s\t", $1)
@@ -74,7 +74,7 @@ while true; do
         --status-file) shift; status_file=$1 ;;
         -l|--list)     mode=list_packages ;;
         -S|--pacman)   mode=list_pacman ;;
-        -u|--upgrades) mode=list_pacman_upgrades ;;
+        -u|--upgrades) mode=list_upgrades ;;
         --repo-list)   mode=repo_list ;;
         -h|--help) usage ;;
         --) shift; break ;;
@@ -84,7 +84,7 @@ done
 unset opt_short opt_long OPTRET
 
 case $mode in
-    repo_list) pacconf | conf_file_repo
+    repo_list) pacconf | conf_file_repo | xargs -0
                exit ;;
 esac
 
@@ -94,7 +94,7 @@ if [[ ${db_name=$AUR_REPO} ]]; then
         db_root=${db_root#*://}
     fi
 else
-    mapfile -t conf < <(pacconf | conf_file_repo)
+    mapfile -d '' -n2 -t conf < <(pacconf | conf_file_repo)
 
     case ${#conf[@]} in
         2) db_name=${conf[0]}
```
`contrib/vercmp-devel` also needs to be updated to use `repo --list` instead.

AladW:
Fixed apart that I can't get `read` to behave.
```
+(./aur-sync:172): main(): aur repo --list --status-file=db                     
+(./aur-sync:174): main(): read -rd db_name                                     
+(./aur-sync:175): main(): read -rd db_path 
# exit status 1
```
```
[vagrant@archlinux sync.e5xjahj6]$ IFS=: read -r a b < <(printf 'foo:bar\n')    
[vagrant@archlinux sync.e5xjahj6]$ echo $IFS                                    
                                                                                
[vagrant@archlinux sync.e5xjahj6]$ echo $a                                      
foo                                                                             
[vagrant@archlinux sync.e5xjahj6]$ echo $b                                      
bar
```
```
[vagrant@archlinux sync.e5xjahj6]$ IFS=$'\0' read -r a b < <(printf 'foo\0bar\n')
[vagrant@archlinux sync.e5xjahj6]$ echo $a
foobar
[vagrant@archlinux sync.e5xjahj6]$ echo $b
```

AladW:
That pacman repositories may contain arbitrary characters makes the `--status-file` idea annoying...

lib/aur-build
```diff
@@ -59,93 +56,95 @@ var_tmp=$(mktemp -d "${TMPDIR:-/var/tmp}/$argv0".XXXXXXXX)
...
-    db_path=$(realpath -e -- "$db_root/$db_name".db)
-    db_root=$(realpath -e -- "$db_root")
-else
-    error "$argv0: database: missing argument"
-    usage
+if type -P xdelta3 >/dev/null; then
+    repo_add_args+=(-d)
+fi
+
+if ! [[ ${db_path=$AUR_DBPATH} ]]; then
```
DamienRobert: I am a bit confused here, aur-repo only uses `$AUR_REPO` and `$AUR_DBROOT` (I guess because you want to use the name of the repo before symlink resolution), while aur-build checks `$AUR_DBPATH` too. But then if the user set both `$AUR_DBPATH` and `$AUR_REPO` there could be an inconsistency if AUR_REPO does not refer to the correct database.

This is unlikely to happen since `aur sync` will call `aur build` with the `-p` option anyway, but i find strange that only `aur-build` looks for a `$AUR_DBPATH` env var.

lib/aur-sync
```diff
@@ -108,58 +80,55 @@ if [[ -t 2 && ! -o xtrace ]]; then
...
     usage
 fi
 set -- "${OPTRET[@]}"
 
-unset pkg pkg_i root repo
+unset pkg pkg_i
 while true; do
     case "$1" in
-        -d|--repo|--database) shift; repo=$1 ;;
+        -d|--repo|--database) shift; repo_args+=(-d "$1") ;;
```
DamienRobert: `aur-sync` is the only binary accepting `--repo` as an alternative to `-d` or `--database`. Shouldn't we had this to the other binaries for consistency?

lib/aur-build
```diff
@@ -59,93 +56,95 @@ var_tmp=$(mktemp -d "${TMPDIR:-/var/tmp}/$argv0".XXXXXXXX)
...
+        -c|--chroot)    chroot=1 ;;
+        -f|--force)     overwrite=1 ;;
+        -s|--sign)      sign_pkg=1
+                        repo_add_args+=(-s) ;;
+        -v|--verify)    repo_add_args+=(-v) ;;
+        -D|--directory) shift; chroot_args+=(-D "$1") ;;
+        -N|--no-sync)   no_sync=1 ;;
+        -R|--remove)    repo_add_args+=(-R) ;;
+        --pacman-conf)  shift; chroot_args+=(-C "$1") ;;
+        --makepkg-conf) shift; chroot_args+=(-M "$1") ;;
```
DamienRobert: Nitpick: this would have been better to make these changes in commit 90a317513a311bb430983606f73597aa3b2dfa2b, (add long options) since this is independent of "aur-repo: implement --status-file (#448)"

lib/aur-repo
```diff
@@ -0,0 +1,132 @@
...
+case $mode in
+    repo_list) pacconf | conf_file_repo
+               exit ;;
+esac
+
+if [[ ${db_name=$AUR_REPO} ]]; then
+    if ! [[ ${db_root=$AUR_DBROOT} ]]; then
+        db_root=$(pacconf --single --repo="$db_name" Server) || exit 1
+        db_root=${db_root#*://}
+    fi
```
DamienRobert: Maybe add a test that db_root is a 'file://' and not an 'http://'?

lib/aur-repo
```diff
@@ -0,0 +1,132 @@
...
+unset opt_short opt_long OPTRET
+
+case $mode in
+    repo_list) pacconf | conf_file_repo
+               exit ;;
+esac
+
+if [[ ${db_name=$AUR_REPO} ]]; then
+    if ! [[ ${db_root=$AUR_DBROOT} ]]; then
+        db_root=$(pacconf --single --repo="$db_name" Server) || exit 1
```
DamienRobert: Rather than `exit 1` it would be nice to have an error message explaining the the path was not found.

AladW:
Thanks for the review. I'll take a look as time allows.

lib/aur-sync
AladW: I agree it should be consistent. I'm unsure if `--repo` is such a useful alias though.

lib/aur-build
AladW: I had a similar concern. The idea was that `aur-build` should remain usable if `aur-repo` breaks for some reason, though having a possible inconsistency is arguably worse.

lib/aur-repo
AladW: Agreed

lib/aur-repo
AladW: This should cover all cases:
```
    if ! [[ ${db_root=$AUR_DBROOT} ]]; then
        db_root=$(pacconf --single --repo="$db_name" Server)
        db_root=${db_root#file://}

        # check if variable was set correctly
        if ! [[ $db_root ]]; then
            error "$argv0: Server directive not found"
            exit 1
        fi

        # check if variable points to a local path (file://)
        if [[ $db_root == *://* ]]; then
            error "$argv0: Server: not a local directory"
            exit 1
        fi

        # check if local path is a directory
        if ! [[ -d $db_root ]]; then
            error "argv0: Server: not a directory"
            exit 20
        fi
    fi
```

lib/aur-build
AladW: More fun with `git rebase`...

lib/aur-build
AladW: Okay, my concern was that the symlink to `foo.db` was resolved twice (once in `aur-sync` through `aur-repo --status-file`, and once in `aur-build`). However, there's no real reason for `aur-repo` to already resolve this symbolic link (whose target may change at any point).

AladW:
I think this should cover the commentary above. (Apart from rebasing the "implement status file" commit, which is too time-consuming.) Please test if everything works as intended.

AladW:
20 force push laters and I'm satisfied!

eli-schwartz:
> Fixed apart that I can't get `read` to behave.

I think this is what you want?

` { read -rd '' a; read -rd '' b; } < <(printf 'foo\0bar\0')`

maximbaz:
Just a bit of testing, will continue within an hour I hope:

- The absence of tree mode is uglier, seeing the same tmp prefix in front of every file 😞 

    image

- vercmp-devel isn't working anymore, probably needs adjustments, didn't look into it

AladW:
Both should be fixed now

AladW:
There was a missing `cd`, so now `vifm` should look nice again. Btw @Morganamilo your one-letter change broke this PR \o/

AladW:
Fixed. Time to annoy everyone again with a broken master branch.

PR #453, opened 2018-10-08
Allow to override default scripts

abouvier:
Make it possible to override one of the built-in command (like aur-sync) by creating a script with the same name in a folder of the PATH (like ~/bin).

Example of `~/.local/bin/aur-sync` (with ~/.local/bin in $PATH):
```bash
#!/bin/bash
packages=${XDG_CONFIG_HOME:-$HOME/.config}/aurutils/sync/ignore
packages=$(sed '/^#/d' "$packages" | fmt -1 | paste -sd,)
exec /usr/lib/aurutils/aur-sync --ignore="$packages" "$@"
```

### DIFF
--- a/aur.in
+++ b/aur.in
@@ -22,7 +22,7 @@ if [[ -z $1 ]]; then
     exit 1
 fi
 
-readonly PATH=$lib_dir:$PATH
+readonly PATH=$PATH:$lib_dir
 
 if type -P "aur-$1" >/dev/null; then
     exec "aur-$1" "${@:2}"


rafasc:
That order is deliberate. 
We claim ownership of those verbs and do not allow overriding them.

Allowing users to override random aurutils scripts breaks the expectation of what the script actually does. This may affect either yourself or any third parties using aurutils scripts.

It could totally break core-aurutils if people start copying and pasting scripts there and forget to update them whenever aurutils gets updated.

You may argue: "they messed up, it's their fault."

But when it makes problems harder to debug, because a call to aur-sync does different things depending on how the user decided to mess with it, it becomes our problem too.

We allow and even encourage users to come up with their own aur-scripts. 
I suggest just choosing a name different from `aur-sync` for that one.

I'll wait what @AladW has to say about this, but I am pretty sure this should be a 'wont-fix'.

maximbaz:
Also, this extension mechanism was inspired by `git`, and git itself doesn't allow overriding built-in commands like "commit".

AladW:
I agree with the arguments above. I suppose you could mention the `PATH` order in `aur(1)`, and that modifying it (e.g. through a local copy of `/bin/aur`) is unsupported.

rafasc:
I think we should rename this PR as "Allow to override default scripts", and close this as 'wontfix' for future reference. While the documentation fix can be done on its own PR  referring back to this one.

AladW:
:+1:

PR #459, opened 2018-11-26
aur-build: Fix documentation inconsistencies

Wuestengecko:
`aur build`'s man page is inconsistent with its actual behaviour; options are passed to `makechrootpkg` and not `aur chroot` itself.

I also added a short note about `-C`, `-D` and `-M` options, as these aren't mentioned anywhere else in `aur-build(1)`.

### DIFF
--- a/man1/aur-build.1
+++ b/man1/aur-build.1
@@ -23,7 +23,7 @@ prefix in
 All arguments after \-\- are passed to
 .BR makepkg (8),
 or
-.BR aur\-chroot (1)
+.BR makechrootpkg
 when
 .B \-c
 is specified.


man1/aur-build.1
```diff
@@ -30,7 +30,8 @@ specified, the current directory is assumed.
 
 .TP
 .B \-c
-Build packages with \fBaur\-chroot\fR(1).
+Build packages with \fBaur\-chroot\fR(1). Also accepts
```
AladW: "Also accepts" is somewhat vague, perhaps replace it with "The ... options are forwarded to aur-chroot"?

PR #468, opened 2018-12-05
 Install bash completion by default

rafasc:
* fix multiple issues related to completion generation. Some caused by issues on the completions/Makefile and another caused by the m4 macro itself. Both could result in installation of stale completion scripts.
* Add vercmp-devel to completion macro
* Install bash completion by default


AladW:
:+1:

PR #469, opened 2018-12-05
aur-chroot: fix local repo path for bind mounts

amakinen:
The line from get_conf_repo is in format `Server = file:///path/to/repo`. Without successful bind mount, `pacman -Syu` fails in prepare stage, stopping the build.

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -77,7 +77,7 @@ makepkg_conf=${makepkg_conf-/usr/share/devtools/makepkg-$machine.conf}
 
 # FIXME: local repos may contain multiple file:// entries
 for repo in "${host_repo[@]}"; do
-    server=$(get_conf_repo "$repo" | tee "$tmp"/custom.conf | grep -Em1 '^file://' || true)
+    server=$(get_conf_repo "$repo" | tee "$tmp"/custom.conf | grep -Eom1 'file://.*$' || true)
     server=${server#file://}
 
     if [[ $server ]]; then


lib/aur-chroot
```diff
@@ -22,6 +22,10 @@ get_conf_repo() {
     pacconf --repo "$1"
```
rafasc: https://github.com/AladW/aurutils/blob/a001f5cd2ebd053a05488e73797aa836def8632e/lib/aur-chroot#L25-L27
The above function can be avoided if we pass the additional arguments to pacconf in `get_conf_repo`.

```suggestion
     pacconf --repo "$@"
```

lib/aur-chroot
```diff
@@ -77,7 +81,8 @@ makepkg_conf=${makepkg_conf-/usr/share/devtools/makepkg-$machine.conf}
 
 # FIXME: local repos may contain multiple file:// entries
 for repo in "${host_repo[@]}"; do
-    server=$(get_conf_repo "$repo" | tee "$tmp"/custom.conf | grep -Em1 '^file://' || true)
+    get_conf_repo "$repo" > "$tmp"/custom.conf
+    server=$(get_server_repo "$repo" | grep -Em1 '^file://' || true)
```
rafasc: Using the suggested above, this turns into:
```suggestion
    server=$(get_conf_repo "$repo" "Server" | grep -Em1 '^file://' || true)
```

lib/aur-chroot
```diff
@@ -77,7 +81,8 @@ makepkg_conf=${makepkg_conf-/usr/share/devtools/makepkg-$machine.conf}
 
 # FIXME: local repos may contain multiple file:// entries
 for repo in "${host_repo[@]}"; do
-    server=$(get_conf_repo "$repo" | tee "$tmp"/custom.conf | grep -Em1 '^file://' || true)
```
rafasc: Just updating the regex from `grep` to either of these would also work:
```suggestion
    server=$(get_conf_repo "$repo" | tee "$tmp"/custom.conf | grep -oEm1 'file://.*$' || true)
```
```suggestion
    server=$(get_conf_repo "$repo" | tee "$tmp"/custom.conf | grep -Pom1 '(?<=^Server = )file://.*$'
```

amakinen: Ah, that's cleaner. Didn't think of `grep -o`.

AladW:
Thanks. Perhaps I should find a better way in future rather than relying on `grep -Eo`...

AladW:
For example, I'm sure there will be someone who will come up with `[file://] Server = file://` :joy:

PR #471, opened 2018-12-06
aur-chroot: bind-mount all file:// paths

AladW:
Alternative to #469

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -74,7 +74,7 @@ makepkg_conf=${makepkg_conf-/usr/share/devtools/makepkg-$machine.conf}
   printf '[options]\n'; pacconf --verbose CacheDir
 
   for repo in "${host_repo[@]}"; do
-      printf '[%s]n' "$repo"; pacconf --repo="$repo"
+      printf '[%s]\n' "$repo"; pacconf --repo="$repo"
   done
 } > "$tmp"/custom.conf
 
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -70,12 +70,13 @@ fi
 pacman_conf=${pacman_conf-/usr/share/devtools/pacman-extra.conf}
 makepkg_conf=${makepkg_conf-/usr/share/devtools/makepkg-$machine.conf}
      
-for repo in "${host_repo[@]}"; do
-    printf '[%s]n' "$repo"; pacconf --repo="$repo"
-done > "$tmp"/custom.conf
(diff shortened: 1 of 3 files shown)


lib/aur-chroot
```diff
@@ -75,15 +75,14 @@ fi
...
-        bindmounts_ro+=("--bind-ro=$server")
-        makechrootpkg_bindmounts_ro+=(-D "$server")
-    fi
+    while read -r server; do
+        if [[ $server =~ ^file:// ]]; then
+            bindmounts_ro+=("--bind-ro=${server#file://}")
+            makechrootpkg_bindmounts_ro+=(-D "${server#file://}")
+        fi
+    done < <(get_conf_repo "$repo" | tee "$tmp"/custom.conf | pacini - Server)
 done
```
rafasc: Since we're using bash regex match, you can use capture groups:
``` diff
diff --git a/lib/aur-chroot b/lib/aur-chroot
index 217e18e..477c6c6 100755
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -81,3 +81,3 @@ for repo in "${host_repo[@]}"; do
-        if [[ $server =~ ^file:// ]]; then
-            bindmounts_ro+=("--bind-ro=${server#file://}")
-            makechrootpkg_bindmounts_ro+=(-D "${server#file://}")
+        if [[ $server =~ ^file://(.*)$ ]]; then
+            bindmounts_ro+=("--bind-ro=${BASH_REMATCH[1]}")
+            makechrootpkg_bindmounts_ro+=(-D "${BASH_REMATCH[1]}")
```
But I could not measure any performance gains.

Either way is better than #469.
👍

AladW: After some discussion on IRC I went with a case match instead, which also removes the need for `pacini`. (https://github.com/AladW/aurutils/pull/471/commits/1f76b26bb4cdb1b6c52b04d2cf963f50f0f0123a) Let me know what you think.

rafasc: Nice. `get_conf_repo()` is gone also gone, two birds with one stone. 👍

AladW: I guess I can merge it now, even though it still says "Changes requested"... the stuff github keeps piling onto their interface is confusing.

rafasc: I resolved the conversation thinking it would change that, but apparently I needed to click somewhere else. But since it got merged, I have no way to accept after the fact.
github githubbing. 🤷‍♂️

PR #473, opened 2018-12-06
aur-search: colorize if stdout is a tty

jcrd:
`aur-search` only prints colors on stdout.

### DIFF
--- a/lib/aur-search
+++ b/lib/aur-search
@@ -106,7 +106,7 @@ usage() {
 source /usr/share/makepkg/util/message.sh
 source /usr/share/makepkg/util/parseopts.sh
 
-if [[ -t 2 && ! -o xtrace ]]; then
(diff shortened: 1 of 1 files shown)


AladW:
All scripts use -t 2, as libmakepkg did when I implemented did. I guess libmakepkg now prints most of these messages to stdout now so that -t 1 is needed... :slightly_frowning_face:

jcrd:
Other commands with the `-t` check use `error` and `plain` from libmakepkg. `aur-search` is the only command that prints its own colors to stdout (in the search results) and doesn't use any libmakepkg functions other than `colorize`.

AladW:
Makes sense, though I have to wonder, in what situation will `stdout` be connected to a terminal, and `stderr` not? (Or vice-versa)

jcrd:
My use case was `aur search pacman | less`.

PR #481, opened 2018-12-08
Update and sync aur.1

### LINKED ISSUE #480: `man aur` should reference `aur-repo`

alfunx:
* Mention `aur-repo` and `aur-vercmp-devel`
* Use imperative
* Use description provided in sub-man-pages
* Use Aurweb instead of AurJson

Fixes #480

### DIFF
--- a/man1/aur-depends.1
+++ b/man1/aur-depends.1
@@ -1,6 +1,6 @@
 .TH AUR-DEPENDS 1 2018-04-29 AURUTILS
 .SH NAME
-aur\-depends \- retrieve dependencies using AurJson
+aur\-depends \- retrieve dependencies using aurweb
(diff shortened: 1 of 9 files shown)


AladW:
Should aurweb be capitalized or not? https://aur.archlinux.org/rpc.php
>the aurweb RPC Interface

alfunx:
Oh I was looking [at the Arch Wiki](https://wiki.archlinux.org/index.php/Aurweb_RPC_interface). I guess aur.archlinux.org/rpc.php is more official, should I lowercase it?

AladW:
Yeah, we should use lower case if upstream does.

AladW:
:+1:

PR #488, opened 2018-12-11
Make usage outputs more uniform

alfunx:
Some more documentation stuff:

* sort flags
* use `<repo>` and `<path>` more consistently
* remove `-h` / `--help` flags
* remove unused flags

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -44,7 +44,7 @@ trap_exit() {
 }
 
 usage() {
-    plain "usage: $argv0 -d database [-afNRsv] [--] <makepkg args>" >&2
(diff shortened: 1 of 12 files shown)


Earnestly:
Most `-h` flags could be avoided by simply having manuals.  A useful diagnostic message and simple usage synopsis is all a program need print if used incorrectly.

rafasc:
To be honest, I think we can probably drop `-h|--help` flags. They creeped in with the move to `parseopts.sh` because at the time the plan was to make shell completions by parsing the help messages. That's no longer the case.

AladW:
I would accept commits removing the `-h` flags.

alfunx:
Done!

AladW:
It's fine, thank you.

PR #490, opened 2018-12-13
aur-chroot: add --sysroot

AladW:
Additionally, use pacini and a static pacman.conf file rather than call `pacman-conf` repeatedly (with possibly differing results).


AladW:
Man page also needs an update.

AladW:
Another question is if rootdir makes any sense if we leave out stuff like DBPath anyway... in that case only the pacini changes are of some (slight) relevance.

PR #492, opened 2018-12-14
aur-build: Allow passing multiple local repos to aur-chroot

Wuestengecko:
This used to work in the past, but was lost at some point during the refactoring.

Personally, I use one local repo `aur` for packages built directly from the AUR without modifications, and one repo `custom` which holds modified and self-created packages. As most of the dependencies will land in the `aur` repo, it is almost always required during `aur build`.


alfunx:
Relevant commit: 2dca77720dca71f42a213ab84582c00dbf56292c from #452

PR #493, opened 2018-12-15
aur-repo: fix --sync and --upgrades

diabonas:
Fix two bugs introduced in #491:
- `aur repo -Sl` fails with the error
    
    ```
    awk: cmd. line:2:         $0 ~ rname { print $NF }
    awk: cmd. line:2:                    ^ syntax error
    ```
    due to a spurious pair of braces in the AWK program used by `ex_namever`.
- `aur repo -u` only returns the path to the database file due to a typo (`list_updates` instead of `list_upgrades`). Additionally correct the name of the long option name in the man page from `--updates` to the actually used `--upgrades`.

### DIFF
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -18,9 +18,7 @@ db_namever() {
 }
 
 ex_namever() {
-    awk -F"/" -v rname="^$1/" '{
-        $0 ~ rname { print $NF }
-    }'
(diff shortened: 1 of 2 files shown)


AladW:
Maybe one day I'll get a PR right...

AladW:
Maybe one day I'll get a PR right...

Issue #438, opened 2018-08-20
aur-repo-filter: support --sysroot


AladW:
To sift packages in containers. Unsure there's a corresponding `expac` option.

PR #496, opened 2019-01-08
aur-sync: implement agrep

AladW:
This is similar to git's "did you mean", for purely virtual packages in
the AUR. A package search alone does not suffice, since package
names may differ in minute ways. (For example, foo-pkg vs foo_pkg.)

Implementation is done using agrep from the tre package. Potential
candidates are presented to the user.

----
This does not work with chroot builds, as the pacman (sys)root cannot be specified to `expac`.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -282,6 +282,38 @@ if ((build)); then
         aur build "${build_args[@]}" -- "${makechrootpkg_args[@]}" \
             -- "${makechrootpkg_makepkg_args[@]}"
     else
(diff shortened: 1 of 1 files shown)


AladW:
While it is unlikely the AUR ever includes a `provides` search, as noted this won't work with chroot builds (#438), and can't be nicely implemented due to `tre`/`agrep` limitations.

Morganamilo:
Not sure if you follow the ML but the provides patch it at least making some progress now.

AladW:
As much progress as my babushka has in fighting capitalist America.

PR #498, opened 2019-01-10
aur-build: add --build-command

AladW:
Example uses:

* https://bbs.archlinux.org/viewtopic.php?pid=1824423#p1824423
* "--gendb"
* Build helpers with special abilities, e.g. rua
* extra-x86_64-build


AladW:
Probably better if we can achieve the same without exporting `AUR_DBROOT` and `AUR_REPO` (which is what `env` does).

AladW:
For host builds, PKGDEST is now set a temporary makepkg.conf file, rather than exported. This is similar to what `makechrootpkg` does for its makepkg instance. For the custom command, I've left PKGDEST and AUR* in the environment for now; wrapper scripts can easily unset it.

PR #499, opened 2019-01-11
Add zsh run-help assistant

nl6720:
This allows for zsh run-help to open the correct man page for the aur commands, instead of opening aur(1).
To use it, it must be autoloaded:

autoload -Uz run-help-aur

### DIFF
--- a/completions/Makefile
+++ b/completions/Makefile
@@ -16,3 +16,4 @@ install-bash: bash/aur
 
 install-zsh: zsh/_aur
 	@install -Dm644 zsh/_aur -t $(DESTDIR)$(SHRDIR)/zsh/site-functions
+	@install -Dm755 zsh/run-help-aur -t $(DESTDIR)$(SHRDIR)/zsh/functions/Misc
--- a/completions/zsh/run-help-aur
+++ b/completions/zsh/run-help-aur
@@ -1,0 +1,5 @@
+if [ $# -eq 0 ]; then
+	man aur
+else
+	man aur-$1
+fi


AladW:
cc. @alyptik

alyptik:
@AladW 

It seems a bit overengineered to me to just have this incredibly specialized zle just for aurutils. If the user wants this behavior it is incredibly trivial for them to implement it themselves. Let's not start heading down to the road to omz-ng here :)

Sorry, nack from me.

AladW:
Ok, thanks for your input.

nl6720:
`run-help` assistants are a feature of Zsh. It even ships with seven of them included (for `git`, `ip`, `sudo`, ...), so the comparison with omz was unwarranted.

alfunx:
For reference, the `run-help-git` script for Git provided by Zsh looks like this: https://github.com/zsh-users/zsh/blob/master/Functions/Misc/run-help-git

This implementation looks good IMO.

AladW:
For the record, I have no experience with shell completion whatsoever so I leave the decision to those who have added shell completion to aurutils (being @alyptik and @rafasc). If @rafasc has a different opinion I'm happy to take a look, possibly for a later milestone.

alyptik:
@nl6720 Sorry, didn't mean it in that sort of way, just me making a terrible attempt at a joke :(

My only issue with it is that, especially with zsh, I really tend towards only adding what is absolutely needed or thing which are also useful in general situations.

This is just a bit too specialized and since it's more of an icing on the cake thing to me it's just not something I feel comfortable approving of.

alyptik:
@nl6720 Let's respin this into a new PR which also includes updates to the documentation saying "what it's for" and "how to use it". We can revisit discussion there, that would be a lot more enticing a PR.

nl6720:
I created this because I regularly use `run-help` via the `Alt+h` shortcut to read the man page of the currently typed command, and aurutils 2.0 broke it. I've only seen run-help assistants shipped with Zsh and I'm not aware of any software shipping their own, so I was not surprised to see this getting rejected.
Anyway...

@alyptik , what do you mean by "updates to the documentation"? I haven't looked too closely, but aren't man pages the only documentation?
Is this really something that belongs in a man page? In `aur(1)`?

alyptik:
@nl6720 just something in the repo like a small addition to the readme explaining use

alyptik:
It's up to @AladW if that's something appropriate for his man page too

AladW:
How about including this in a README in the completion folder? (In the repository, that is.)

PR #518, opened 2019-01-25
lib/aur-sync: cd before invoking $PAGER again

CyberShadow:
Previous versions of aur-sync used to set the current directory to the per-sync temporary directory before invoking the `$PAGER` command. Recent changes (65517a01fdcad12e4cb75ce05d636d8571dc0e21) broke this assumption. (This was broken for both `$PAGER` and vifm, but was later fixed, only for vifm, in 854c85ee0612ff5ca0ab982e41dd5126b07a8df6).

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -256,7 +256,7 @@ if type -P >/dev/null vifm; then
     # avoid directory prefix in printed paths (#452)
     viewer() ( cd "$1"; vifm -c 'view!' -c '0' - )
 else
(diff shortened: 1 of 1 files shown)


AladW:
:+1:

AladW:
Tagged a new release: https://github.com/AladW/aurutils/releases/tag/2.2.1

CyberShadow:
Thanks. I was probably relying on an implementation detail there, so I ended up fixing my pager script anyway to read stdin, but it was a simple enough fix either way.

PR #519, opened 2019-01-31
aur-srcver: use exported functions

AladW:
This removes the `readlink`/`find` hacks from `aur-srcver`, using exported functions with `parallel` instead.

Note: as before, results are only printed if all arguments were processed successfully. Not sure if this is the best approach; I only vaguely recall discussing it.

### DIFF
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -5,7 +5,7 @@ readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 readonly startdir=$PWD
 
 srcver_pkgbuild_info() {
-    env -i bash -c '
+    env -C "$1" -i bash -c '
         PATH= source PKGBUILD
 
         if [[ -v epoch ]]; then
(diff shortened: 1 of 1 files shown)


maximbaz:
Something is not right, haven't been able to look into details yet, posting what I see, maybe you quickly spot where typo is:

```
❯ aur vercmp-devel                                                   
srcver: failed to update 10 packages
8<----
srcver_pkgbuild_path antigen-git
environment: line 0: cd: antigen-git: No such file or directory
8<----
srcver_pkgbuild_path aur-talk-git
environment: line 0: cd: aur-talk-git: No such file or directory
8<----
srcver_pkgbuild_path aurutils-git
environment: line 0: cd: aurutils-git: No such file or directory
...
```

Earnestly:
Relative paths claim another victim!  The `$1` in `srcver_pkgbuild_path` should probably take a value including `$startdir` (or where ever the respective `PKGBUILD` directory is) to ensure it's an absolute path.

AladW:
http://ix.io/1zJR/bash

Should work with both relative and absolute paths as argument now. If you used `"$startdir"/{}`, it looks like absolute paths don't work:
```
~$ cd ~//
~$ cd ~//usr
bash: cd: /home/user//usr: No such file or directory
```

Earnestly:
Sure it does.  You can have any number `/`'s inbetween path components.  Only a leading `//` has potential significance as implementation defined under POSIX.  However linux doesn't ascribe any meaning to a leading `//`.

AladW:
> In your example, you used cd to change directory to ~ and then tried to access ~/usr which evidently doesn't exist.

Right, that's what would happen if you did `parallel "$startdir"/{}"` from `~`, or differently:
```
$ startdir=$PWD
$ arg1=/usr
$ cd "$startdir/$arg1"
bash: cd: /home/user//usr: No such file or directory
```
Shouldn't be an issue with `env -C` and making `"$tmp"` explicit either way.

Earnestly:
For what it's worth, if you intended to support relative paths it would be customary to check if the argument is absolute before prefixing anything else.  It's just one more complication when using relative paths. :(

Issue #514, opened 2019-01-21
completion: put short options next to matching long options


AladW:
Compare: (`aur-sync`, `zsh`)
```
-A                --directory       --makepkg-conf    --rebuild
-B                -f                -n                --rebuild-tree
--bind            --force           --no-confirm      --rm-deps
--bind-rw         -g                --no-ver          --root
--build-command   --git             --no-ver-shallow  -s
-c                --ignore          --no-view         --sign
-C                --ignore-arch     -p                -t
--chroot          --ignore-file     -P                -T
--continue        -k                --pacman-conf     --tar
-d                -L                --pkgver          --temp
-D                --log             --print           -u
--database        -M                --provides        --upgrades
```
to (`pacman`, `zsh`)
```
--database  -D  -- Modify database                                                
--deptest   -T  -- Check if dependencies are installed                            
--files     -F  -- Query the files database                                       
--help      -h  -- Display usage                                                  
--query     -Q  -- Query the package database                                     
--remove    -R  -- Remove a package from the system                               
--sync      -S  -- Synchronize packages                                           
--upgrade   -U  -- Upgrade a package                                              
--version   -V  -- Display version and exit 
```
Note: the `pacman` completion for `bash` looks similar as the first output, so I'm not sure what's possible in `bash` completion. (The `bash` completion for `git` only completes long options.)

rafasc:
We can group short and long options together instead of showing them in alphabetical order. 

But with the way we are generating the completion scripts I don't think we can easily do anything more complex than that without writing non-generated scripts.

It's technically possible but would require to add supporting code to associate short with their long counterparts, and I feel that's that's probably going a bit too far.

If we want completions improvements we can be either go back to the parse `--help` output idea or write dedicated scripts from scratch. Both will require more maintenance (which was the thing the current one was trying to minimize) but otherwise result in prettier completions for the `zsh` case (there's no much we can do about `bash`).

AladW:
#520 works for me, thanks.

PR #520, opened 2019-02-01
completions: group options by type

rafasc:
Completions options are being shown in lexicographical order. The result
is that long and short options are shown mixed up.

This makes the completions hard to read as they're not well aligned due
to prefix length differences between short and long options (--/-).

To minimize these misalignments lets show them grouped together, showing
long options first followed by short options.

In response to #514


AladW:
cc. @alyptik

Issue #511, opened 2019-01-20
man7: add aurvcs(7)


AladW:
- [x] Add `aurvcs(7)` (https://github.com/AladW/aurutils/issues/500#issuecomment-455892225, `vercmp-devel` replacement)
- [x] Allow `aur-srcver` to skip non-existing directories

maximbaz:
So to confirm, the scenario we are trying to support is some kind of `--ignore` for `aur-vercmp-devel`?

AladW:
The idea is trying to find some more general procedure to deal with VCS packages, rather than trying to make aur-vercmp-devel suitable to everyone. There's not that much going on in `aur-vercmp-devel`; non-trivial bits (if any) can be moved elsewhere and the documentation expanded.

AladW:
I've merged some other potentially useful procedures from #217 to this issue.

AladW:
This is the plan I have in mind:
1. Move some functionality from `aur-vercmp-devel`
* Implement `--devel` in aur-repo to list all VCS packages. Since the AUR now has a definition of what a VCS package entails (see https://lists.archlinux.org/pipermail/aur-dev/2019-May/004649.html), the `AURVCS` variable is no longer required.
* Let `aur-srcver` skip non-existing directories with a warning.

With the above points, `aur-vercmp-devel` would be changed to:
```
#!/bin/bash
readonly XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}

tmp=$(mktemp -d)
trap 'rm -rf "$tmp"' EXIT

if "cd ${AURDEST:-$XDG_CACHE_HOME/aurutils/sync}"; then
    aur repo --list --devel "$@" | tee "$tmp"/db cut -f1 | xargs -r aur srcver >"$tmp"/new || exit

    if [[ -s $tmp/new ]]; then
       aur vercmp -p "$tmp"/new < "$tmp"/db
    fi
fi
```
This is similar in complexity to `aur-gc` and `aur-remove` and could thus be included in `aur(1)`, instead of a seperate script in `lib`.

2. Move the contents of aur-vercmp-devel(1) to aurvcs(7), and include some additional topics.
* Add methods on retrieving the latest revision directly, e.g. through `git ls-remote` or site-specific APIs (github, gitlab). Since VCS packages should follow https://wiki.archlinux.org/index.php/VCS_package_guidelines, the required information can directly be retrieved from the local repository, without external mechanisms such as a commit database.
[...]

PR #523, opened 2019-02-08
srcver: re-add --noprepare

rafasc:
d2c26cc removed the `--noprepare` flag from `makepkg` calls in `aur srcver`. 

This made calls to `aur-srcver-devel` slow because in my particular case a go package's `prepare()` was running `dep ensure` which took some time to finish.

If `pkgver()` is allowed to depend on the result of `prepare()` maybe this is wrong and is something we'll need to live with?

### DIFF
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -3,6 +3,29 @@
 readonly argv0=srcver
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 readonly startdir=$PWD
+makepkg_args=('--nobuild' '--nodeps' '--skipinteg')
+
+source /usr/share/makepkg/util/parseopts.sh
+opt_short=
+opt_long=('noprepare')
+opt_hidden=('dump-options')
+if ! parseopts "$opt_short" "${opt_long[@]}" "${opt_hidden[@]}" -- "$@"; then
+    echo "usage: $argv0 [--noprepare] <pkgbase> [<pkgbase> ...]" >&2
+    exit 1
+fi
+set -- "${OPTRET[@]}"
+unset opt_short opt_long opt_hidden OPTRET
+
+while true; do
+    case "$1" in
+        --noprepare)    makepkg_args+=("$1") ;;
+        --dump-options) printf -- '--%s\n' "${opt_long[@]}"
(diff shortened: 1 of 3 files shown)


rafasc:
Example of a legitimate use of `pkgver()` depending on the result of `prepare()`:
https://github.com/Earnestly/pkgbuilds/blob/a1d685b80a6df9a3578b3a31e3aa020ec3f88228/linux-mainline/PKGBUILD

Earnestly:
I'm not sure if that's legitimate or abusive :P

rafasc:
Instead of reverting to the previous behaviour, let's keep the current and add `--noprepare` as `srcver` option. (b03e6be)

lib/aur-srcver
```diff
@@ -3,6 +3,29 @@
...
+if ! parseopts "$opt_short" "${opt_long[@]}" "${opt_hidden[@]}" -- "$@"; then
+    echo "usage: $argv0 [--noprepare] <pkgbase> [<pkgbase> ...]" >&2
+    exit 1
+fi
+set -- "${OPTRET[@]}"
+unset opt_short opt_long opt_hidden OPTRET
+
+while true; do
+    case "$1" in
+        --noprepare)    makepkgargs+=("$1") ;;
```
AladW: nitpick: elsewhere (`aur-sync`, `aur-build`) we use `makepkg_args` (with underscore)

AladW:
This looks a reasonable solution, or at least, I can't think of a better one. I'd say to not wrap this in `aur-vercmp-devel` - the option logic would be longer than the script itself!

lib/aur-srcver
rafasc: ok. will fix.

rafasc:
wrt `aur-vercmp-devel` I have mixed feelings about it being longer than the script itself too. One of the reasons this PR doesn't include that commit.

Allowing to pass it as a environment variables is probably a viable solution. As we already allow passing `AURDEST` and `AURVCS`. 

Or alternatively, adding options for setting those vars might help justify the size of the option handling. Along with adding an option to output a `cut -d: -f1` for easy piping into `sync` for example.

But that is something that users can decide for themselves by writing their own scripts if we go forward with moving `aur-vercmp-devel` to a "procedures" man page.

AladW:
Merged, I'll leave the `aur-vercmp-devel` topic to #511.

PR #525, opened 2019-02-16
aur-fetch: default to patch files in current directory

### LINKED ISSUE #522: aur-fetch: git diff and escape sequences
PKGBUILDs or other AUR files with embedded "text hide" escape sequence could trick the output from `git diff`. See https://github.com/Jguer/yay/issues/880, https://www.twistlock.com/labs-blog/hiding-content-git-escape-sequence-twistlock-labs-experiment/ and https://github.com/twistlock/gitPocDiff.

AladW:
Use the optional `--print` to print diffs to standard output.

Closes #522

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -8,34 +8,27 @@ readonly startdir=$PWD
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
 # default options
-mode=git recurse=0
+log=directory mode=git recurse=0
 
 git_stat_patch() {
     git --no-pager log -O "$1" --patch --stat '..@{upstream}'
 }
 
-merge_upstream() {
-    # set destination of patch file
-    local pkg=$1 log_dir=$2 output
+upstream_sync() {
+    local pkg=$1 patch=${2:-/dev/stdout}
+    local orderfile=${GIT_DIR:-$PWD/$pkg/.git}/orderfile
 
-    if [[ -d $log_dir ]]; then
-        output=$log_dir/$pkg.patch
-    else
-        output=/dev/stdout
-    fi
-
-    # default to showing PKGBUILD diffs first (#399)
-    local orderfile=${GIT_DIR:-$PWD/.git}/orderfile
-
+    # default to showing PKGBUILD in diffs first (#399)
     if [[ ! -s $orderfile ]]; then
         echo PKGBUILD >"$orderfile"
     fi
 
+    # get last upstream revision
     git fetch -v >&2
 
-    # only print log on upstream changes
     if [[ $(git rev-parse HEAD) != $(git rev-parse '@{upstream}') ]]; then
-        git_stat_patch "$orderfile" >> "$output"
+        # only print log on upstream changes
+        git_stat_patch "$orderfile" >> "$patch"
 
         # discard any local changes (#349)
         git reset --hard "HEAD@{upstream}" >&2
@@ -43,7 +36,7 @@ merge_upstream() {
 }
 
 fetch_git() {
-    local log_dir=$1 pkg uri
+    local uri pkg output_dir=$1
 
     while IFS= read -r uri; do
         pkg=${uri##*/}   # strip path
@@ -51,28 +44,16 @@ fetch_git() {
 
         if [[ -d $pkg/.git ]]; then
             # avoid issues with filesystem boundaries (#274)
-            GIT_DIR="$PWD/$pkg"/.git GIT_WORK_TREE="$PWD/$pkg" merge_upstream "$pkg" "$log_dir"
+            GIT_DIR="$pkg"/.git GIT_WORK_TREE="$pkg" upstream_sync "$pkg" "${output_dir:+$output_dir/$pkg.patch}"
         else
             git clone "$uri"
         fi
     done
 }
 
-diff_log() {
-    local base=$1 log_dir=$2
-
-    if [[ -d $log_dir ]]; then
-        tee -a "$log_dir/$base".diff | diffstat -Ckq -f3
-    else
-        tee
-    fi
-}
-
 tar_no_mode_diff() {
-    local archive=$1 base=$2 dir=$3
-
-    if [[ -d $dir/$base ]]; then
-        LC_MESSAGES=C tar -df "$archive" -C "$dir" | grep -Evq '(Mode|Uid|Gid)'
+    if [[ -d $1/$2 ]]; then
+        LC_MESSAGES=C tar -df "$3" -C "$1" | grep -Evq '(Mode|Uid|Gid)'
     else
         return 1
     fi
@@ -80,7 +61,7 @@ tar_no_mode_diff() {
 
 fetch_tar() {
     #global tmp
-    local log_dir=$1 archive base
+    local archive base output_dir=$1
 
     wget -i /dev/stdin --directory-prefix "$tmp" || return 1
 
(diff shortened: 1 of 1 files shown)


lib/aur-fetch
```diff
@@ -150,10 +149,16 @@ if ! (($#)); then
...
 fi
 
-if [[ -v log_dir ]] && [[ ! -d $log_dir ]]; then
-    printf >&2 '%s: %q: not a directory' "$argv0" "$log_dir"
-    exit 21
-fi
+case $log in
+    directory)
+	# store patches in current directory by default
+        if [[ ! -d ${patch_dir-$startdir} ]]; then
```
rafasc: You probably meant
```suggestion
        if [[ ! -d ${patch_dir:=$startdir} ]]; then
```

lib/aur-fetch
```diff
@@ -168,6 +173,6 @@ case $mode in
 esac
 
 # pipeline
-deps "$@" | uri | fetch_$mode "$log_dir"
+deps "$@" | uri | fetch_$mode "$patch_dir"
```
rafasc: Otherwise `$patch_dir` will be empty making it fallback to stdout output.

lib/aur-fetch
```diff
@@ -8,88 +8,86 @@ readonly startdir=$PWD
...
-    local pkg=$1 log_dir=$2 output
-
-    if [[ -d $log_dir ]]; then
-        output=$log_dir/$pkg.patch
-    else
-        output=/dev/stdout
-    fi
-
-    # default to showing PKGBUILD diffs first (#399)
+merge() {
```
rafasc: Nitpick: Slightly bothers me that we're calling the function `merge` when we're doing a `reset.` 🙃 

This is about getting in sync with upstream so maybe something like sync or upstream_sync?

AladW:
Addressed the comments, thanks. Is there anything left or does it work as intended now?

lib/aur-fetch
```diff
@@ -8,88 +8,86 @@ readonly startdir=$PWD
...
     fi
 
+    # get last upstream revision
     git fetch -v >&2
 
-    # only print log on upstream changes
     if [[ $(git rev-parse HEAD) != $(git rev-parse '@{upstream}') ]]; then
-        git_stat_patch "$orderfile" >> "$output"
+	# only print log on upstream changes
+	if [[ -f $1 ]]; then
```
rafasc: First argument is the pkgname, second argument is the log destination.
So this check should be done against `$2`, not `$1`

```suggestion
	if [[ -f $2 ]]; then
```

lib/aur-fetch
```diff
@@ -8,88 +8,86 @@ readonly startdir=$PWD
...
+            git_stat_patch "$orderfile" >> "$1"
```
rafasc: Same here
```suggestion
            git_stat_patch "$orderfile" >> "$2"
```

It's easy to get lost when all the checks are done across the calls. Maybe we need to do something about that.

rafasc:
I focused at `fetch_git` and `aur fetch` as a standalone tool. Still didn't look into the tar part.

lib/aur-fetch
rafasc: the `-f` test isn't sufficient though, because the file doesn't exist the first time we run `fetch`. But in that case we still want to write into it and not fallback to stdout.

lib/aur-fetch
```diff
@@ -150,10 +140,16 @@ if ! (($#)); then
...
 
-if [[ -v log_dir ]] && [[ ! -d $log_dir ]]; then
-    printf >&2 '%s: %q: not a directory' "$argv0" "$log_dir"
-    exit 21
-fi
+case $log in
+    directory)
+	# store patches in current directory by default
+        if [[ ! -d ${patch_dir:=$startdir} ]]; then
+            error "$argv0: '$patch_dir': not a directory"
```
rafasc: `error()` is used but `source /usr/share/makepkg/util/message.sh` is missing.

output is monochromatic due to lack of 
```
if [[ -t 2 && ! -o xtrace ]]; then
    colorize
fi
```

Message is missing a trailing new line. 

```suggestion
            error '%s: %s\n: not a directory' "$argv0" "$patch_dir"
```

lib/aur-fetch
```diff
@@ -8,88 +8,77 @@ readonly startdir=$PWD
...
 
     for archive in "$tmp"/*.tar.gz; do
         base=$(basename "$archive" .tar.gz)
         tar -kxf "$archive" --directory "$tmp"
 
-        if tar_no_mode_diff "$archive" "$base" "$startdir"; then
-            diff -ur "$startdir/$base" "$tmp/$base" | diff_log "$base" "$log_dir"
+        if tar_no_mode_diff "$startdir" "$base" "$archive"; then
+            if [[ -d $1 ]]; then
+                diff -ur "$startdir/$base" "$tmp/$base" | diff_log "$1" "$base"
```
rafasc: diffstat is going to stdout. Is this intended?

lib/aur-fetch
```diff
@@ -8,88 +8,77 @@ readonly startdir=$PWD
...
+    local uri pkg output_dir=$1
 
     while IFS= read -r uri; do
         pkg=${uri##*/}   # strip path
         pkg=${pkg%%.git} # strip .git suffix
 
         if [[ -d $pkg/.git ]]; then
             # avoid issues with filesystem boundaries (#274)
-            GIT_DIR="$PWD/$pkg"/.git GIT_WORK_TREE="$PWD/$pkg" merge_upstream "$pkg" "$log_dir"
+            GIT_DIR="$PWD/$pkg"/.git GIT_WORK_TREE="$PWD/$pkg" upstream_sync "$pkg" "${output_dir:+$output_dir/$pkg.patch}"
```
rafasc: these `$PWD` variables can be removed. The `git` bug that required it is already fixed.
We may take this chance to get rid of it.

lib/aur-fetch
AladW: In this case yes, since `diffstat` will only print a summary to `stdout`.

lib/aur-fetch
AladW: `error` doesn't take formatters - it roughly corresponds to `printf %s\n' - but the output should be colorized in any case.

lib/aur-fetch
```diff
@@ -104,9 +104,13 @@ EOF
     exit 1
 }
 
-source /usr/share/makepkg/util/util.sh
```
rafasc: util.sh is still needed because `canonicalize_path` is used in the `-L` parseopt handling. L127

lib/aur-fetch
rafasc: It's personal preference, I don't mind either style, but error does work with formatters. E.g. https://git.archlinux.org/pacman.git/tree/scripts/makepkg.sh.in#n246 and in the source from error itself you can see that "$@" is passed through.

lib/aur-fetch
AladW: Does https://github.com/AladW/aurutils/pull/525/commits/ebbdc5083f635cc3beea9f9aee1f1944c3d75cd9 make it more clear?

rafasc: IMHO, I think it does.

AladW: :+1:

AladW:
To do:
* Rename `--log-dir` to `--patch-dir`?
* Update `aur-fetch(1)`
* ?

rafasc:
`--log-dir` and `--patch-dir` makes me think I am logging/patching this directory. Especially the `patch-dir` for some reason. It took me a second glance to figure out this PR was about writing logs and not about patching files.

So maybe `--write-log`?

I think either `--print`, `--stdout` are ok. But I think I like `--verbose` the best. Teaching the above to understand something like `--write-log=/dev/stdout` could be another choice.

AladW:
Would `--write-log` imply that by default, no `.patch` files are saved? (I'd say this is also a reasonable thing to do)

AladW:
Since I couldn't bother to figure it out in bash, I rewrote `aur-fetch` in perl: http://ix.io/1CWP/perl

Lacks the tar related functionality, but when using `parallel` with `aur-fetch` there's no point anyway (and issues like #379 can only be properly addressed with `git`).

AladW:
In addition, it seems perl's `print` does not output escape sequences by default.

AladW:
Superseded by #548

PR #526, opened 2019-02-16
completion: generate completion for aur-depends

rafasc:
In cadf3f7 (aur-depends: implement --pkgbase, --pkgname, --table,
2019-01-14) aur depends gained option handling but completion for it is
not being generated.

I noticed that there was a mismatch in number of scripts using parseopts
and the number of scripts with autocomplete, but because the macro
arguments were not lexicographically sorted spotting the missing one
took more seconds than it should.

So lets sort the arguments and move depends to HAVE_OPTDUMP so
completions are generated for it.


AladW:
:+1:

PR #527, opened 2019-02-16
stop unsetting parseopt helper variables

rafasc:
opt_hidden is not being unset after option handling in all but one
script. As this seems like an unnecessary step, instead of also
unsetting opt_hidden everywhere let's just stop unsetting those helper
variables.

This was brought up in #aurutils. Is there any real benefits from unsetting those variables here? If not, let's just remove all those lines. If yes, this can be converted into a PR that fixes all the missing `unset opt_hidden`.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -90,7 +90,6 @@ while true; do
     esac
     shift
 done
-unset opt_short opt_long OPTRET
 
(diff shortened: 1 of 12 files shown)


AladW:
I'll merge this, having `parseopts` options `unset` led to mistakes like https://github.com/AladW/aurutils/commit/927e22cfae554c91f5f4fca98bb257d414a1376e and none of these variables are exported in the first place. Thanks.

PR #528, opened 2019-02-17
`repo-add` build delta

bartoszek:
Building deltas is a _time hog_, and deltas are pretty much never used in a small network environment.
It would be nice to have this feature turned off by default, and only activated with a flag.


AladW:
I don't disagree on using a flag on adding deltas, but it seems some of the changes were mixed up?

bartoszek:
Yep, Gave me a moment to tackle those out.

PR #529, opened 2019-02-17
aur-build: make repo-add -d optional

bartoszek:
Building deltas is a _time hog_, and deltas are pretty much never used in a small network environment.
It would be nice to have this feature turned off by default, and only activated with a flag. 

_can't reopen  #528, after force push :disappointed:_

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -107,7 +107,7 @@ if (($#)); then
     fi
 fi
 
-if type -P xdelta3 >/dev/null && ((delta)); then
+if ((delta)); then
     repo_add_args+=(-d)
 fi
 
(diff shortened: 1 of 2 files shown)


lib/aur-build
```diff
@@ -106,7 +107,7 @@ if (($#)); then
     fi
 fi
 
-if type -P xdelta3 >/dev/null; then
+if type -P xdelta3 >/dev/null && ((delta)); then
```
AladW: If there's a flag, we can probably leave out the `type -P`

AladW:
With a rebase and the above change I'll merge it

lib/aur-build
bartoszek: dropped `type -P` conditional
added disclaimer to `aur-build` man page pointing to `xdelta3`

PR #532, opened 2019-02-21
sync: ask for confirmation when using PAGER

### LINKED ISSUE #530: aur-sync: file inspection with PAGER="less -F" does not do the right thing
This is just a minor thing since using `less` for file inspection is not the best thing for me anyway, but still: if aur-sync decides to use `$PAGER` for file inspection which contains `less` using the `-F`/`--quit-if-one-screen` option, the build process does not pause before the inspection.

rafasc:
`aur sync` will use $PAGER if `vifm` is not avaiable, falling back to
`less -K` if $PAGER is unset. `aur sync` uses the exit code from this
invocation to decide if the build process should continue or should be
aborted. This is a important part of the process as continuing implies
the user reviewed and agreed to run the PKGBUILD.

While vifm and the current default (less -K) allows exiting with error
by using :cq or ^C respectively, if the user has $PAGER set to something
that doesn't wait for user input and provide a way to exit with error,
or has LESS=F which causes `less` to exit successfully if the output
fits in one page, aur sync will proceed without explicitly consent from
the user.

Let's override the -F option on our default fallback making sure it
always paginate, and prompt for confirmation whenever a custom $PAGER is
set in order to always have explicit consent from the user and a way to
abort if desired.

Fixes #530

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -264,7 +264,7 @@ if type -P >/dev/null vifm; then
     viewer() ( cd "$1"; vifm -c 'view!' -c '0' - )
 else
     # shellcheck disable=SC2086
-    viewer() ( cd "$1"; command -- ${PAGER:-less -K} )
+    viewer() ( cd "$1"; command -- ${PAGER:-less -K -+F} )
 fi
 
 # check if dependency graph is valid
@@ -283,6 +283,11 @@ if ((view)); then
           # print build directories in dependency order
           xargs -I{} -a "$tmp"/queue find -L "$tmp_view"/{} -maxdepth 1
         } | viewer "$tmp_view"
+
+        # Fix for #530
+        if [[ -v PAGER ]]; then
+            read -p $'Press any key to continue or ctrl+d to abort\n'
+        fi
     fi
 fi
 


AladW:
Thanks. Would you mind adding a comment above `[[ -v PAGER ]]` referencing #530?

rafasc:
@AladW done.

AladW:
Merged (note: I changed the "fix for #530" by the commit title)

lib/aur-sync
```diff
@@ -264,7 +264,7 @@ if type -P >/dev/null vifm; then
     viewer() ( cd "$1"; vifm -c 'view!' -c '0' - )
 else
     # shellcheck disable=SC2086
-    viewer() ( cd "$1"; command -- ${PAGER:-less -K} )
+    viewer() ( cd "$1"; command -- ${PAGER:-less -K -+F} )
```
maximbaz: @rafasc is `-+F` a typo, should it be just `-F`? I believe `+F` means "follow", i.e. never exit, while `-F` is indeed exit if it fits on one screen.

rafasc: No, it's not a typo. The idea is that we *do not* want to quit if one screen, even if the user has `LESS=F`. The `-+` is a mechanism to override options set by the LESS var.

excerpt from `man less` for convenience:
> The environment variable is parsed before the command line, so command line options override the LESS environment variable.  If an option appears in
>        the LESS variable, it can be reset to its default value on the command line by beginning the command line option with "-+".

maximbaz: Wow, I see, thanks :)

rafasc: Thank you for keeping an eye on things. :)

PR #533, opened 2019-02-21
Makefile: quote variables to prevent wordsplit

rafasc:
I came across this while debugging and trying to install aurutils manually on a `DESTDIR` that happened to contain a space.


AladW:
:+1:

PR #535, opened 2019-02-27
aur: allow overriding aur lib_dir at build time.

rafasc:
To facilitate the customization of the aur wrapper, allow overriding the
lib_dir that becomes hardcoded in `aur`.

This can be useful in situations where the user wants to have multiple
installs running from different lib_dir's. While you can, this should
not be used to bypass the limitation that you cannot override default
scripts using aur-scripts in your $PATH.

If you do it, do it at your own risk.


AladW:
:+1:

PR #537, opened 2019-03-01
sync: make confirmation prompt actually allow for any key to be pressed to continue

terencode:
Credits to https://unix.stackexchange.com/a/293941

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -286,7 +286,7 @@ if ((view)); then
 
         # ask for confirmation when using PAGER (#530)
         if [[ -v PAGER ]]; then
-            read -p $'Press any key to continue or ctrl+d to abort\n'
+            read -n 1 -s -r -p $'Press any key to continue or ctrl+d to abort\n'
         fi
     fi
 fi


rafasc:
I'll vote against this as it is. I Feel this makes it way too easy to agree by accident. 

Important operations that ask for user confirmation usually ask for a explicit enter.  
cf. `touch temp; rm -vi temp;` I also see no value in hiding the answer.

We could accept any answer that starts with `y` and `exit` otherwise:
```sh
read -p 'Do you want continue? [y/N] '
[[ ! $REPLY =~ ^[Yy] ]] && exit 1
```

Earnestly:
No prompt system should respond to anything other than `Return`.

terencode:
@Earneslty @rafasc sure but that's not one the message says. This PR only makes it match the original prompt message.

rafasc:
@terencode Maybe the message is the thing that needs changing. 😏
I already suggested changing it to a more traditional `[y/N]` prompt. Feel free to re-spin this PR into one that does that.

PR #539, opened 2019-03-03
aur-repo: --repo-list: use full paths instead of tsv output

AladW:
This has two benefits:
1. Consistent with `aur-repo` when invoked without arguments, which only lists a path.
2. Makes looping over the output easier, i.e. does not require `cut`.

The repository name is simply the database name without `.db` extension, and can be retrieved with `basename`:
```
aur repo --repo-list | while read -r; do
  name=$(basename "$REPLY" .db)
done
```

### DIFF
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -76,7 +76,7 @@ while read -r key _ value; do
             ;;
         Server=file://*)
             server=${value#file://}
-            conf_repo+=("$section" "$server")
+            conf_repo+=("$server" "$section")
 
             case $section in
                 $db_name)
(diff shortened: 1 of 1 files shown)


AladW:
To reflect 1, the `--repo-list` option could also be renamed to `--all`. (`aur repo --repo-list` is somewhat tautological anyway...)

AladW:
For comparison of the "ambigous" output, see https://i.imgur.com/OjtmTkf.png

It mainly removed the need for the complicated while/read loop in https://github.com/AladW/aurutils/pull/539/commits/b002f4ba549f0c412cf2536822fda4b67598be35 (and puts the repo names in direct view on my terminal). If anyone feels strongly about this, you can restore the old output with `column -O 2,1`.

PR #544, opened 2019-03-06
Added pacman.conf overrides to aur-chroot and aur-repo

Foxboron:
Fixes #543

PR #545, opened 2019-03-07
aur-chroot: add --host-conf

AladW:
PR #544 added `--config` to `host_with_cache`, it however used the same
pacman config as used in the chroot. In particular, `CacheDir` and
local repositories on the host would not be added.

Besides some stylistic changes, this PR also fixes the long options
which were missing an argument.

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -17,21 +17,28 @@ bindmounts_ro=()
 # default options
 prepare=1 build=1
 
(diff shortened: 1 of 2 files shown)


AladW:
Tested the `pacman.conf` in the container, though `CacheDir` will not work as expected due to `arch-nspawn` hardcoding `pacman-conf -v`.

PR #548, opened 2019-03-14
aur-fetch: remove --tar support

AladW:
This is based on a `sh` rewrite by @Earnestly of #525, in particular https://github.com/AladW/aurutils/pull/525#issuecomment-471142788. The version in this PR mainly adds some boilerplate for completions. The changes to the current `aur-fetch` are:

* Remove tar support. It was originally added due to git pull being slow on high workloads, but since using `parallel` in `aur-sync` the difference is small.
* Add `--verbose` and `--write-log` to print diffs to stdout and to `.patch` files in a directory, respectively. Both options may be combined.
* Save the git orderfile for all cloned `git` repositories in `$XDG_CONFIG_HOME/aurutils/fetch`.

The man page is not yet complete and should include a description on how diffs are generated and resets are done.

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -1,173 +1,95 @@
 #!/bin/bash
 # aur-fetch - retrieve build files from the AUR
 readonly argv0=fetch
-readonly AUR_LOCATION=${AUR_LOCATION:-'https://aur.archlinux.org'}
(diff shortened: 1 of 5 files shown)


Earnestly:
I'd probably check `$#` instead of `$*` if the makepkg implementation of `getopt` properly shifts off the arguments at the end.

Issue #550, opened 2019-03-17
aur-fetch.1: update for --tar removal


AladW:
Include some information on diffs/branches and `git reset` as well.

AladW:
#563, details can be added later.

PR #551, opened 2019-03-20
aur fetch: fail if history was rewritten.

rafasc:
The AUR enforces a ff-only policy for PKBUILDS, however, in rare
occasions TUs will rewrite and force-push. Prime exemple being removal
of commits that introduce malicious code to the package.

Refuse to fetch in such scenario. When this happens, the user should
investigate why history got rewritten and take appropriate measures.

The check can be bypassed by the new option `--force`.

Users should update any existing cached repos to make sure they have the
non-fastforward refspec:

    find "${AURDEST}" -name .git -execdir git config remote.origin.fetch 'refs/heads/*:refs/remotes/origin/*' \;

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -86,7 +86,10 @@ fi | while read -r pkg; do
             git reset --hard 'HEAD@{upstream}' >&2
         fi
     else
-        if ! git clone -c diff.orderFile="$orderfile" "$AUR_LOCATION"/"$pkg".git; then
+        if git clone "$AUR_LOCATION"/"$pkg".git; then
+            # show PKGBUILDS first (#399)
+            git -C "$pkg" config diff.orderFile "$orderfile"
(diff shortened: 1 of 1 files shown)


lib/aur-fetch
```diff
@@ -33,6 +33,7 @@ while true; do
         -L|--write-log) shift; log_dir=$1 ;;
         -r|--recurse)   recurse=1 ;;
         -v|--verbose)   verbose=1 ;;
+        --force)        fetch_args+=('--force') ;;
```
AladW: Almost forgot: Maybe add an `-f` short option as well?

Morganamilo: That's what the pacman devs once thought ;)

AladW:
:+1:

Probably needs to be documented in the new man page as well, see #550...

PR #552, opened 2019-03-20
 fetch: preserve committed modifications to package 

rafasc:
On bc8d5c8 (fetch-git: discard local changes ... 2018-04-12)
aurutils took a holistic approach when it comes to allowing users
modifying files inside cached repos. Discarding any changes when a
package is to be updated.

This obviously isn't very practical, as some packages require
adjustments because they're either broken or the user wants to tune a
particular aspect of it.

Another concern was that automating non-fast-forward merges could end up
keeping commits that TU wrote out of history. But that being addressed
by ad8723d (fetch: refuse to fetch non-fast-forward refs, 2019-03-18),
allow us take a different approach:

Discard any uncommitted changes but reapply any local commits on top of
the incoming changes.

Only local user commits have the potential to generate conflicts when
applied, in which case aur fetch will fail with error.

This should allow us to have user persistent modifications and be in
sync with AUR with minimal chances for conflicts.

prerequisite: #551

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -78,21 +78,25 @@ fi | while read -r pkg; do
         # Avoid issues with filesystem boundaries. (#274)
         export GIT_DIR=$pkg/.git GIT_WORK_TREE=$pkg
 
-        git fetch "${fetch_args[@]}" || exit 1
+        # Discard any local uncommited changes. (#552)
+        git reset --hard HEAD >&2
+        git pull --rebase "${fetch_args[@]}" || exit 1
 
-        if [[ $(git rev-parse HEAD) != $(git rev-parse '@{upstream}') ]]; then
+        if ! git diff-files --diff-filter=U --quiet; then
+            warning '%s: Local commits conflict with upstream. Fix the conflicts and commit them.' "$pkg"
+            exit 1
+        fi
+
+        if [[ $(git rev-parse ORIG_HEAD) != $(git rev-parse HEAD) ]]; then
             # Only print log on upstream changes.
             if (( verbose )); then
-                git --no-pager log --patch --stat '..@{upstream}'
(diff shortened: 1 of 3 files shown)


rafasc:
This PR doesn't handle the cases where fetch fails.
One of the most noticeable problems is that you may miss `package.patch` from unrelated patches when `sync` fails. 

The patches are generated, but if `aur fetch` fails, it causes `aur sync` to abort, you never get to see them. The next time sync is ran, the cache is already up to date, so no `package.patch` are generated. Work to fix those issues is still work in progress. [Preview][1] for those interested.

[1]:https://github.com/rafasc/aurutils/compare/ra/fetch-keep-local-commits-rebased...rafasc:ra/differentiate-new-and-no-change-pkgbuilds

AladW:
Sorry for the delay. The changes look great to me. Could you rebase it due to #563?

rafasc:
Done.

AladW:
👍

lib/aur-fetch
```diff
@@ -78,24 +83,41 @@ fi | while read -r pkg; do
...
             if (( verbose )); then
-                git --no-pager log --patch --stat '..@{upstream}'
+                git --no-pager log --patch --stat "${log_range}"
             fi
 
             if [[ $log_dir ]]; then
-                git --no-pager log --patch --stat '..@{upstream}' > "$log_dir/$pkg".patch
-                printf '%s/%s.patch\n' "$log_dir" "$GIT_DIR"
+                logfile="$log_dir/$pkg".patch
+                gen_patchfile "${log_range[@]}" >"$logfile"
```
AladW: I missed this in my initial review, but  `gen_patchfile` is undefined (it's used in #565)

rafasc: (ノಠ益ಠ)ノ彡┻━┻
Hotfix in #566

PR #553, opened 2019-03-27
make the usage function consistent across scripts

rafasc:
Make the usage function print the usage string to stderr and
exit with error.

### DIFF
--- a/lib/aur-search
+++ b/lib/aur-search
@@ -106,7 +106,7 @@ trap_exit() {
 }
 
 usage() {
-    printf >&2 'usage: %s: [-adimnqrsv] [-k key] pkgname...\n' "$argv0"
+    plain 'usage: %s [-adimnqrsv] [-k key] pkgname...' "$argv0" >&2
     exit 1
 }
 
--- a/lib/aur-vercmp
+++ b/lib/aur-vercmp
@@ -68,7 +68,7 @@ trap_exit() {
 }
 
 usage() {
-    plain "usage: $argv0 [-p path] [-acq]" >&2
+    plain 'usage: %s [-p path] [-acq]' "$argv0" >&2
     exit 1
 }
 
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -39,7 +39,7 @@ trap_exit() {
 }
 
 usage() {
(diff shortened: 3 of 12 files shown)


lib/aur-vercmp
```diff
@@ -68,7 +68,7 @@ trap_exit() {
 }
 
 usage() {
-    plain "usage: $argv0 [-p path] [-acq]" >&2
+    plain 'usage: %s [-p path] [-acq]' "$argv0" >&2
```
AladW: I guess if we're aiming for consistency and add `%s` everywhere (less shellcheck complaints?), the same should be done for `error`/`warning`

lib/aur-search
```diff
@@ -106,7 +106,7 @@ trap_exit() {
 }
 
 usage() {
-    printf >&2 'usage: %s: [-adimnqrsv] [-k key] pkgname...\n' "$argv0"
+    plain 'usage: %s [-adimnqrsv] [-k key] pkgname...' "$argv0" >&2
```
AladW: Maybe add a comment in the relevant `[[ -t 1 ]]` section too (re. #473, as a reminder to fix it in a different PR)

lib/aur-vercmp
rafasc: > I guess if we're aiming for consistency and add %s everywhere (less shellcheck complaints?)

I don't think shellcheck complains about having the format:
```diff
-    plain "usage: $argv0 [-p path] [-acq]" >&2
```

I just picked one and went with it. If you prefer the this, I can easily change all to this format. I only care about being consistent across the board.

> the same should be done for error/warning
Agreed, The original intent for this PR was to make `usage()` consistent. So I restricted to changes to usage.

We can rename this PR something else and add commits fixing there rest.

lib/aur-search
rafasc: Out of scope here. I'll open an issue to keep track of it.

lib/aur-vercmp
AladW: Works for me as is. We can make the changes to plain et al in a second PR.

AladW:
Self-requested, really... :roll_eyes:

PR #560, opened 2019-04-21
Add manjaro chroot marker file

bartoszek:
On manjaro `mkarchroot` creates different root marker file.
Arch=.arch-chroot, Manjaro=.manjaro-chroot

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -103,7 +103,7 @@ if ((prepare)); then
     # do not print DBPath (#412)
     pacman_conf_raw "$pacman_conf" | cat - "$tmp"/custom.conf > "$tmp"/pacman.conf
 
-    if [[ -f $directory/root/.arch-chroot ]]; then
+    if [[ -f /tmp/x86_64/root/.arch-chroot || -f /tmp/x86_64/root/.manjaro-chroot ]]; then
         # locking is done by systemd-nspawn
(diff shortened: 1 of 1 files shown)


rafasc:
While this looks like a very simple change, `aurutils` is written for the [Arch User Repository](https://wiki.archlinux.org/index.php/Arch_User_Repository) and depends on the Arch Linux version of dependencies, in this case: `devtools`. Keeping track of what Arch based distros are doing with their own fork of Arch packages is hard to maintain, as mentioned in the [Arch Linux Code of Conduct](https://wiki.archlinux.org/index.php/Code_of_conduct#Arch_Linux_distribution_support_*only*).

I'd say this is a ![wontfix][wontfix], but @AladW has obviously the final say.

[wontfix]:https://img.shields.io/badge/-wontfix-a8a8a8.svg

Foxboron:
I think the comment from Dave is applicable here as well; https://github.com/archlinux/devtools/pull/9

Earnestly:
Yes, this is the wrong place to fix it.  `mkarchroot` is unnecessarily not general purpose, so anything to work around this issue will just require more and more code instead of less.

bartoszek:
Perhaps we could have this resolved in `prepare` function directly in `PKGBUILD` on aur ?

bartoszek:
Now i see it, thanks for clarifying :blush:

PR #562, opened 2019-04-23
Do not prompt after vifm review when PAGER is set

rafasc:
aurutils should trust `vifm` exit codes as a signal that build files
were reviewed. But (532), the fix for (530), does not properly handle
all use cases. In particular, it asks for confirmation after review in
`vifm` if PAGER is set, when PAGER wasn't used in the first place.
Fix that.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -255,12 +255,19 @@ fi
 # link build files in the queue (absolute links)
 aur jobs -X ln -s "$(pwd -P)"/{} "$tmp_view" :::: "$tmp"/queue
 
+# Explicitly define if exit success is to be trusted (#561)
+trust_exit_success=0
 if type -P >/dev/null vifm; then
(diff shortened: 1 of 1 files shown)


Wuestengecko:
You're missing a semicolon on line 268:

```
    if [[ -z $PAGER ]] then
```

Other than that, it works. Thanks!

rafasc:
@Wuestengecko Thanks for the catch.

AladW:
You probably don't need the `trust_exit_success` if you move the `if [[ -z $PAGER ]]` directly inside the `viewer` function

rafasc:
`viewer()`'s `stdin` is the pipe that generates the file list. So you cannot prompt and `read` the answer in case of an arbitrary pager being used. Unless I am missing something...

AladW:
I guess that a faulty PAGER may not read all of `stdin`, which then affects the `read` prompt.

PR #564, opened 2019-04-30
scripts: respect environment variable AUR_COLOR, part 2

rafasc:
Some of the scripts failed to receive the #557 treatment,
9dba6dd (scripts: respect environment variable AUR_COLOR, 2019-04-15).

Correct that.


rafasc:
For some unknown reason, I missed these  scripts in the original PR. :unamused:

PR #565, opened 2019-05-01
fetch: Change patch format

rafasc:
Available formats:
    - log: show all commits between HEAD and last seen.
    - diff: show differences between HEAD and last seen.

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -51,12 +51,12 @@ while true; do
 done
 
 if [[ -v $log_dir ]] && [[ ! -d $log_dir ]]; then
-    printf 'fetch: %s: Not a directory\n' "$log_dir" >&2
-    exit 1
+    error '%s: %s: Not a directory' "$argv0" "$log_dir"
+    exit 20
 fi
 
 if ! (( $# )); then
-    printf 'fetch: No pkgname given\n' >&2
+    error '%s: No pkgname given' "$argv0"
     exit 1
 fi
 
@@ -123,7 +123,7 @@ fi | while read -r pkg; do
             # only allow fast-forward fetches. (#)
             git -C "$pkg" config remote.origin.fetch 'refs/heads/*:refs/remotes/origin/*'
         else
-            printf 'fetch: %s: Failed to clone repository\n' "$pkg" >&2
+            error '%s: %s: Failed to clone repository' "$argv0" "$pkg"
             exit 1
         fi
     fi
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -91,10 +91,8 @@ fi | while read -r pkg; do
 
         # Discard any local uncommited changes. (#552)
         git reset --hard HEAD >&2
-        git pull --rebase "${fetch_args[@]}" || exit 1
-
-        if ! git diff-files --diff-filter=U --quiet; then
-            warning '%s: Local commits conflict with upstream. Fix the conflicts and commit them.' "$pkg"
+        if ! git pull --rebase "${fetch_args[@]}"; then
+            error '%s: Failed to integrate changes.' "$pkg"
             exit 1
         fi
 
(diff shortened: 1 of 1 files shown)


rafasc:
In this PR, the new `diff` format is made default for the sake of testing. That decision is up to discussion. So are the formats  themselves, as both have pros and cons.

Maybe the ideal would be having a combination of these two: `diff` at the top, some kind of separator, followed by `log`.

Depending on feedback, this may be postponed until the next milestone.

Cc: @maximbaz as he helped me test this feature.

AladW:
Could you maybe add some screenshots on the different formats?

rafasc:
Consider you run fetch, and there are 3 new commits. 

The `diff` format will generate a single diff considering the before and after, while `log` will show the current `git log` based patch, which shows 1 diff per commit, and its respective message.

I am not sure if a screenshot is the best way to show the difference, but here's an attempt:
formats

`diff` results in a much shorter patch, making the review easier at the expense of not showing intermediate changes.

rafasc:
As it is impossible to cover everyone's cases, another interesting idea would be to allow a format `git-<name>`; where `gen_patchfile` would call `$git name;`, allowing users to define `git` aliases that control the format.

We could support a couple of sane defaults, and leave the rest for the user to implement as an alias.

Or go the other route, and use a predetermined `git alias` that if found would be used, falling back to what we do currently.

AladW:
From a purely personal perspective, I prefer the `diff` format since I'm usually not very interested in scrolling the page to view all commits. Especially when every commit includes changes to .SRCINFO as well...

I guess a `git alias` could be placed in another environment variable, though I'm unsure if we need this additional granularity. It's hard to compete to dedicated GUIs like `tig` anyway, e.g.

tig

rafasc:
Since I also prefer `diff` myself, maybe we can simplify the logic and use a predetermined alias if it exists, falling back to `diff` mode. I.e.:

``` sh
if git config --get alias.aurutils-patch-gen; then
    git aurutils-patch-gen "${log_range[@]}"
else
    git diff ... "${log_range[@]}"
fi
```
The current mode (`log`), could be achieved via the `git alias` mechanism if desired.

The user can configure the `git alias` globally in their `~/.gitconfig` , or per repo if they want. And since git aliases can be shell functions, that gives it a lot of flexibility and removes our burden to support an endless list of formats to suit everyone's needs.

`vifm` can be made to launch `tig` automatically with "`:autocmd DirEnter .git !tig`" in `vifmrc` but, unfortunately, doing it destroys the `vifm view` we create.

For those reasons, I think there's value in supporting the `git alias` feature.

maximbaz:
I also prefer `diff`, maybe we should just switch to `diff` entirely and wait with further generalizations until we actually hear from someone not happy that `log` got replaced with `diff`?

rafasc:
I was trying to not lose the feature of being able to conveniently read the log.

But since `vifm` can be made to display repository entries in a friendlier way with:
```
autocmd DirEnter ~/.cache/aurutils/sync/*/.git,/tmp/view.????????/*/.git !tig
fileviewer {{~/.cache/aurutils/sync/*/.git/,/tmp/view.????????/*/.git/}}
           \ git -C %c log --color HEAD $(git -C %c rev-parse --revs-only ^AURUTILS_SEEN)
```
there is no net-loss from dropping `"log"`.

The above could be added as documentation (after changing some of the hardcoded values with their correspondent environment variables) in a PR that also makes the default view tmpdir a bit more recognizable as coming from aurutils. (not /tmp/view.XXXXXXXX)

Thanks for the feedback.

lib/aur-fetch
```diff
@@ -9,6 +9,10 @@ readonly AUR_COLOR=${AUR_COLOR-auto}
 # default options
 verbose=0 recurse=0 fetch_args=('--verbose') confirm_seen=0
 
+gen_patchfile() {
+    git diff-tree --patch --stat "${2:-$(git hash-object -t tree /dev/null)}" "$1";
```
AladW: Nitpick: redundant semi-colon at the end (same for L101/103)

AladW:
:+1:

PR #566, opened 2019-05-02
fetch: remove use of undefined function gen_patchfile

rafasc:
`gen_patchfile` is defined in PR-565 and was used here by mistake.

Correct that by reverting to the explict git call.


AladW:
Thanks.

rafasc:
I suspect this will not be the last one related to those changes.

PR #572, opened 2019-06-12
Add aurutils logo to repository

AladW:
It used to be hosted on ptpb, which has been defunct for a while.


AladW:
Yay! It's now nice again to look at `aurutils`. :blush:

Issue #570, opened 2019-06-07
aur-remove suggestion in aur(1) wrongfully deletes files


rafasc:
In `aur(1)` we suggest the following script:
https://github.com/AladW/aurutils/blob/6da4f82975efdc0faa52dc9ba15274b00211a7ff/man1/aur.1#L383-L398

This script has a bug that results in data loss:

- `package=$1` will be empty if the user runs the script without arguments, which results in `-name "$package*"` expanding to `-name "*"` deleting all packages and the database itself.
- If the user has packages that share the same prefix, e.g. `aurutils` and `aurutils-git`,  
`aur remove aurutils` will remove `aurutils` from the db but delete `aurutils` *and* `aurutils-git` packages from disk.

Earnestly:
Yes, apart from checking `$package` is non-empty, the `find` command would need to use `-exec rm -rf -- {} +` instead of `-delete`.  With this you can also add `-type d`.

rafasc:
The problem isn't using <code>find -name *\<pattern>* -delete</code>, it's about making `find` only match the intended packages.

As far as I am aware, it needs to match the following format:
`$pkgname-$version-$pkgrel-$architecture$PKGEXT`

The closest I could get on a short notice was:
```
find -regex '.*/$pkgname-[^-]+-[^-]+-[^-]+'
                           \    \     \_____ $architecture$PKGEXT
                            \    \__________ $pkgrel
                             \______________ $version
```
Didn't have much time to think of a more elegant solution than this..

I think your suggestion could be referring to files in `$AURDEST` (`.cache/aurutils/sync`), which this script doesn't even attempt to remove, but probably should.

Earnestly:
You might be able to derive the package name from `makepkg --packagelist`.

AladW:
That only gives the version of whatever version described by the PKGBUILD though, besides that calling `makepkg` may be out of scope for such a tool.

Does https://github.com/AladW/aurutils/issues/570#issuecomment-500056042 take `epoch` into account?

rafasc:
> Does [#570 (comment)](https://github.com/AladW/aurutils/issues/570#issuecomment-500056042) take `epoch` into account?

It should, as neither `epoch` or `pkgver` can contain the `-` character, it would be picked up by the `$version` part of the pattern.

Earnestly:
A better solution might be to use `paccache` which already deals with all of this logic, e.g.:

```sh
#!/bin/sh --
[...]

PR #573, opened 2019-06-21
man1: Improve suggested remove script

rafasc:
The aur-remove script suggested in `aur(1)` has a bug that results in
data loss:

If the user runs the script without arguments, package=$1 will be empty,
which results in -name "$package*" expanding to -name "*" deleting all
packages and the database itself.

If the user has packages that share the same prefix, e.g. aurutils and
aurutils-git, aur remove aurutils will remove aurutils from the db but
delete aurutils and aurutils-git packages from disk.

Refactor the script using `paccache` from `[community]`.
Closes: #570 
Helped-by: Earnestly

### DIFF
--- a/man1/aur.1
+++ b/man1/aur.1
@@ -380,22 +380,20 @@ in $PATH, for example
 .BR aur\-remove
 
 .EX
-  #!/bin/bash
-  # alternative to 'repoctl remove' which loops over all local repositories
-  package=$1
+  #!/bin/sh --
+  # aur-remove - remove listed packages from all local repositories
+
+  if [ "$#" -eq 0 ]; then
+    printf 'usage: aur remove package [package ...]\en' >&2
+    exit 1
+  fi
 
   aur repo --repo-list | while read -r repo_path; do
-    # remove package from database
-    repo-remove "$(readlink -e "$repo_path")" "$package"
-
-    # remove all versions of package
-    find "$(dirname "$repo_path")" -name "$package*" -delete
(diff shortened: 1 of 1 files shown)


AladW:
FWIW I made a perl version before thinking about `paccache`: https://paste.xinu.at/DWtmF/

man1/aur.1
```diff
@@ -380,22 +380,20 @@ in $PATH, for example
...
-    # remove all versions of package
-    find "$(dirname "$repo_path")" -name "$package*" -delete
+  aur repo --repo-list | while read -r repo_path; do
+    if repo_path=$(readlink -e "$repo_path"); then
+      repo-remove "$repo_path" "$@"
+      paccache -c "${repo_path%/*}" -rvk0 "$@"
+    fi
   done
-
-  # remove package from system
```
AladW: I guess this is left to the user, or are the packages supposed to linger as foreign packages?

rafasc: It is a bit out of scope for a repo-management script to remove the packages on the host. 

This could be addressed by implementing `--uninstal` for example, but in the current context this is inserted, which is to show a minimal example of the custom script feature, that is left to the user as any other change they see fit.

Also missing from the script, and maybe also out of scope, is the removal of files in`.cache/aurutils/sync`.

cc: @Earnestly

Earnestly: Stop trying to conflate repo management and package management

AladW:
Thanks.

PR #579, opened 2019-07-25
scripts: retire aur-jobs

rafasc:
The aur-jobs wrapper does little more than remove the citation nagging,
which was since removed by upstream packager[1], thus we can retire
aur-jobs.

[1]:https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/parallel&id=e2abd3a81e6848bc18993ed878ba45ec4f39ef9d


Earnestly:
On the other hand it might be worth investigating the alternatives as the author suggests.  You maybe(?) don't need the edge case handling parallel purports to have.

AladW:
Opened #579 to discuss alternatives

PR #581, opened 2019-07-25
aur-repo: list devel packages

rafasc:


lib/aur-repo
```diff
@@ -7,12 +7,21 @@ readonly AUR_COLOR=${AUR_COLOR-auto}
...
         getline; pkgname=$1
-    } /%VERSION%/ {
-        getline; printf("%s\t%s\n", pkgname, $1)
+    }
+    /%BASE%/ {
+        getline; pkgbase=$1
+    }
+    /%VERSION%/ {
+        if (!devel ||
+            pkgbase ~ /-cvs$|-svn$|-git$|-hg$|-bzr$|-darcs$/ ||
```
Earnestly: You could assign `pkgname` to `pkgbase` (or another variable) and only do one comparson.

rafasc: I am testing both `pkgname` and `pkgbase` to be closer to the implementation that was done in https://lists.archlinux.org/pipermail/aur-dev/2019-May/004649.html . Was keeping this wip commit as a placeholder for a proper equivalent implementation.

But testing only the `pkgname` could be enough. These approaches are nothing more than heuristics that may fail to determine vcs packages correctly anyway.

We could keep things simple, test only pkgname and allow users to customize the pattern in order to minimize false positives/negatives. i.e. keeping the `$AURVCS` idea, letting users to hardcode corner cases. Although https://github.com/AladW/aurutils/issues/511#issuecomment-508938979 suggests getting rid it.

Thoughts?

Earnestly: Check just `pkgbase` for now as every package will have that.  Generally keep this easy to change because the real fix here is for the AUR RPC to report if a package is a VCS package as part of the info report.

lib/aur-srcver
```diff
@@ -21,11 +21,14 @@ srcver_pkgbuild_info() {
 }
 export -f srcver_pkgbuild_info
 
+source /usr/share/makepkg/util/message.sh  #    hack
```
AladW: This may be easier to implement in https://github.com/AladW/aurutils/pull/583

AladW:
FWIW https://wiki.archlinux.org/index.php/VCS_package_guidelines only mentions `pkgname`. In particular, while `pkgbase` does not necessarily have to contain a VCS suffix (why?), `pkgname` should for any valid VCS package.

From a practical perspective, `aur-srcver` always deals with `pkgbase` (it appears `vercmp-devel` was never tested with split packages...). I guess you could leave it to the user and add a `--pkgbase` option, with the VCS filter seperate.

AladW:
i.e. something like:
```

db_namever() {
    awk '/%NAME%/ {
        getline; pkgname=$1
    } /%VERSION%/ {
	getline; printf("%s\t%s\n", pkgname, $1)
    }'
}

db_basever() {
    awk '/%BASE/ {
	getline; pkgbase=$1
    } /%VERSION%/ {
	getline; printf("%s\t%s\n", pkgbase, $1)
    }'
}

vcs_check() {
    awk -v bool="$1" '!bool || $2 ~ 
    	/-cvs$|-svn$|-git$|-hg$|-bzr$|-darcs$/ {print}'
}
```

lib/aur-repo
```diff
@@ -3,16 +3,22 @@
 readonly argv0=repo
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 readonly AUR_COLOR=${AUR_COLOR-auto}
+readonly AURVCS=${AURVCS:-^.*-(cvs|svn|git|hg|bzr|darcs)$}
```
AladW: I would argue we don't need this for the simple reason that the user can do more complicated things already, i.e.
```
aur repo --list | pcregrep <some_morbid_perl_regex>
```

rafasc: With "this", do you mean the PR in general, or allowing to override `$AURVCS`?

We're already storing that for `--devel` use, allowing to override this has no extra maintenance.

But using the same logic, why implement `--devel` at all? We can just recommend

`aur repo --list | grep -P '^.*-(cvs|svn|git|hg|bzr|darcs)$'` on `aurvcs(7)`. And when the RPC adds proper support for marking packages as devel, we add `--devel`.

I'm fine with either.

AladW: I just meant AURVCS in this case. I guess the "warning" on VCS packages is about the closest we'll ever get to proper support for marking packages as devel.

maximbaz:
I didn't look much into the code, I simply took it for a spin, and `--pkgbase` behaves weirdly, it prints the same package multiple times with different versions. Here's an example, `--devel` is used to get shorter output, but it reproduces even without it.

```
❯ aur repo -l --devel          
antigen-git     v2.2.3.r6.gc91f77c-1
aurutils-git    2.3.1.r64.ga8432751-1
bingo-git       r929.76bcd77-1
connman_dmenu-git       67.d8e961e-1
dragon-drag-and-drop-git        1.1.0.r0.gf7f2121-1
duration-git    1.0.0.r4.g41f55b0-1
gtk-theme-arc-gruvbox-git       r12.e5ea0ff-1
kakmerge-git    18.580c832-1
lscolors-git    r314.6fe816f-1
xkblayout-state-git     20180812-1

❯ aur repo -l --devel --pkgbase
antigen-git     v2.2.3.r6.gc91f77c-1
aurutils-git    2.3.1.r64.ga8432751-1
bingo-git       r929.76bcd77-1
connman_dmenu-git       67.d8e961e-1
connman_dmenu-git       1.1.0-1
dragon-drag-and-drop-git        1.1.0.r0.gf7f2121-1
duration-git    1.0.0.r4.g41f55b0-1
duration-git    1-2
duration-git    3-1
gtk-theme-arc-gruvbox-git       r12.e5ea0ff-1
kakmerge-git    18.580c832-1
lscolors-git    r314.6fe816f-1
xkblayout-state-git     20180812-1
```

rafasc:
I would expect that from split packages, but these doesn't seem to be split packages... (Although not sure how the output of `--pkgbase` would be useful is in that case either)

Can you provide your `repo.db`?

maximbaz:
@rafasc http://pkgbuild.com/~maximbaz/repo/ 🙂

rafasc:
@maximbaz 
``` diff
     } /^%VERSION%$/ {
         if (!devel || pkg ~ /-cvs$|-svn$|-git$|-hg$|-bzr$|-darcs$/) {
             getline; printf("%s\t%s\n", pkg, $1)
+            pkg="(null)"
         }
     }'
 }
```
This would fix it, in the same way expac deals with missing pkgbases. But I am seriously considering just dropping the `add --pkgbase` commit...

AladW:
Maybe we could leave out the `--pkgbase` option and introduce a `--table` format instead, similar to `aur depends`:
```
<pkgname>\t<pkgbase>\t<pkgver>
```
or with the repo name as fourth field:
```
<pkgname>\t<pkgbase>\t<pkgver>\t<repo>
```

AladW:
Also could you base this on master again (`srcver-no-parallel` was merged)

rafasc:
As it currently stands, `aur repo` requires you specify the repo via `-d` argument. So I went with your first suggestion.

The other could be done with something like:


`for repo in custom vcs ; do aur repo --table -l -d $repo | sed "s/$/\t$repo/"; done;`

maximbaz:
`aur repo -l` and `aur repo -l --devel` work well, I like the `--table` approach but it still reports some weird results on my repo, where `pkgbase` in some packages is missing (see `cryptboot` and `fonts-*`):

```
❯ aur repo -l --table        
antigen-git     antigen-git     v2.2.3.r6.gc91f77c-1
aurutils-git    aurutils-git    2.3.1.r71.g4a5de452-1
bingo-git       bingo-git       r929.76bcd77-1
brave   brave   0.68.132-1
chromium-vaapi  chromium-vaapi  76.0.3809.132-1
cryptboot               1.1.0-1
dell-command-configure  dell-command-configure  4.2.0-1
duration-git    duration-git    1.0.0.r4.g41f55b0-1
fonts-meta-base         1-2
fonts-meta-extended-lt          3-1
hunspell-da     hunspell-da     2.4-1
jetty   jetty   9.4.19-2
```

rafasc:
`curl -s https://pkgbuild.com/~maximbaz/repo/maximbaz.db --output - | bsdtar -xOf - cryptboot-*/desc`
There's no `%PKGBASE%` field, so it reflects that by leaving it empty.

But in the output, the field should still be tab-delimited even though it is empty. This allows users to pipe the output into `column`  making it more human-friendly.
E.g. `aur repo -l -d maximbaz --table | column -t -s $'\t' -N PKGNAME,PKGBASE,PKGVER`

---
We also rely on `%VERSION%` showing after both `%NAME%` and `%BASE%` which is probably good enough for most cases, since [`repo-add`](https://git.archlinux.org/pacman.git/tree/scripts/repo-add.sh.in#n306) uses (`NAME,BASE,VERSION`) in order, but something I am not sure we should be doing in general.

AladW:
Maybe if `%BASE%` is empty, a hyphen could be added to the output? (again similar to `aur-depends`)

I don't have a solution for the order of `%VERSION%`, but I guess we're not supporting anything but `repo-add` anyway. `aur-graph` also relies on a certain input order for `pkgname`, `pkgbase` and other fields.

rafasc:
> guess we're not supporting anything but `repo-add` anyway.

In that case bugs from @maximbaz repo can be blamed by a non-conforming db file. :smirk: 

Changed the missing case from empty to `"-"` and clarified the dependency on the order of tokens on the commit message. In case we stumble on this again.

AladW:
As a follow-up to what we've discussed on IRC, how about doing something like `aur-move` (or `aur-remove --move`). This would be combined with an extended version of `aur-sync --provides`, i.e. `--provides=custom` or `--provides-any` as follows:

1. Build vcs packages + regular or vcs dependencies
2. Move packages to either `custom` or `custom-vcs`, depending on AURVCS or a fixed pattern
3. List "devel" packages by listing `custom-vcs`

Thoughts?

rafasc:
After we talked, I thought a bit more about it. 

`aur-move` should be small enough in complexity to fit inside `aurvcs(7)` where this idea can be described along with ideas to check if there are updates available.

And since the rules that define what constitutes a vcs package are blurred, I don't mind step #&zwnj;2 being more manual / left to the user to decide.

I think that's a step in the right direction. Feel free to close this, the `--table` part of this can be done separately.

maximbaz:
I missed the discussion on IRC, but does that mean I will now be forced to use two repos if we want aurutils to understand my definition of a "VCS" package?

To be honest I like that so far we are not enforcing any rules on how many repos users must maintain and which packages must go into which repos  :)

AladW:
No - it's entirely up to the user whether a single repo is chosen for all packages, or a separate repo for the VCS packages. The benefit of latter approach is transparency to `aur-repo --list` and `aur repo -u` without additional processing. Both approaches can however be documented in `aurvcs(7)`.

Issue #582, opened 2019-07-27
scripts: alternatives to parallel


AladW:
`parallel` is bulky and has questionable ethics regarding licensing (i.e. ignoring https://www.gnu.org/licenses/gpl-faq.html#RequireCitation as a GPL-licensed project). Look for some alternatives as suggested in #579.

AladW:
Previous discussion: https://github.com/AladW/aurutils/issues/391

Earnestly:
As I once was playing with `make` to handle dependency resolution specifically to find parallel regions, perhaps it can be employed for this?  Though if handling spaces is necessary things could become ugly.

I'm not entirely sure based on the code what the inputs and outputs of the use of `parallel` are supposed to look like, I'll have to try get my head around it.

Earnestly:
On second thought, how about a simple shell implementation such as the one listed on wooledge?

Something like this:

```bash
nprocs=$(("$(nproc)" + 2))

for pkg in "$@"; do
    ((i++ >= nprocs)) && wait -n
    (renice 10 "$BASHPID" && srcver_pkgbuild_path "$pkg" "$tmp"/results "${makepkg_args[@]}") &
done
wait
```

While the make based one does work, the movement of directories is a little awkward and so is the bookkeeping, not even considering dealing with directories that contain spaces.

AladW:
I came to the same conclusion, but used seperate output files to preserve output order:
```
for n in "$@"; do
    { env -C "$n" -i nice -n20 \
	  makepkg "${makepkg_args[@]}" >"$tmp/$n" 2>&1
      echo $? >"$tmp/$n~exit"
    } &

    while (( $(jobs | wc -l) >= 8)); do
	sleep 0.1
	jobs >/dev/null
    done
done
wait

failed=()
for n in "$@"; do
    if (( $(<"$tmp/$n~exit") == 0 )); then
	srcver_pkgbuild_info "$n"
    else
	failed+=("$n")
    fi
done

for f in "${failed[@]}"; do
    warning 'makepkg %s failed for package %s\n' "${makepkg_args[@]}" "$f"

    printf '8<----\n'
    cat >&2 "$tmp/$f"
done
```
The `sleep` approach is probably less effective than `wait -n`...

AladW:
That leaves the two calls here: https://github.com/AladW/aurutils/blob/master/lib/aur-sync#L250-L257

`parallel ln` could be replaced with `ln -t`. Not sure on the parallel aur-fetch.

[...]

PR #583, opened 2019-07-29
aur-srcver: remove parallel

AladW:
See #582. This commit also adds the `-j`/`--jobs` flag for the upper bound of makepkg jobs.

### DIFF
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -2,8 +2,6 @@
 # aur-srcver - update and print package revisions
 readonly argv0=srcver
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
-readonly startdir=$PWD
-makepkg_args=('--nobuild' '--nodeps' '--skipinteg')
 readonly AUR_COLOR=${AUR_COLOR-auto}
 
 srcver_pkgbuild_info() {
@@ -19,16 +17,6 @@ srcver_pkgbuild_info() {
         printf %s\\t%s\\n "${pkgbase:-$pkgname}" "$fullver"
     '
 }
-export -f srcver_pkgbuild_info
-
-srcver_pkgbuild_path() {
-    if env -C "$1" makepkg "${@:3}" ; then
-        srcver_pkgbuild_info "$1" >> "$2"
-    else
-        return
-    fi
-}
-export -f srcver_pkgbuild_path
 
 trap_exit() {
     if ! [[ -o xtrace ]]; then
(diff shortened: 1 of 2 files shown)


lib/aur-srcver
```diff
@@ -48,60 +36,70 @@ if [[ ( -t 1 && -t 2 && ! -o xtrace && $AUR_COLOR == auto ) || $AUR_COLOR == alw
...
+done
 
-            cat "$tmp/${seq}_makepkg" >&2
-            cat "$tmp/${seq}_makepkg.err" >&2
-        else
-            continue
-        fi
-    done < "$tmp"/makepkg_log
+for f in "${failed[@]}"; do
+    warning 'makepkg %s failed for package %s\n' "${makepkg_args[@]}" "$f"
```
rafasc:
```suggestion
    warning 'makepkg %s failed for package %s\n' "${makepkg_args[*]}" "$f"
```
Otherwise this will expand to multiple arguments.

lib/aur-srcver
```diff
@@ -48,60 +36,70 @@ if [[ ( -t 1 && -t 2 && ! -o xtrace && $AUR_COLOR == auto ) || $AUR_COLOR == alw
...
 
-    exit "$jobs_exit"
-else
-    cat "$tmp"/results
-fi
+    printf '8<----\n' >&2
+    cat "$tmp/$f"/log >&2
```
rafasc: Swap the order of these two lines. So the error message becomes upper bounded by the 'warning', and lower bounded by the 'scissors'.

AladW:
I wonder if with a small change (`basename $n`), we could have `aur-srcver` take any paths to source directories, i.e. drop the assumption it is run in some directory containing them :thinking:

AladW:
Other than the above case this seems to work quite well. cc. @maximbaz

lib/aur-srcver
```diff
@@ -96,10 +96,10 @@ for n in "$@"; do
 done
 
 for f in "${failed[@]}"; do
-    warning 'makepkg %s failed for package %s\n' "${makepkg_args[*]}" "$f"
+    warning 'makepkg %s failed for package %s' "${makepkg_args[*]}" "$f"
 
     cat "$tmp/$f"/log >&2
-    printf '8<----\n' >&2
+    plain '8<----\n'  >&2
```
rafasc: Not sure if intentional, but `plain` already takes care of the trailing new line, this will add a empty line between messages.

AladW: It wasn't, good catch.

AladW:
Thanks for the review. For now I believe `parallel` should stay in depends, until we also change `aur-sync` to not use it (I will look at this in a later PR)

PR #584, opened 2019-07-29
aur-repo: fix --sync option

rafasc:
fc4fc06 (aur-repo: remove expac, 2019-03-14) breaks --sync by
incorrectly determining pacman_dbpath.

Fix that.

### DIFF
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -70,8 +70,8 @@ while read -r key _ value; do
         \[*\]*)
             section=${key:1:-1}
             ;;
-        DBPath=)
-            pacman_dbpath=$key
+        DBPath=*)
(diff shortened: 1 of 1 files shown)


AladW:
Nice catch :skull:

lib/aur-repo
```diff
@@ -70,8 +70,8 @@ while read -r key _ value; do
         \[*\]*)
             section=${key:1:-1}
             ;;
-        DBPath=)
-            pacman_dbpath=$key
+        DBPath=*)
+            pacman_dbpath="$value"
```
AladW: the quotes shouldn't be required

rafasc: changed.

PR #587, opened 2019-08-03
Fix aur color

### LINKED ISSUE #585: aur-sync: no color on error messages
For some reason I'm having no colored messages on `aur-sync`:

image

`aur-search` however works fine. Doing an `xtrace`, it seems to fail at the `[[ -t 1 ]]` check, so maybe it's related to my shell configuration. Can anyone reproduce this? `aurutils 2.3.1.r64.ga843275-1`

AladW:
All `aur` scripts currently check if both `stdout` and `stderr` are connected to a terminal when determining if color should be enabled (`AUR_COLOR=auto`). However, only `aur-search` prints colored messages to `stdout`.

Change this so that scripts apart from `aur-search` only check `stderr` for colorization.

Closes #585

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -48,7 +48,7 @@ source /usr/share/makepkg/util/util.sh
 source /usr/share/makepkg/util/message.sh
 source /usr/share/makepkg/util/parseopts.sh
 
-if [[ ( -t 1 && -t 2 && ! -o xtrace && $AUR_COLOR == auto ) || $AUR_COLOR == always ]] ; then
+if [[ ( -t 2 && ! -o xtrace && $AUR_COLOR == auto ) || $AUR_COLOR == always ]] ; then
     colorize
 fi
 
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -41,7 +41,7 @@ usage() {
 source /usr/share/makepkg/util/message.sh
 source /usr/share/makepkg/util/parseopts.sh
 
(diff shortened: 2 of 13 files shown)


rafasc:
I am not confident on making `aur-search` an exception.

Since we now have `$AUR_COLOR` to explicitly enable or disable colorization, we could uniformly change what is the `auto` behaviour.

Either by only checking `stderr` and  users that want paginated `aur-search` results would use `aur search | less -R`, or `AUR_COLOR=never aur search | less` depending if they want colorization or not.

Or we could `export AUR_COLOR` making scripts inherit the colorization intention of their parents.
```
if [[ ( -t 1 && -t 2 && ! -o xtrace && $AUR_COLOR == auto ) || $AUR_COLOR == always ]] ; then
￼	    colorize
            export AUR_COLOR=always
￼else
            export AUR_COLOR=never
fi
```
Thoughts?

AladW:
Let's continue discussing this in #585. I'd go with a different approach than in this PR:
* Removing color from "plumbing" scripts (179bb64ed181e2f6f32eeb5bd30bdda1b7c858b6) should remove the need for exporting variables, in particular in the "porcelain".
* `NO_COLOR` (effectively removing `AUR_COLOR=always`, 65b2b54a109cb8b5adc6a09e7176d5c0040adf03) is supported by some other programs and removes one case to distinguish. I guess for `aur-search` it means you can no longer do `AUR_COLOR=always aur search | less -R` or so.

PR #590, opened 2019-08-04
aur-fetch: change patch extension to '.diff'

AladW:
The command used to generate the diff files is `git diff`(-tree), and having a ".patch" file on new repositories (i.e. compared against `/dev/null`) looks odd.


rafasc:
Seems sensible

AladW:
There's a lot more things to do than change the .diff extension (see `aladbrokeit` branch), so I'll close this one and file a new PR later.

PR #599, opened 2019-09-19
implement NO_COLOR

### LINKED ISSUE #585: aur-sync: no color on error messages

AladW:
This commit replaces `AUR_COLOR` with `NO_COLOR`, effectively removing `AUR_COLOR=always`. `aur-search` is now colorized unconditionally, leaving disabling colors to the user.

For simplicity, colors are removed from the plumbing commands as well.

Fixes #585

----
Taken from https://github.com/rafasc/aurutils/commit/4c8222fe2f487f696f751608954b6ff1906896ac

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -6,7 +6,6 @@ readonly argv0=build
 readonly startdir=$PWD
 readonly XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 readonly PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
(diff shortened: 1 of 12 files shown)


rafasc:
```
for f in lib/*; do 
    echo "$f"
    sed -E 's/(plain|msg|msg2|error|warning|ask) /printf /g' "$f" | shellcheck -x -i2059 -
done
```
Catches a couple more potential cases to be squashed into to that.

AladW:
:+1:

PR #601, opened 2019-09-19
aur-srcver: remove makepkg --log

AladW:
`--log` was added as workaround when `makepkg` output went "missing" while using `parallel`. Since we no longer use parallel (and `--log` apparently stacks the logs, potentially resulting in dozens of log files), remove the option again.

### DIFF
--- a/lib/aur-srcver
+++ b/lib/aur-srcver
@@ -45,7 +45,7 @@ if ! parseopts "$opt_short" "${opt_long[@]}" "${opt_hidden[@]}" -- "$@"; then
 fi
 set -- "${OPTRET[@]}"
 
-makepkg_args=('--nobuild' '--nodeps' '--skipinteg' '--log')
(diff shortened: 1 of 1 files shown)


rafasc:
I use a slightly modified version of `aur gc` from `aur(1)`
```diff 
- find "$AURDEST" -name .git -execdir git clean -xf \;
+ find "$AURDEST" -name .git -execdir git clean "${@:--xn}" \;
```
Without arguments, `aur gc` would only tell you what `-fx` would have done. And since it forwards all arguments to `git clean`,  all the log files could be wiped with `aur gc -fx '*.log.*'`

Maybe a better suggestion is to just have:
`find "$AURDEST" -name .git -execdir git clean "$@" \; ` and let the user decide which arguments they want to run.

AladW:
I'll merge this then and we can discuss the `aur-gc` at any later point. Thanks.

Issue #594, opened 2019-08-31
aur-sync: expected download size exceeded after pacman -Fy


terencode:
Sometimes, running `aur sync -u` will fail to install a package after the `pacman -Fy` step, printing:
> failed retrieving file 'aur.files' from disk : expected download size exceeded

Running `pacman -Syu` makes it work until the next call to `pacman -Fy`

### Known workaround:

Navigate to the aurutils repo cache directory then
```
mv dbname.files.tar dbname.files  
```
where `dbname` is the name of the aurutils repo you created.

AladW:
https://github.com/AladW/aurutils/pull/595

Issue #593, opened 2019-08-12
aur-build: do not remove temp dirs on error


AladW:
I'm facing the following situation (which should apply to all versions of `aur-build` since `0.1.0`...):

1. Build a package with `--rmdeps`
2. Have the removal of packages fail, in this case due to packages installed manually during the (long duration) build.
3. `makepkg` fails with an "unknown error" due to https://lists.archlinux.org/pipermail/pacman-dev/2019-January/023103.html
4. `aur-build` exits before the built package could be moved to the repository
5. on exit, the temporary `/var/tmp` directory containing the built package is removed.

Clearly this annoying as now the build process has to start over for no reason. Some options to make this more robust:

1. Only remove the temporary directories (`$tmp`, `$var_tmp`) if no occurs occured. This could also improve debugging in general, as errors may not be easily reproducible.
2. Try to complete as many remaining steps of the build (`mv`, `repo-add`, ...), and exit before the next build/iteration begins so the user can investigate the error.
3. Have `makepkg` not fail on superficial errors such as removing `makedepends` by trapping `SIGINT`.

----
- [ ] Do not remove temporary files if an error occurred or `AUR_DEBUG` is set
- [ ] Namespace temporary files under `$TMPDIR/aur/$argv0` for easy retrieval

Earnestly:
You want `1.`.  The mantra of failing early and failing hard is one fairly nice feature of resilient software.  The trouble with `2.` is that you're now in an erroneous state so making decisions like that become more complicated.  `3.` is an option if you're willing to essentially create a makepkg wrapper which handles all such cases.

As long as your use of `/tmp` is neatly isolated behind a single directory then manual cleanup for `1.` becomes trivial.

AladW:
Perhaps a fourth option which more directly addresses the issue at hand (losing actual build products), is to `rm -d` the `/var/tmp` directory containing the built packages. Preserving the other temporary files is not that important when most issues requiring them have shown to be reproducible.

AladW:
https://github.com/AladW/aurutils/pull/602/commits/2d41ea16003885c17d54178cf51e5077b0468745

PR #602, opened 2019-09-19
add AUR_DEBUG

AladW:
Setting the option to any value enables `xtrace` for every `aur` script using `bash`. Accordingly, check for `AUR_DEBUG` when enabling color or removing temporary directories.


man1/aur.1
```diff
@@ -113,11 +113,19 @@ applicable.
...
+.BR bash (1)
+scripts) in
+.B aur
+programs.
 
 .TP
 .B NO_COLOR
 Output colorization can be disabled by setting this environment
-variable to any value.
+variable to any non-empty value.
```
rafasc: This change is incorrect. Colored output will be disabled by just having the variable set, regardless of its value.

```suggestion
variable.
```

man1/aur.1
```diff
@@ -113,11 +113,19 @@ applicable.
 
 .SH ENVIRONMENT
 .TP
+.B AUR_DEBUG
+Enable debug mode (typically
```
rafasc: Similarly to below, 
```suggestion
Setting this variable enables debug mode (typically
```

AladW:
An alternative approach discussed on IRC:
* Set `BASH_XTRACEFD` to a file descriptor pointing to a log file;
* If an error occured, present this file to the user.

The only drawback I can think of is size of the log file, and this catches logs for "intermittent" issues where a next invocation would run successfully (e.g. #594). `xtrace` is not exported to makepkg, so the size should remain low for the foreseeable future. 

A more "measured" approach is to trap ERR and print `'+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'` for the line where an error occured; a complete log would be still accessible through `AUR_DEBUG`.

AladW:
Scrapped the above approach for `rm -fd` in `aur-build`, see https://github.com/AladW/aurutils/issues/593#issuecomment-536322066.

PR #604, opened 2019-09-21
Update xargs example in aur.1

Strykar:
The example `xargs -a installed.txt aur sync --repo=custom` should be `xargs -a installed.txt aur sync -d custom`.


AladW:
Thank you. Note that you can keep existing PRs by just force pushing to your branch.

PR #605, opened 2019-09-26
shellcheck: minimize exclusions

AladW:
Only ignore unsuported languages (SC1071)

Hide false positives on a case by case basis


AladW:
:+1:

PR #610, opened 2019-09-29
aur-sync: add --provides-any

AladW:
This allows to take packages from other (local) repositories as dependencies. See also https://github.com/AladW/aurutils/pull/581#issuecomment-532862450.

### DIFF
--- a/man1/aur-sync.1
+++ b/man1/aur-sync.1
@@ -138,11 +138,17 @@ upgrades are taken as target regardless of this setting.
 
 .RS
 .B Note:
-This option does not take other repositories than the local repository 
-into account. Furthermore, the version output from
+The version output from
 .BR aur\-vercmp (1)
 is unaffected by this option.
 .RE
+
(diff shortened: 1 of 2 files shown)


man1/aur-sync.1
```diff
@@ -138,12 +138,18 @@ upgrades are taken as target regardless of this setting.
...
 .B Note:
-This option does not take other repositories than the local repository 
-into account. Furthermore, the version output from
+The version output from
 .BR aur\-vercmp (1)
 is unaffected by this option.
 .RE
 
+.TP
+.BR \-\-provides\-any
```
rafasc: Nit: Only .B is needed here. The rest look ok.
```suggestion
.B \-\-provides\-any
```

AladW:
As discussed on IRC, I'll make a new PR which adds `--provides` ("provides-any") and `--provides-from` ("provides=repo1,repo2").

PR #617, opened 2019-10-16
Document more workflows for vcs packages

rafasc:
Initially this PR removed aur-vercmp-devel and introduced documentation changes that were biased towards a repo-based workflow for vcs packages.

Since these are actually two actions, the removal was split into PR #622.


man7/aurvcs.7
```diff
@@ -0,0 +1,162 @@
...
+.SH NAME
+aurvcs \- manage development packages
+
+.SH DESCRIPTION
+There's two main ideas on how to deal with vcs pacakges. One consists in
+building vcs-pacakges into a dedicated repository, while the other is
+based on pattern-matching common vcs pkgnames.
+
+Using a dedicated repo has some advanges especially when it comes to
+decide what is and what isn't considered a vcs pacakge.
```
maximbaz: Well I wouldn't call this an advantage because as you describe below all dependencies are incorrectly considered to be vcs packages too.

While pattern-matching solution might have a few false positives, the separate repo is practically guaranteed to be filled with dependencies that are not vcs packages.

The current wording makes me think that we almost recommend everyone to try using the "separate repo" approach first while the "prefix" based one has far fewer false positives, and I propose we change the wording and always start with "prefix" approach first and then explain what needs to be different if one chose to use the "repo" approach.

Remember that AUR itself uses a "prefix-based" definition of what [a vcs package is](https://lists.archlinux.org/pipermail/aur-dev/2019-May/004649.html), it's not an essential part of AUR but still a precedent 🙂

AladW: I would argue that -git packages should properly depend on -git packages. Do you have some examples where they don't, and the motivation for doing so? (I only remember mesa-git not depending on llvm-git.)

maximbaz: mmm help me understand, `aurutils-git` depends on `('git' 'jq' 'pacutils' 'parallel' 'wget')` and not on `('git-git' 'jq-git' 'pacutils-git' 'parallel-git' 'wget-git')`... Or if dependencies are in official repos then this doesn't apply? 🙂 What if `parallel` and `parallel-git` were both on AUR, would `aurutils-git` then depend on `parallel-git`?

rafasc: > Well I wouldn't call this an advantage because as you describe below all dependencies are incorrectly considered to be vcs packages too.

This part is still missing, but the intention would be to promote a workflow that rebuilds packages in the vcs repo leveraging `--provides-any`, or (not implemented yet) `--provides=<repo>`. Keeping an explicit list of packages that should be treated as vcs on the vcs repo, regardless of naming conventions or other factors.

> Remember that AUR itself uses a "prefix-based" definition of what [a vcs package is](https://lists.archlinux.org/pipermail/aur-dev/2019-May/004649.html), it's not an essential part of AUR but still a precedent

*In my opinion* the vcs repo approach is better because the action of building towards the vcs repo is explicit. If you sync something that depends on a vcs package, that dependency is only treated as a vcs if you manually move to the vcs repo. I think this explicitness is needed since the AUR doesn't have a way to *define* without doubt what vcs packages are.

Even if we consider that a precedent, it's a very weak one because if you read the rest of the thread you'll see that the patch was later toned down from "This is  a VCS package" to "This seems to be a VCS package. ". It makes sense in the context that change was made as flagging a OOD is something done by humans (supposedly) read big red warnings, but doesn't translate very well for our context.

On a final note, @AladW was hesitant to add `vercmp-devel` as an official script back then, I pushed it in favor of it, but the reason behind this discussion is the very same arguments that can be made for removing official support for it. Without an official definition, it becomes subject to preference what is considered better/correct.

man7/aurvcs.7
```diff
@@ -0,0 +1,162 @@
+.TH AURVCS 7 2019-10-13 AURUTILS
+
+.SH NAME
+aurvcs \- manage development packages
+
+.SH DESCRIPTION
+There's two main ideas on how to deal with vcs pacakges. One consists in
```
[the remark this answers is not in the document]
rafasc: Thanks

man7/aurvcs.7
bartoszek: This is somewhat week argument, most VCS packages has it's dependencies pinned to some stable version during development cycle, in most cases building them against VCS versions of dependencies is not supported upstream, and is going to fail badly. 

From my point of view, sane PKGBUILD should maintain versioned `{make}dependes` array (example: gimp-git) and  the end user should be responding for pulling and building packages with corresponding versions.

man7/aurvcs.7
```diff
@@ -0,0 +1,162 @@
...
+    $ aur sync \-d vcs aurutils-git .\|.\|.
+.EE
+.
+.IP \(bu
+Leverage
+.BR aur\-repo (1)
+to list them all rebuild them.
+E.g.:
+.EX
+    $ aur repo \-\-list \-d vcs | cut \-f1 | aur sync \-d vcs \-\-rebuild
```
jrudess: This example is missing xargs before 'aur sync'.

AladW:
See #961 for a new idea on this topic. If we store the "seen" upstream `git` commits somewhere, they can be queried them via APIs (`github`, `gitlab`) instead of running `aur-srcver`. This require some effort though and is specific for a small subset of VCS packages...

The "seen" commit would be more informative if it is linked to the built package. Apart from storing `pkgname-pkgver <-> upstream commit` manually, this could be part of the package itself: see https://lists.archlinux.org/pipermail/pacman-dev/2022-February/025511.html for a proposal on this regard.

PR #620, opened 2019-10-17
Use AUR_SEEN instead of AURUTILS_SEEN

Morganamilo:
AUR_SEEN is becoming the standard ref name, yay and pacaur also support
this.

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -106,7 +106,7 @@ fi | while read -r pkg; do
     export GIT_DIR GIT_WORK_TREE=$pkg
 
     if (( confirm_seen )); then
-        git update-ref AURUTILS_SEEN HEAD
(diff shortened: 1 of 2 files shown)


rafasc:
While Alad supports this change, https://github.com/Jguer/yay/pull/1062#issuecomment-541668054 I think we should hold this back for now.

Mainly because it requires manual intervention on cached repos to migrate the reference. (Would be nice to add instructions on how to do that in the commit message, i.e. mention `git update-ref AUR_SEEN AURUTILS_SEEN`.
There's also the `aur-fetch` changes on some internal branches (aladbrokeit/webrokeit) that will require refreshing the cache. So it would be nice to bundle those two in the same branch in order to reduce "manual intervention events" on master.

AladW:
Thanks for the update 🥳

PR #623, opened 2019-10-22
Add devtools dependency in PKGBUILD

adrien1018:
`devtools` is needed to build the package.


rafasc:
While building in chroot is recommended, `aurutils` doesn't force its users to do it. `aurutils` should be functional without it as long as you don't use `-c/--chroot`.

`devtools` is already included as an optional dependency. If you want to use chroot build, you will have to install it yourself.

https://github.com/AladW/aurutils/blob/ffc08900adee4f8771f2bb5be5093a5663047895/makepkg/PKGBUILD#L15-L16

maximbaz:
Let's close the PR then 👍

PR #625, opened 2019-10-23
README: add rebuild-detector

### LINKED ISSUE #612: scripts: check if packages require a rebuild
I'm using this `oneliner` to update AUR packages that deps get updated in next sys upgrade.
```sh
aur repo --list|cut -f1|xargs expac '%n\t%E' |eval grep "$(printf -- " -e \" %s \"" $(pacman -Supd --print-format '%n'))"|cut -f1
```
Think it would be beneficial to add this to `aur sync -u` (perhaps as extra flag) to allow user to preserve system coherency across sys upgrade.

AladW:
As building AUR packages is the user's responsibility, so is rebuilding them for example on `soname` changes in official repository packages. Add a hint in the `README` for a project which helps with this task.
    
Fixes #612

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ See [ISSUE_TEMPLATE.md](ISSUE_TEMPLATE.md). For informal discussion, see the
 
 The following (third-party) projects may be used together with `aurutils`:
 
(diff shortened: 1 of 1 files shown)


AladW:
Thanks for the approval :smirk:

PR #631, opened 2019-10-26
aur-sync: add --provides-from, change --provides

AladW:
Change `--provides` to take any `pacman` sync repository into account, as this case is more common than the "self-sufficient" case where only virtual dependencies in the local repository are considered.

To preserve and extend the latter approach, add a new option `--provides-from`. This option can be seen as a special version of `--provides`, and is checked for first in the `aur-repo-filter` block.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -99,7 +99,7 @@ opt_long=('bind:' 'bind-rw:' 'database:' 'directory:' 'ignore:'
           'force' 'ignore-arch' 'log' 'no-confirm' 'no-ver' 'no-graph'
           'no-ver-shallow' 'no-view' 'print' 'provides' 'rm-deps'
           'sign' 'temp' 'upgrades' 'pkgver' 'rebuild' 'rebuild-tree'
-          'build-command:' 'ignore-file:' 'remove')
+          'build-command:' 'ignore-file:' 'remove' 'provides-from:')
 opt_hidden=('dump-options' 'allan' 'ignorearch' 'ignorefile'
             'noconfirm' 'nover' 'nograph' 'nover-shallow' 'noview'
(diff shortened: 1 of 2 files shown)


lib/aur-sync
```diff
@@ -144,6 +144,9 @@ while true; do
             makepkg_args+=(--log) ;;
         -P|--provides)
             provides=1 ;;
+        --provides-from)
+            shift; IFS=, read -a repo -r <<< "$1"
+            repo_p+=("${repo[@]}") ;;
```
rafasc: Isn't `read -a repo_p -r <<< "$1"` enough here?

man1/aur-sync.1
```diff
@@ -113,17 +113,20 @@ Do not present build files for inspection.
...
+.TP
+.BI \-\-provides\-from= DIR1,...
+As
+.BR \-\-provides ,
+but only the specified (comma\-separated) repositories are taken into
+account. If the same package is provided in multiple repositories,
+ordering is ignored (for example, \-\-provides\-from=a,b is equivalent
+to \-\-provides\-from=b,a) and dependencies are installed according to
+the order defined in
+.BR pacman.conf (5).
```
rafasc: :+1:

AladW:
:tada:

lib/aur-sync
AladW: I guess it's too late now, but I copied it from `--ignore`, where the idea was to combine all the arguments into one. So `--provides-from=a,b --provides-form=c,d` would be the same as `--provides=a,b,c,d`. With `read -a`, it'd be whatever `--provides-from` is specified last.

rafasc: Makes sense. Wasn't thinking in the multiple `--provides-from` case. :+1:

PR #635, opened 2019-11-02
aur-build: remove default arguments

AladW:
Traditionally, `aur-build` has a set of default (`makepkg`, `makechrootpkg`) options that can be overridden by a new set of options, specified after `--`. This means adding a simple option like `-r` is done as follows:
```
$ aur build -- -r -sc
```
where `-sc` are the default `aur-build` options (here, we've replaced `makepkg -sc` with `makepkg -src` in the `aur-build` invocation). A possible option is to append to the default options, instead of overwrite them:
```
$ aur build -- -r
```
In this case `aur-build` contains a set of `hardcoded` options which may not be suitable to all users.

This commit removes any notion of default arguments for `makepkg` or other wrapped build commands. For convenience, common options are parsed directly from the command-line, including `makechrootpkg` options included in `aur-sync`. The above example would then be written as:
```
  $ aur build -rsc
```
which runs `makepkg -src`. The base command:

  $ aur build

would then run `makepkg` without arguments. As the `-s` /  `--syncdeps` option conflicts with `--sign`, rename the short option for `--sign` to `-S`. Coincidentally, this is the same short option used in `git-commit(1)` for GPG signing.

`aur-sync` operation is mostly unaffected, as it specifies --clean --syncdeps as default arguments to aur-build. Otherwise, aur-build is now fully decoupled from aur-sync.

**Note:** A workaround to keep `aur-sync -Ac` functional with the above changes is added through `build_extra_args`. In particular, `aur-sync -Ac` now translates to:
```
$ aur sync -c -- -- -A
```

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -72,9 +72,9 @@ opt_short='a:B:d:D:AcfnrsvLNRST'
 opt_long=('arg-file:' 'chroot' 'database:' 'repo:' 'force' 'root:' 'sign'
           'verify' 'directory:' 'no-sync' 'pacman-conf:' 'results:'
           'makepkg-conf:' 'remove' 'build-command:' 'pkgver' 'prefix:'
-          'rmdeps' 'noconfirm' 'ignorearch' 'log' 'recreate' 'new' 'bind:'
+          'rmdeps' 'no-confirm' 'ignore-arch' 'log' 'recreate' 'new' 'bind:'
           'bind-rw:' 'prevent-downgrade' 'temp' 'syncdeps' 'clean')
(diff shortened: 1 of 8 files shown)


AladW:
@maximbaz @rafasc Any further ideas / thoughts for improvement? I don't think it's the end of all option parsing (in particular for `makepkg` options in `makechrootpkg`), but to me it's an improvement over what we have now.

maximbaz:
I haven't tested yet but it all looks reasonable and I'm happy to see this merged 👍  If I spot anything afterwards I'll let you know 😄

PR #634, opened 2019-11-01
aur fetch: learn to parallelize

rafasc:
Rewrite of `aur-fetch` to implement features such as
- ~ability to choose between `git reset (default)`/`pull --rebase` to update cached repos. (discard local commits vs trying to reapply local commits on top of the incoming changes)~ #639
- ~ability to choose between a `git log` or `git diff` to generate review patches.~ #640
- fetch packages in parallel

Issue #497, opened 2019-01-09
Regression tests


AladW:
I've identified some issues where regression tests would be useful. This issue keeps track of those.

- [x] aur-build: build directory empty after makepkg => bash expansion fails · Issue #513
- [x] aursearch: fails silently on too many results (search type) · Issue #257
- [x] aurbuild: corrupted packages with --force · Issue #85
- [x] Aursync "cannot create directory /aursync" · Issue #216
- [x] aursearch: exit 0 when no results found (json_long/json_short) · Issue #187
- [x] Fix aurbuild fail on creating a new container by vorph1 · Pull Request #167
- [x] aurbuild: local repository is not available in the arch-nspawn container for pacman -Syu · Issue #461
- [x] AUR_REPO doesn't actually work · Issue #378
- [x] aur-sync: build order is inverted when installing poezio · Issue #414
- [x] aur-search hangs indefinitely. · Issue #403

AladW:
All tests are pushed to the `tests` branch.

Issue #356, opened 2018-04-16
tests: TAP support


AladW:
https://github.com/andrewgregory/tap.sh
http://testanything.org/

To avoid recurring issues like #355, #329, #321, and so forth.

jcf:
As per [my comment](https://github.com/AladW/aurutils/issues/355#issuecomment-381571470), there may be value in static analysis via [ShellCheck](https://www.shellcheck.net/). It's certainly less work than writing a full test suite.

AladW:
I already did most of the works in `aurutils-test`, unfortunately I seem to have removed that repository in an overzealous cleanup... :disappointed:

AladW:
This will be covered by https://github.com/polygamma/aurts.

bb010g:
That repository doesn't exist anymore?

rafasc:
on the TAP side, have you considered using [chriscool/sharness](/chriscool/sharness)?
It's basically the `git` suite extracted from `git`. It's being actively maintained by a regular `git` contributor. (@chriscool). 

Not sure if master is stable enough for use or if people stick to the v1.0.0 release, but definitely worth looking.

AladW:
The old tests were still in git: https://github.com/AladW/aurutils/tree/74e230ac032fe5f6da0005a0d97222aabb24a887/test

rafasc:
@tomeon Answering your https://github.com/AladW/aurutils/pull/636#issuecomment-557319410 here to be more in-topic.

> @rafasc -- sounds like a bigger discussion, and out-of-scope for this issue, but: what about [BATS](https://github.com/bats-core/bats-core/)?
> 
> Seems to tick some of the boxes you listed:
> 
> * [TAP support](https://github.com/bats-core/bats-core/blob/87b16eb8ec381beca7f05d028e4ad7cf3f647a0f/libexec/bats-core/bats-format-tap-stream)
> * [Support for function libraries](https://github.com/bats-core/bats-core#load-share-common-code)
> 
> Plus there's an available [Docker image](https://github.com/bats-core/bats-core#running-bats-in-docker) for use with (say) Travis CI, and an [AUR package](https://aur.archlinux.org/packages/bats-core/) that could be used with (e.g.) [builds.sr.ht](https://man.sr.ht/builds.sr.ht/compatibility.md#arch-linux).
> 
> Food for thought, perhaps :)

[...]

PR #636, opened 2019-11-04
aur-fetch: unset GIT_DIR and GIT_WORK_TREE

tomeon:
Using `master` at 1b4782e2e10acb807c87d81b03f1d9b583bb9d4, `aur-fetch` exits with nonzero status under the following conditions:

1. It needs to fetch more than one repository
2. At least one repository (other than the first one listed on the command line) is being fetched for the first time (initial `git clone`):

```
$ cd "$(mktemp -d)"
$ aur fetch aurutils aurutils-git 
Cloning into 'aurutils'...                                         
remote: Enumerating objects: 348, done.                            
remote: Counting objects: 100% (348/348), done.                    
remote: Compressing objects: 100% (232/232), done.                 
remote: Total 348 (delta 117), reused 347 (delta 116)              
Receiving objects: 100% (348/348), 66.21 KiB | 622.00 KiB/s, done. 
Resolving deltas: 100% (117/117), done.                            
HEAD is now at 3872019 upgpkg: aurutils 2.3.2-1                    
From https://aur.archlinux.org/aurutils                            
 = [up to date]      master     -> origin/master                   
Already up to date.                                                
Current branch master is up to date.                               
fatal: working tree 'aurutils' already exists.                     
==> ERROR: fetch: aurutils-git: Failed to clone repository 
```

The issue appears to be that `aur-fetch` doesn't unset `GIT_DIR` or `GIT_WORK_TREE` at the beginning of the "main" loop, thus reusing the values of `GIT_DIR` and `GIT_WORK_TREE` set on the *prior* iteration.

This merge request adds `unset GIT_DIR GIT_WORK_TREE` to the beginning of the main loop in `aur-fetch`, as well as a regression test case.

TIA for your consideration :)

### DIFF
--- a/test/issue-636
+++ b/test/issue-636
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+argv0=issue-636
+tmp="$(mktemp -d --tmpdir "aurutils-$argv0.XXXXXXXX")"
+
+trap_exit() {
+    rm -rf "$tmp"
+}
+
+trap 'trap_exit' EXIT
+
+cd "$tmp" || exit
+
+aur fetch aurutils aurutils-git


AladW:
This should be fixed along the way in https://github.com/AladW/aurutils/pull/634/commits/4dbab5224277659b0e5a0e5155f3d94deae13c21

Anyway, a test case, nice. :+1:

test/issue-636
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+AURDEST="$(mktemp -d --tmpdir "aurutils-${BASH_SOURCE[0]##*/}.XXXXXXXX")"
+export AURDEST
+
+cleanup() {
+    rm -rf --one-file-system --preserve-root "${AURDEST?}"
```
AladW: A question since "cleanup" is used in other places with temporary files: why those extra flags to `rm`, if the target string is quoted?

test/issue-636
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+AURDEST="$(mktemp -d --tmpdir "aurutils-${BASH_SOURCE[0]##*/}.XXXXXXXX")"
```
AladW: I think it suffices to set `argv0=issue-636`, `aurutils-$argv0` as other `aur` scripts

AladW:
Should be fixed in master, but I'd still be interested in merging the regression test!

test/issue-636
tomeon: Two things: an abundance of caution and (marginally) better diagnostics (well, three, including force of habit).  Certainly not wedded to this approach, though, so I've harmonized this trap function with others in the existing codebase.

test/issue-636
tomeon: Done.

tomeon:
> Should be fixed in master

Indeed!  I have rebased the branch on `master`, dropping my change to `aur-fetch`, and the regression test exits without error.  So, barring false negatives... :)


> but I'd still be interested in merging the regression test!

I've updated the PR in (what I hope is) accordance with your comments; LMKWYT.

AladW:
Thanks for the update. Just a small nitpick: since `aur-fetch` does not actually use AURDEST, I'd remove it (and the `export`) and simply set `$tmp` like other scripts do.

tomeon:
> Thanks for the update. Just a small nitpick: since `aur-fetch` does not actually use AURDEST, I'd remove it (and the `export`) and simply set `$tmp` like other scripts do.

Ah, yes -- suspect that was an artifact of an earlier test implementation that used `aur-sync` rather than calling `aur-fetch` directly.  `s/AURDEST/tmp/g` done.

tomeon:
> I seem to have missed this in my previous review.
> The PR is not based on current master. Would be nice if you did that along with the suggestions from @AladW

Just pushed a rebase on `master`; are you also looking for the test script to exercise some newly-added feature(s)?

AladW:
It looks good to me. If we wanted to test more features, I'd say this belongs to other tests.

rafasc:
@tomeon tests for new features would be nice, but we need to address some aspects regarding tests in general. I.e. Implement proper infrastructure for testing to avoid issues like #645.
We have #356 suggesting TAP, but at some point it was also discussed using a simpler shell/makefile based suite.

Regardless of the approach chosen, we also need some sort of library with functions for common tasks, e.g. setting up repos, configs, etc. Making tests easier to write and run will, hopefully, help turning the #497 list into actual tests. 

Speaking for myself here, but I rarely run the ones we have. Proper infrastructure, and perhaps automation (hello github actions), would help making sure we actually run the tests. Otherwise it becomes hard to justify the effort that needs to be put in.

tomeon:
@rafasc -- sounds like a bigger discussion, and out-of-scope for this issue, but: what about [BATS](https://github.com/bats-core/bats-core/)?

Seems to tick some of the boxes you listed:

- [TAP support](https://github.com/bats-core/bats-core/blob/87b16eb8ec381beca7f05d028e4ad7cf3f647a0f/libexec/bats-core/bats-format-tap-stream)
- [Support for function libraries](https://github.com/bats-core/bats-core#load-share-common-code)

Plus there's an available [Docker image](https://github.com/bats-core/bats-core#running-bats-in-docker) for use with (say) Travis CI, and an [AUR package](https://aur.archlinux.org/packages/bats-core/) that could be used with (e.g.) [builds.sr.ht](https://man.sr.ht/builds.sr.ht/compatibility.md#arch-linux).

Food for thought, perhaps :)

PR #638, opened 2019-11-08
Remove m4 use from build

E5ten:
replace simple m4 generating aur from aur.in with sed
replace completion generation m4 with a bash script

The command_opts script output is identical to the output from the m4 script except without unneeded trailing whitespace.


rafasc:
If we are using bash for this, instead of parsing the .in files, we can make them scripts. And either call/source command_opts.
Here's what I mean: https://github.com/E5ten/aurutils/compare/remove-m4...rafasc:pr/638

E5ten:
Looks good, will you open your own PR, or should I merge those changes?

rafasc:
That was more to demonstrate what I meant. I didn't even bother to make the changes for the zsh completion. You can absorb them and make them your own.

rafasc:
Wrt using `printf`, I would say no. I think the `here-docs` improve readability.
For instance, with `vim`, you can simply comment the `cat<<\EOF` lines and have syntax highlight work out of the box.

Readability was also why I chose to use two `cat<<\EOF` instead of a single `cat<<EOF`. Otherwise we would need to escape anything that bash would substitute.

AladW:
Nice work 👯

Issue #642, opened 2019-11-15
Repo is required in /etc/pacman.conf despite a dedicated pacman.conf is given to aur build


mipimipi:
I observed a strange phenomenon: I use `aur build` to build a package in a chroot container and add it to a repository. Via the `--pacman-conf` option, I provide the command with a dedicated `pacman.conf` file. Despite `aur build` is provided with this file, it requires the repository to be configured in `/etc/pacman.conf` in addition. Surprisingly, it is sufficient to only have a row `[myrepo]` in `/etc/pacman.conf` - without any other information (no `Server = ...` row). 

My scenario: I am using `aur build` to add packages to a remote repository. My script rsync's the remote repo to a local directory, `aur build` is called to add a package to local copy of the remote repo. Then the local copy is rsync'ed back to the remote repo and the local copy is deleted. By using a dedicated chroot container and a dedicated `pacman.conf` file, I would like to avoid that my system is affected by that package build/add process (i.e. the package shall not have to be installed to my local system and the repo shall not have to be configured in my `/etc/pacman.conf`).

The `aur build` statement:
```
aur build --database=<MYREPO> \
                  --arg-file=<LIST-OF-DIRS-WITH-PKGBUILDS> \
                  --root=<PATH-TO-LOCAL-REPO-DIRECTORY>" \ 
                  --chroot \
                  --directory=<PATH-TO-CHROOTDIR> \
                  --pacman-conf=<PATH-TO-DEDICATED-PACMAN.CONF> \
                  -- -c -u -T -- -c -i
```
The dedicated pacman.conf file contains a configuration for the local copy of the repo:

```
[myrepo]
SigLevel = Optional TrustAll
Server = file://<PATH-TO-LOCAL-REPO-DIRECTORY>
```
Why does the repo need to be configured in `/etc/pacman.conf`?

Thanks, mipi

mipimipi:
I am pretty sure that I found the root cause of that behavior: In `aur chroot`, `pacman-conf` is called several times to query some config data. Consequently, it should be run against the custom pacman.conf (if the corresponding cli option has been set), but sometimes it is run against `/etc/pacman.conf` even if there is a custom pacman.conf. On my local machine, I changed the code of `aur chroot`, `pacman-conf` is now running against the custom pacman.conf (if it's existing). With this change my scenario works. I will submit a PR.

mipimipi:
Created PR #647 based on the discussion in PR #643

PR #643, opened 2019-11-16
aur-chroot: use custom pacman.conf

mipimipi:
This resolves issue #642. With this PR, pacman-conf runs against the custom pacman.conf file if such a file was specified in the arguments of aur-chroot. 

Signed-off-by: Michael Picht <mipi@fsfe.org>

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -23,15 +23,25 @@ ignore_defaults() {
     } { print }'
 }
 
-host_with_cache() {
-    printf '[%s]\n' options
-    pacman-conf -c "$pacman_conf" --verbose CacheDir
-
-    for a in "$@"; do
-        printf '[%s]\n' "$a"
-        pacman-conf -c "$pacman_conf" --repo "$a"
-    done
-}
+host_with_cache() {    
+    printf '[%s]\n' options    
+     
+    pacman-conf -c "$pacman_conf" --verbose CacheDir    
+     
+    for a in "$@"; do    
+        printf '[%s]\n' "$a"    
+        set +e    
+        pacman-conf -c "$pacman_conf" --repo "$a" > /dev/null 2>&1    
+        if [[ $? = 0 ]]    
+        then    
+            set -e                  
+            pacman-conf -c "$pacman_conf" --repo "$a"    
+        else    
+            set -e                  
+            pacman-conf --repo "$a"    
+        fi        
+    done    
+}    
 
 trap_exit() {
     if ! [[ -o xtrace ]]; then
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -23,25 +23,20 @@ ignore_defaults() {
     } { print }'
 }
 
-host_with_cache() {    
-    printf '[%s]\n' options    
-     
-    pacman-conf -c "$pacman_conf" --verbose CacheDir    
-     
-    for a in "$@"; do    
-        printf '[%s]\n' "$a"    
-        set +e    
-        pacman-conf -c "$pacman_conf" --repo "$a" > /dev/null 2>&1    
-        if [[ $? = 0 ]]    
-        then    
-            set -e                  
-            pacman-conf -c "$pacman_conf" --repo "$a"    
-        else    
-            set -e                  
-            pacman-conf --repo "$a"    
-        fi        
-    done    
-}    
+host_with_cache() {
+    printf '[%s]\n' options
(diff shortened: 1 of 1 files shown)


AladW:
The issue with this PR is that it sets the chroot pacman conf 
unconditionally to whatever is set with -C (or the default devtools 
configuration). In particular, it would no longer suffice to edit 
/etc/pacman.conf to merge local repository configuration from the host, 
to the chroot.

The ad/chroot branch includes a separate --config flag that allows you 
to set the file pacman-conf is run against. In response to #642 I've 
also added a --ignore-host flag that skips this logic entirely.

Excerpts from mipimipi's message of November 16, 2019 1:42 pm:

mipimipi:
This sounds quite complex.
I installed the AUR package aurutils (i.e. I am using release 2.3.3). The docu (`man aur chroot`) says:
```
[...]
-C FILE, --pacman-conf=FILE
The pacman.conf(5) file used inside the container. Defaults to devtools' pacman-extra.conf.  This file is processed with pacconf(1) to include necessary information from the host, such as mirrors. [...]
```
If I understand the coding of aur chroot (release 2.3.3) correctly, then the logic that is implemented is as follows:

1. If a pacman.conf file has been specified via the `--pacman-conf` option, then proceed with this file.
2. Otherwise proceed with `/usr/share/devtools/pacman-multilib.conf`(!) -> that deviates from the docu
3. In any case: Read the CacheDir and the info about the repositories that have been specified by the `-d` option (in particular the Server directive is important) from `/etc/pacman.conf`

Firstly, this is inconsistent to the documentation. Secondly, this behavior makes only sense if no pacman.conf has been specified via `--pacman-conf` since the default (the devtools pacman.conf files) does neither contain a CacheDir nor information about repositories.

IMO, the implementation should be as follows:
1. If a pacman.conf file has been specified via the `--pacman-conf` option, then proceed with this file.
2. Otherwise proceed with `/usr/share/devtools/pacman-extra.conf` (as written in the docu)
3. Regardless of the pacman.conf file that has been selected to proceed with in 1. and 2.: Read the CacheDir and the info about the repositories that have been specified by the -d option from that pacman.conf file.
4. If that file does not contain the required information, then (and only then) read this information from `/etc/pacman.conf`.

So, yes - you are right, my PR was not OK since some conditional logic was missing. The logic that I propose above is - at least IMO - easier to understand than having another 2 options in addition to `--pacman-conf`. In any case the documentation needs to be adjusted.

AladW:
That looks like a lot of steps. After some more work, I've come up with f04e5f2eee37ecd19318e28560c171bfd1458d4e and 24c4e3e344a83b73840b3fb7c526c2f27cb00f20. That way it's entirely up to the user what pacman configuration the container uses, and `aur-build` would default to a path like `/usr/share/devtools/pacman-aur.conf` for convenience.

----
The old 2.3 behavior could then be restored with a snippet like:
```
#!/bin/bash
devtools_conf=/usr/share/devtools/pacman-extra.conf
pacman_conf=$(mktemp)

print_section() {
    printf '[%s]\n' "$1"
    pacconf --repo="$1"
}

aur repo --repo-list | while read -r repo; do
    print_section "$repo"
done | cat "$devtools_conf" - > "$pacman_conf"

aur build --chroot --pacman-conf "$pacman_conf" "$@"
```
Or, perhaps through introduction of a new `--verbose` option for `aur-repo`:
```
#!/bin/bash
devtools_conf=/usr/share/devtools/pacman-extra.conf
pacman_conf=$(mktemp)

aur repo --repo-list --verbose | cat "$devtools_conf"  - > "$pacman_conf"
aur build --chroot --pacman-conf "$pacman_conf"  "$@"
```

mipimipi:
Sounds good - and I understand that you do not want to make incompatible changes in 2.3. But as I pointed out, the current 2.3 implementation is inconsistent with the 2.3 docu. The implementation that I suggested would (a) make the 2.3 implementation consistent with the docu and (b) does not change the cli. Would you accept a corresponding PR for 2.3? 2.4 would then change the behavior as you proposed.

AladW:
I'd say the release of 2.4 is right around the corner so I'm not sure 
it's worth it. But if you want to file a PR for 2.3, go ahead - just no 
promises it will be reviewed.

Cheers

Excerpts from mipimipi's message of November 23, 2019 12:54 pm:

mipimipi:
Create PR #647

PR #648, opened 2019-11-26
adjusted data retrieval from pacman.conf

mipimipi:
1. First try to retrieve data from custom pacman.conf, then from default.
1. Default is pacman-extra.conf (instead of pacman-multilib.conf

Signed-off-by: Michael Picht <mipi@fsfe.org>

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -29,7 +29,13 @@ host_with_cache() {
 
     for a in "$@"; do
         printf '[%s]\n' "$a"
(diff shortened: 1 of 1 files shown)


mipimipi:
Any feedback so far? Does this PR make sense?

AladW:
Running `pacman-conf` twice, a second time if the first failed, doesn't really make sense to me. Since 2.3 isn't prone to change, I would suggest to apply this patch locally if you need it.

Issue #577, opened 2019-07-06
Prepare 3.0 release


AladW:
This is another big one so we should have some plan + prerelease cycle. I propose the following:

### General
- [x] Remove `aur-jobs` (#579)
  * The script is only used in `aur-sync`/`aur-srcver` and has no non-trivial functionality otherwise.
- [x] Rename `aur-rpc` to `aur-query` (#580)
  * "rpc" describes the protocol, while "query" describes the intent.
- [x] Replace `aur-vercmp-devel` with `aurvcs(7)` (#511)
  * See https://github.com/AladW/aurutils/issues/511#issuecomment-508938979
- [x] Replace `AUR_COLOR` with `NO_COLOR` (#599)

### Release
- [x] Update the changelog
  * @rafasc started work on this, see https://github.com/rafasc/aurutils/commit/6bb045d25bf04fb0612ddb5a487f61b5512d0975
- [x] `git shortlog --no-merges -s '2.3.1..3.0.0' | sort -r | awk '{ if ($1 > 2) { print $0; }; }'`

### Documentation/tests
- [x] `aur-build.1`
- [x] `aur-repo.1`
- [x] `aur-sync.1`
- [x] `aur-chroot.1`
- [x] `aur-fetch.1`
- [x] `aur-query.1` (description text still missing)
- [x] Regression tests (#497)

### `aur`
- [x] Add `AUR_EXEC_PATH` (#672)

### `aur-chroot`
- [x] Use static `pacman.conf` for chroot (#655)
  * Remove`-d`

### `aur-build`
- [x] Append options after `--` to the chroot makepkg (`-c`, #570)
- [x] Expose common makepkg options (compare `aur-sync`)
- [x] Only remove temporary files on clean exit (#593)
- [x] Remove the `SHELLOPTS` workaround, implement `AUR_DEBUG` in `aur` instead.
- [x] Remove default options to `makepkg` (#635)
- [x] Rename `-s` to `-S` (`--sign`)
- [x] Split makepkg command to `makepkg -od` and `makepkg --noextract` if `--pkgver` is specified (#708)

### `aur-fetch`
- [x] Add `--sync` (`=auto`, `=rebase`, `=reset`) to preserve or reset local commits (#639)
- [x] Add `--results` to indicate if a given `git` repo was cloned, rebased or fetched (#379, #639)
- [x] `orderfile` is now set in `aur-sync` (`$XDG_CONFIG_HOME/aurutils/sync`)
[...]

Issue #353, opened 2018-04-14
"-d database" vs "--repo repo"


FichteFoll:
Different commands use different command flags for specification of a repo. Most notably, `aur-sync` has `--repo repo` and `aur vercmp` has `-d database`. Other affected scripts are `aur-rfilter` and `aur-build`, which both use `-d`. (At least according to the man pages, which where the simples to scan.)

It would be nice if this was unified.

AladW:
@FichteFoll what short option would you use for the recursive flag of `aur-fetch`? Right now it's also `-d`.

FichteFoll:
Is aur-fetch's -d inspired by something?

I'd try looking at what pacman does. I remember -Rs with S specifically chosen so it is synchronous to a sync action (not sure about the exact wording as I'm on mobile). Is -r taken already? That's also used frequently, like in rm, cp, rsync.

AladW:
I was looking at `cower -d`, but `-r` sounds good - thanks

FichteFoll:
:+1: 

Need to update man pages still, but other than that this seems to be done. Will leave open until it's documented.

AladW:
Updated the man pages.

PR #649, opened 2019-11-28
aur-repo: add --path

### LINKED ISSUE #555: aur-repo: alternative to command used in 1.5
I formerly used trizen -G (aurcheck -q -d revenge_repo) to fetch a list of my packages that needed updates and then automatically clone their aur repos. What would be the best alternative to this in the latest version of aurutils?

AladW:
Add the `--path` command-line option to print paths, and exit with `usage` if not specified. This way, the user must explicitely specify the operation they want to perform (issue #555).

Change `--repo-list` to only print the names of `file://` repositories, for easy processing where only these names are required (e.g. `aur-build -d` or `aur-sync -d`). The previous behavior, where paths are printed in full, can be achieved by combining `--repo-list` with `--path`.

----
In summary, we have the following options:

1. `aur repo --repo-list`
-> Print `file://` repository names
    
2. `aur repo --repo-list --path`
-> Print paths of all `file://` repositories
    
3. `aur repo --path`
-> Print path of single `file://` repository

4. `aur-repo`
-> Print usage text

----
Closes #555

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -70,7 +70,7 @@ fi
 
 ## option parsing
 opt_short='a:B:d:D:AcfnrsvLNR'
-opt_long=('arg-file:' 'chroot' 'database:' 'force' 'root:' 'sign'
+opt_long=('arg-file:' 'chroot' 'database:' 'repo:' 'force' 'root:' 'sign'
           'verify' 'directory:' 'no-sync' 'pacman-conf:' 'results:'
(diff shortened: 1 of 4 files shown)


AladW:
I also added back `--repo` as alias to `--database` for all scripts, as it seems odd to have `aur repo --repo-list` but not `aur repo --repo`. This revisits issue #353.

AladW:
Before merging, the documentation should be updated as well.

AladW:
To keep some reasonable progress, let's keep track of documentation updates in #577.

PR #652, opened 2019-12-14
aur-fetch: add --sync

AladW:
Make `aur-fetch` non-destructive by default, to either update AUR_SEEN (if `--confirm-seen` is specified) or run `git-fetch -v`. 

If `--sync=reset` or `--sync=rebase` are specified, `git reset --hard HEAD@{upstream}`, and `git reset --HARD; git merge` are run, respectively. 

`sync=auto` decides the action based on the `aurutils.rebase` configuration setting, defaulting to `--reset`, and this option is used by `aur-sync`.

----
As usual, documentation updates are missing.

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -105,11 +105,11 @@ if (( recurse )); then
 else
    printf '%s\n' "$@"
 fi | while read -r pkg; do
-    # Avoid issues with filesystem boundaries. (#274)
-    export GIT_DIR=$pkg/.git GIT_WORK_TREE=$pkg
+    # Reset/rebase if we are on valid repo
+    if GIT_DIR=$pkg/.git git rev-parse --git-dir >/dev/null 2>&1; then
+        # Avoid issues with filesystem boundaries. (#274)
+        export GIT_DIR=$pkg/.git GIT_WORK_TREE=$pkg
 
-    # Reset/rebase if we're on valid repo
-    if git rev-parse --git-dir >/dev/null; then
         if (( confirm_seen )); then
             git update-ref AUR_SEEN HEAD
 
@@ -160,10 +160,12 @@ fi | while read -r pkg; do
         fi
 
     # Otherwise, try to clone anew
-    elif git clone "$AUR_LOCATION/$pkg".git "$GIT_DIR"; then
+    elif git clone "$AUR_LOCATION/$pkg"; then
+        # Avoid issues with filesystem boundaries. (#274)
+        export GIT_DIR=$pkg/.git GIT_WORK_TREE=$pkg
+
         if (( confirm_seen )); then
             git update-ref AUR_SEEN HEAD
-
             msg2 'Marked new repository %s as seen' "$pkg"
         fi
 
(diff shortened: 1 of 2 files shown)


AladW:
@rafasc I wonder instead of using `$git_empty_tree` whether to `continue` to the next package if `AUR_SEEN` is not found. That way, `aur fetch -L` only work for cases it is defined to work on (AUR_SEEN available), rather than make an assumption that the user made no inspection through other means.

AladW:
In particular: f53f3936452ac21098e2366c8bfcbfedc8eeaff1QD

rafasc:
Seems reasonable as standalone, but considering the use in aur-sync it creates a new corner case:
The warnings would be quickly hidden by the viewer in use, and these not reviewed packages would be marked in the results file as `fetch://` with no diff, which matches the `no 
 changes` case.

Maybe marking these as 'error:' in the results file would be enough?

AladW:
I've considered to make aur-fetch exit > 0 if for some package AUR_SEEN was not available. `error:` in the results file would need to be extended to other cases (`rebase` failing, `clone` failing, ...) as well.

I guess the main intent was to not suddenly have large changesets in the viewer, assuming in good faith that anything lacking AUR_SEEN was already properly inspected. But I can see the problematic. I've left things as is (only adding an `exit 1` if `git-fetch` fails.)

AladW:
Maybe the following is an alternative, preserving the `2.3` behavior if `AUR_SEEN` is not set:
```
diff --git a/lib/aur-fetch b/lib/aur-fetch
index ab745f3..3de96cd 100755
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -134,14 +134,14 @@ fi | while read -r pkg; do
             exit 1
         }
 
-        if ! seen=$(git rev-parse --quiet --verify AUR_SEEN); then
-            warning '%s: AUR_SEEN object not found for %s, assuming empty tree' "$pkg" "$argv0"
-            seen=$(git hash-object -t tree /dev/null)
+        if seen=$(git rev-parse --quiet --verify AUR_SEEN); then
+            range=${seen:+$seen..}HEAD
+        else
+            warning '%s: %s: AUR_SEEN object not found, assuming ORIG_HEAD' "$pkg" "$argv0"
+            range=ORIG_HEAD..HEAD
         fi
 
         if [[ $seen != "$(git rev-parse HEAD)" ]]; then
-            range=${seen:+$seen..}HEAD
-
             # Contents have changed since last inspection; print differences.
             if (( verbose )); then
                 git --no-pager log "${diff_args[@]}" "$range"
```

rafasc:
Perhaps that and mark them as `clone` or something else in results, and/or make the output go into a `pkgname.unseen` file? 
I am mostly concerned about how easy such cases could go unnoticed, essentially negating the entire point of having and trusting  `AUR_SEEN`.

There were some issues in the past concerning `ORIG_HEAD` and `rebase`, because commands that run inside a rebase can modify `ORIG_HEAD` . I didn't check, but I believe currently the rebase process sets the ORIG_HEAD when the rebase finishes to the value you would expect.

But I would really prefer people avoid this problem by explicitly running `--confirm-seen`.

AladW:
Alright then, let's keep to this approach then :+1:

PR #614, opened 2019-10-13
aur-build: ensure aur-repo always gets the pacman config

Foxboron:
* [aur-build] Ensure we propegate pacman.conf to aur-repo when finding
              the repository we build against.
* [aur-build] Renamed p_conf to pacman_conf for internal consistency

* [aur-chroot] Ensure pacman.conf is propegated to aur-repo when we find
               the build repo.

PR #653, opened 2019-12-16
Allow overriding pacman.conf for repo selection

thomwiggers:
The ``pacman.conf`` file might specify an alternate location for your
packages repository – for example, it might specify the ``file://`` path
while you usually use a path on the internet. This is for example useful
if you use the same ``pacman.conf`` on a bunch of machines.

This change passes on ``--pacman-conf`` from ``sync`` to the
repo-manipulating commands.


thomwiggers:
Actually, #614 is a better way of doing this.

Foxboron:
Yes. I just rebase the patch :thinking:

PR #655, opened 2019-12-25
scripts: readd aur-chroot

AladW:
This is a compromise between the previous `aur-chroot` and a full delegation to `archbuild`. In particular, a dedicated `pacman.conf` is used for the chroot (defaulting to `/usr/share/devtools/pacman-extra.conf`, with `extra` as a configurable prefix) with no additional processing.

`aur-chroot` differs as follows from `archbuild`:

* `Include` directives are supported;
* `makechrootpkg` arguments can be overridden from the command line;
* the distinction between `multilib-devel` and `base-devel` is derived from the supplied prefix instead of `$0`;
* the `arch-nspawn`, `mkarchroot` and `makechrootpkg` steps can be used seperately;
* both `makepkg.conf` and `pacman.conf` can be specified from the command-line.

### DIFF
--- a/completions/command_opts.sh
+++ b/completions/command_opts.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
-have_optdump=('build' 'depends' 'fetch' 'pkglist' 'repo' 'repo-filter' 'query'
-              'search' 'srcver' 'sync' 'vercmp')
+have_optdump=('build' 'chroot' 'depends' 'fetch' 'pkglist' 'repo' 'repo-filter'
+              'query' 'search' 'srcver' 'sync' 'vercmp')
 no_optdump=('graph')
 
 default_opts() {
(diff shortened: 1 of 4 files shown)


AladW:
Bump? This should work as intended.

AladW:
:+1:

maximbaz:
OK there are a few issues 😄 

1. If you just try the master branch now, it fails with a cryptic error "Please specify working directory", which comes from `arch-nspawn` command, specifically:

```
$ sudo arch-nspawn -C /usr/share/devtools/pacman-aur.conf -M /usr/share/devtools/makepkg-x86_64.conf /var/lib/aurbuild/x86_64/root --bind-ro=/var/cache/pacman/maximbaz pacman -Syu --noconfirm
```

In my case, the folder `/var/lib/aurbuild/` did not exist.

I temporarily patched it with adding `--create` flag when calling `aur-chroot` from `aur-build`, that fixed the issue. We should provide `--create` by default perhaps?

1. Now I'm at the stage where this error surfaces, looking into why it happens:

```
:: Starting full system upgrade...
 there is nothing to do
Running makechrootpkg -cu
nnn-git
aur-chroot: invalid option '--bind-rw'
usage: chroot [--create] [-CDM path] -- <makechrootpkg args>
```

PR #657, opened 2019-12-30
scripts: fix use of msg/msg2

AladW:
`msg`/`msg2` using `stdout` breaks the version output in `aur-sync`, and there is otherwise no reason to use these functions for non-stderr output.


AladW:
Also added a fix where `sync=rebase` would be ignored for `aur-fetch -S`.

PR #663, opened 2020-01-15
Fix building in chroot

maximbaz:
Follow-up to https://github.com/AladW/aurutils/pull/655

Some typos fixed.

Most controversial question: as you can see I removed `--create` because I dont really see the need for it, aurutils is perfectly capable to figure out on its own if it needs to initialize chroot before building. Am I missing something?

This works when both `/var/lib/aurbuild` doesnt exist and when it is already initialized.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -11,9 +11,6 @@ PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 # default options
 chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0 results=0 prefix=aur
 
-# default options
-chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0 results=0 chroot_prefix=aur
-
 # default arguments
 chroot_args=()
 conf_args=()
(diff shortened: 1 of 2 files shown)


AladW:
Can you do a seperate PR for the create stuff, and leave this one for the typos?

AladW:
Also the -Dfoo -dfoo arguments are because aur-build passes --bind-rw $db_root, and aur-chroot adds it a second time as --bind-ro. Suggested change: remove the flag from aur-build, change --bind-ro to --bind-rw for file:// paths in aur-chroot

AladW:
https://github.com/AladW/aurutils/blob/2e5c3ef54e213314bd7b211acbae657a68a7b56b/lib/aur-build#L13-L15 some redundant lines here as well (oversight of mine in a merge conflict, probably)

maximbaz:
Done, applied your suggestion, tested that it still works 😄 

> Maybe it's worth discussing if bindmounts_rw should only be populated automatically if no --bind-rw arguments are specified on the command-line?

I don't know if I understand codebase enough to comment on this...

AladW:
Probably not worth it - I remember people specifying `-d` multiple times to `aur-build` just so that multiple `file://` paths would be mounted. Thanks!

PR #664, opened 2020-01-15
aur-chroot: create automatically when necessary

maximbaz:
I don't understand the purpose of `--create` and how it's intended to be used, I think `aur-chroot` can perfectly well understand when chroot needs to be initialized, so let's not even ask user to do so?

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -86,15 +86,15 @@ while read -r key _ value; do
     esac
 done < <(pacman-conf --config "$pacman_conf")
 
-# parent path is not created by mkarchroot (#371)
-if [[ ! -d $directory ]]; then
-    sudo install -d "$directory" -m 755 -v
-
-    sudo mkarchroot -C "$pacman_conf" -M "$makepkg_conf" \
-         "$directory"/root "${base_packages[@]}"
-fi
-
 if (( prepare )); then
+    # parent path is not created by mkarchroot (#371)
+    if [[ ! -d $directory ]]; then
+        sudo install -d "$directory" -m 755 -v
+
+        sudo mkarchroot -C "$pacman_conf" -M "$makepkg_conf" \
+             "$directory"/root "${base_packages[@]}"
+    fi
+
     # locking is done by systemd-nspawn
     sudo arch-nspawn -C "$pacman_conf" -M "$makepkg_conf" "$directory"/root \
          "${bindmounts_ro[@]/#/--bind-ro=}" \
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -14,10 +14,10 @@ makepkg_conf=/usr/share/devtools/makepkg-$machine.conf
 prefix=extra
 
 # default options
-prepare=1 build=1 create=0
+prepare=1 build=1
 
 usage() {
-    printf >&2 'usage: %s [--create] [-CDM path] -- <makechrootpkg args>\n' "$argv0"
+    printf >&2 'usage: %s [-CDM path] -- <makechrootpkg args>\n' "$argv0"
     exit 1
 }
 
@@ -25,7 +25,7 @@ source /usr/share/makepkg/util/parseopts.sh
 
 opt_short='C:D:M:'
 opt_long=('directory:' 'pacman-conf:' 'makepkg-conf:' 'no-build' 'no-prepare'
-          'create' 'prefix:' 'bind:' 'bind-rw:' 'temp')
+          'prefix:' 'bind:' 'bind-rw:' 'temp')
 opt_hidden=('dump-options' 'nobuild' 'noprepare')
 
 if ! parseopts "$opt_short" "${opt_long[@]}" "${opt_hidden[@]}" -- "$@"; then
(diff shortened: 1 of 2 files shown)


AladW:
The main benefit is that it allows to recreate an existing chroot with one flag (see https://github.com/AladW/aurutils/blob/master/lib/aur-build#L66 with the actual option missing for some reason...). Maybe this can be achieved without requiring an explicit `--create` if no chroot directory exists.

maximbaz:
Oh, I see! Let's not require explicit creation (so that when you run aur sync for the first time, it doesn't fail because of missing chroot, but it creates chroot on its own), and chroot recreation can be implemented as simply removing chroot (and next time it will be created). What do you think?

AladW:
Looks like removal is non-trivial: https://git.archlinux.org/devtools.git/tree/archbuild.in#n56 and the non-exposed functions https://git.archlinux.org/devtools.git/tree/lib/archroot.sh. So let's just rebase this PR.

maximbaz:
ok! I won't be near laptop tonight, merge if everything looks good and I'll update to master branch tomorrow morning 😉

AladW:
#️⃣🆈🅾🅻🅾👨‍💻

AladW:
I added 2 more fixes: 
https://github.com/AladW/aurutils/commit/315f3aa9bb3c73a55fb616fc681a0eb7190678f0
https://github.com/AladW/aurutils/commit/8ef7acd09dc27aee517814d6139b130d961a1f81

lib/aur-chroot
```diff
@@ -88,15 +86,12 @@ while read -r key _ value; do
     esac
 done < <(pacman-conf --config "$pacman_conf")
 
-if (( create )); then
-    # parent path is not created by mkarchroot (#371)
-    if [[ ! -d $directory ]]; then
-        sudo install -d "$directory" -m 755 -v
-    fi
+# parent path is not created by mkarchroot (#371)
+if [[ ! -d $directory ]]; then
```
AladW: Shouldn't this be moved under `prepare` as before?

maximbaz: Would make sense 👍

maximbaz:
Something I wanted to talk to you, I think branches like ad/next don't really work out, almost nobody uses them so we don't get issue reports until too late, and even when I used the branch to test, what gets merged to master is often a bit different, whether due to accidents like merge conflicts or something more significant. 

I would therefore like to propose to think if we could treat master branch as "next", i.e. accept the fact that it can be broken at times, you could say a bit #yolo, but we will gain more users testing it and faster fixes, and if master is stable for a while, we will know for sure that its safe to cut a release. 

We can chat on irc tomorrow, I just can't login from phone now...

maximbaz:
Moved inside `prepare`, tested, all works well 👍

AladW:
I guess the main issues are that 1. PRs aren't really tested sufficiently, especially mine 2. in particular, none of us seem to have sufficient time for testing/reviewing PRs. I don't think removing the protection for the master branch will help with those issues.

AladW:
Also, could you do me a favor and test the two commits in https://github.com/AladW/aurutils/pull/664#issuecomment-575321771?

maximbaz:
> Also, could you do me a favor and cherry-pick/test the two commits in #664 (comment)?

https://github.com/AladW/aurutils/commit/315f3aa9bb3c73a55fb616fc681a0eb7190678f0 lacks `\n` in the end otherwise looks good and makes sense, the https://github.com/AladW/aurutils/commit/8ef7acd09dc27aee517814d6139b130d961a1f81 I'm not sure how to reproduce failure but everything works in the positive cases of me just building some packages 🙂 

> I don't think removing the protection for the master branch will help with those issues.

There are more people besides us who are using `master` branch, so we will get a lot more testing 🙂  I also use aurutils on daily basis, so I know I personally always have motivation to keep `master` branch working 😄 

But I guess what I really want to propose is to try to avoid stagnating PRs, of course get some code review and feedback first, but try to merge PRs fast and if necessary improve later. For example, I reviewed and tested #634 a while ago, but it feels like by the time it gets merged `master` will be quite different to when I tested it 🙂 On my side I'll try to commit to review everything reasonably quickly 🙂

AladW:
> 315f3aa lacks \n in the end otherwise looks good and makes sense, the 8ef7acd I'm not sure how to reproduce failure but everything works in the positive cases of me just building some packages 🙂

If you run `aur-build` without `-a` in a directory that does not contain a PKGBUILD, without 8ef7acd you'll get a "no PKGBUILD" error twice.

> There are more people besides us who are using master branch, so we will get a lot more testing 🙂 I also use aurutils on daily basis, so I know I personally always have motivation to keep master branch working 😄

If I'm completely honest, I mostly use 2.3 - mostly due not being satisfied with `aur-fetch` in `master` yet. Otherwise I'm fine with removing the protection if @rafasc is fine with it - I'll just have to show some restraint in pushing random changes. 🤣

rafasc:
I think removing the protection would not improve the underlying issue we have, which is mostly:
> stagnating PRs

Missing documentation for things that are already merged will probably be another hang.

Better/more tests would help, but it's a Big task.

I think the big block right now is with `aur-fetch` mainly because @AladW and I have slightly different opinions on how it show work.

> try to merge PRs fast and if necessary improve later

If we merge broken code and are slow on merging the fixes, I think people will start to postpone or update to master conservatively. Which then becomes more or less untested like PRs are now.

Neither of these issues are solved by removing the protection from master.

Issue #665, opened 2020-01-19
`aur build -D /tmp/` not working after #655


bartoszek:
I've tried building in `/tmp/` with `aur build -L -f -c -D /tmp -d aur -- -T` but got this error from `aur-chroot`
```
==> ERROR: Please specify a working directory.
```
Look like `-D /tmp` is not passed to `aur chroot` here:
https://github.com/AladW/aurutils/blob/89648ed7b72013f207201c48b810149bf19a92ab/lib/aur-build#L210
bashdb trace: `bashdb /usr/lib/aurutils/aur-build -L -f -c -D /tmp -d aur -- -T`
[![asciicast](https://asciinema.org/a/vYpkfa4u7FUkdoewdN9Y9pEda.svg)](https://asciinema.org/a/vYpkfa4u7FUkdoewdN9Y9pEda)

maximbaz:
This was fixed in https://github.com/AladW/aurutils/pull/664, could you please give it another try?

bartoszek:
@maximbaz Yep, it's works now but still not honoring `aur build -D /tmp` flag.
This `aur-build/aur-chroot` section is inconsistent. It creates `chroot` in default `$directory` defined in `aur-chroot`.
https://github.com/AladW/aurutils/blob/89648ed7b72013f207201c48b810149bf19a92ab/lib/aur-build#L203-L211
Then we call `aur chroot` passing `-r /tmp` flag with `makechrootpkg_args`.
https://github.com/AladW/aurutils/blob/89648ed7b72013f207201c48b810149bf19a92ab/lib/aur-build#L272-L275
Which in turns call `makechrootpkg` with `-r $directory` and also `-r /tmp` from `aur-build -D /tmp`.
https://github.com/AladW/aurutils/blob/89648ed7b72013f207201c48b810149bf19a92ab/lib/aur-chroot#L104-L107

bartoszek:
It should looks like this
```diff
diff --git a/lib/aur-build b/lib/aur-build
index eeb0f42..efd3859 100755
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -106,7 +106,7 @@ while true; do
             repo_add_args+=(-s) ;;
         # chroot options
         -D|--directory)
-            shift; makechrootpkg_args+=(-r "$1") ;;
+            shift; chroot_args+=(-D "$1/aurbuild/$(uname -m)") ;;
         --bind)
             shift; makechrootpkg_args+=(-D "$1") ;;
[...]

PR #666, opened 2020-01-21
Fix aur-build -D flag.

bartoszek:
Rationale for #666 :smiling_imp:  explained in #655 in [this comment.](https://github.com/AladW/aurutils/issues/665#issuecomment-576718516)

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -106,7 +106,7 @@ while true; do
             repo_add_args+=(-s) ;;
         # chroot options
         -D|--directory)
-            shift; makechrootpkg_args+=(-r "$1") ;;
+            shift; chroot_args+=(-D "$1") ;;
         --bind)
             shift; makechrootpkg_args+=(-D "$1") ;;
         --bind-rw)
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -90,7 +90,9 @@ if (( prepare )); then
     # parent path is not created by mkarchroot (#371)
     if [[ ! -d $directory ]]; then
         sudo install -d "$directory" -m 755 -v
+    fi
 
+    if [[ ! -d $directory/root ]]; then
         sudo mkarchroot -C "$pacman_conf" -M "$makepkg_conf" \
              "$directory"/root "${base_packages[@]}"
     fi


lib/aur-build
```diff
@@ -106,7 +106,7 @@ while true; do
             repo_add_args+=(-s) ;;
         # chroot options
         -D|--directory)
-            shift; makechrootpkg_args+=(-r "$1") ;;
+            shift; chroot_args+=(-D "$1") ;;
```
rafasc: When we experimented with `aurbuild`: 9e709249b6c23465729056bc1edcf36748263b74 [L83](https://github.com/AladW/aurutils/commit/9e709249b6c23465729056bc1edcf36748263b74?diff=unified#diff-9ae9c47a4e26cbd989e2a9e7b916abe5L83-R87) `-D` was moved from `chroot_args` to `aurbuild_args`. Apparently we reverted it incorrectly.

lib/aur-chroot
```diff
@@ -46,7 +46,7 @@ while true; do
         -C|--pacman-conf)
             shift; pacman_conf=$1 ;;
         -D|--directory)
-            shift; directory=$1 ;;
+            shift; directory="$1/aurbuild/$machine" ;;
```
rafasc: Initially I was wondering if we really needed to force the `.../aurbuild/$machine` structure when the user is passing the path explicitly, but `mkarchroot` is really picky about the workdir arg you pass. It's an ok trade off vs massaging the path inside prepare.

AladW: I don't like this approach of hardcoding paths - where does `mkarchroot` have any requirements that a path should look like this?

bartoszek: Both `aurbuild` and prev `aur-chroot` have had an extra path component appended to one provided with `-D`. I thought this behavior is nice to keep for easy cleanup and backward compatibility with previous implementations.

But if you want `mkarchroot` to work directly in the path provided with `-D` flag, we need to adjust the `chroot prepare` section, currently it fires only if `$directory` doesn't exist.
We will end up with something like this:
```diff
diff --git a/lib/aur-build b/lib/aur-build
index eeb0f42..2d45ac1 100755
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -106,7 +106,7 @@ while true; do
             repo_add_args+=(-s) ;;
         # chroot options
         -D|--directory)
-            shift; makechrootpkg_args+=(-r "$1") ;;
+            shift; chroot_args+=(-D "$1") ;;
         --bind)
             shift; makechrootpkg_args+=(-D "$1") ;;
         --bind-rw)
diff --git a/lib/aur-chroot b/lib/aur-chroot
index 21935f2..17ebdc5 100755
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -90,7 +90,9 @@ if (( prepare )); then
     # parent path is not created by mkarchroot (#371)
     if [[ ! -d $directory ]]; then
         sudo install -d "$directory" -m 755 -v
+    fi
 
+    if [[ ! -d $directory/root ]]; then
         sudo mkarchroot -C "$pacman_conf" -M "$makepkg_conf" \
              "$directory"/root "${base_packages[@]}"
     fi
```

rafasc: > where does `mkarchroot` have any requirements that a path should look like this?

It requires that all path components must exist except the last. (`readlink -f`). Which doesn't allow us just rely on it by simply give it "$directory/root", we would need the multiple checks mentioned in the comment above.

@bartoszek has mentioned the good points of keeping at least an extra level hardcoded. Other minor ones could include the namescheme from arch-nspawn lock files being quite simple (user.lock) having the extra layer could help avoid collisions, especially when used directly in /tmp which I assume is a common use case for this flag.

bartoszek: @rafasc One more thing came to my mind. We might also check if the `$directory` mount point has `nosuid` flag set up. This prohibits `mkchrootpkg` to run but not `mkarchroot`.
```sh
if [[ $(findmnt -n -o OPTIONS --target "$directory") =~ nosuid ]]; then
  error "Unable to build inside \"%s\" as it's mounted with \"nosuid\" option." $(findmnt -n -o TARGET --target "$directory")
  exit 2
fi
```

Earnestly: Let the code fail naturally.

rafasc: I don't like all this massaging. :unamused:
The way this is going, i'd rather tell users to use <code>-D /tmp/<em>something</em></code>

AladW: I guess the proposed change of 
```
     if [[ ! -d $directory ]]; then
         sudo install -d "$directory" -m 755 -v
+    fi
 
+    if [[ ! -d $directory/root ]]; then
         sudo mkarchroot -C "$pacman_conf" -M "$makepkg_conf" \
              "$directory"/root "${base_packages[@]}"
     fi
```
is reasonable. It follows the mentioned simple naming scheme (`$USER`, root) and doesn't hardcode `.arch-chroot` like in 2.3.

AladW:
@bartoszek any follow up?

bartoszek:
I've made the requested changes.
* using `-T` switch results in a bunch of `$USER-$PID{.lock}` dirs/files in `$directory` which is a bit tedious to clean up, and breaks established behavior.
* if we can determine beforehand that `$directory` can't facilitate the chroot, we should do it (or at least put a disclaimer in `aur-{sync,build}` man page.

<details>
<summary>btw.</summary>
_Sorry for the delay, thought you'll just mush your changes to this PR to mach your vision and merge, or make your own PR and close this one as resolved :thinking:_
</details>

AladW:
> Sorry for the delay, thought you'll just mush your changes to this PR to mach your vision and merge, or make your own PR and close this one as resolved 🤔

Uh, I'd need to like, push more buttons for that. Thanks for the update.

> using -T switch results in a bunch of $USER-$PID{.lock} dirs/files in $directory which is a bit tedious to clean up, and breaks established behavior.

Established where? I don't see how this is any different from 2.3.

I don't mind to extend the documentation, in any case.

lib/aur-chroot
```diff
@@ -46,7 +46,7 @@ while true; do
         -C|--pacman-conf)
             shift; pacman_conf=$1 ;;
         -D|--directory)
-            shift; directory=$1 ;;
+            shift; directory="$1" ;;
```
AladW: unneeded change

lib/aur-chroot
```diff
@@ -88,9 +88,11 @@ done < <(pacman-conf --config "$pacman_conf")
 
 if (( prepare )); then
     # parent path is not created by mkarchroot (#371)
-    if [[ ! -d $directory ]]; then
+    if [[ ! -d "$directory" ]]; then
```
AladW: unneeded change

lib/aur-chroot
```diff
@@ -88,9 +88,11 @@ done < <(pacman-conf --config "$pacman_conf")
...
         sudo install -d "$directory" -m 755 -v
+    fi
 
+    if [[ ! -d "$directory/root" ]]; then
```
AladW: quotes aren't needed with `[[`

AladW:
Added some nitpicks.

lib/aur-chroot
Earnestly: On the LHS (which this is)

man1/aur-build.1
```diff
@@ -172,6 +172,20 @@ subdirectory that serves as template for user containers (named after
 or
 .B /copy
 if unset).
+.IP
```
AladW: I wonder if this is not too much of an edge case to document. I was more thinking of documenting that users may want a dedicated directory (such as the mentioned `/tmp/something`) for the container, if that's not obvious already.

bartoszek: Well, beside `/` every file-system is, or at leas should be, mounted with `nosuid`.
It's a default option for every `tmpfs` on Arch, which basically prevent user from building in ram with `-D /tmp` without preforming the remount first, or creating temporary `tmpfs` with `suid` to facilitate `chroot`.

Please excuse my boldness, but would you consider adding `--ram` switch to `aur-build` to automatically create temporary ram drive with `suid` to host chroot build?
```diff
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -9,7 +9,7 @@ XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
 # default options
-chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0 results=0 prefix=aur
+chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0 ram_drive=0 results=0 prefix=aur
 
 # default arguments
 chroot_args=()
@@ -36,6 +36,9 @@ trap_exit() {
         printf >&2 'AUR_DEBUG: %s: temporary files at %s\n' "$argv0" "$tmp"
         printf >&2 'AUR_DEBUG: %s: temporary files at %s\n' "$argv0" "$var_tmp"
     fi
+    if (( ram_drive )); then
+        sudo umount $(ram_drive_path)
+    fi
 }
 
 usage() {
@@ -97,6 +100,9 @@ while true; do
         --pkgver)
             run_pkgver=1
             makepkg_args+=(--holdver) ;;
+        --ram)
+            ram_drive=1
+            chroot=1 ;;
         --results)
             shift; results_file=$1 ;;
         --root)
@@ -200,6 +206,13 @@ if [[ -v results_file ]]; then
     results=1
 fi
 
+if (( ram_drive )); then
+    # Create temporary ram drive
+    ram_drive_path=$(mktemp -d --tmpdir "aurutils-aurbuild.XXXXXXXX") || exit
+    sudo mount -t tmpfs -o suid tmpfs "$ram_drive_path"
+    chroot_args+=(-D "$ram_drive_path")
+fi
+
 if (( chroot )); then
     # Defaults to /usr/share/devtools/pacman-aur.conf
     chroot_args+=(--prefix "$prefix")

Comment (Earnestly) [reply man1/aur-build.1]: `nosuid` doesn't prevent building in a tmpfs (`/tmp`).  Nor does it prevent the creation of `suid` binaries:

```
% /tmp touch foo

% /tmp chmod -c 4755 foo
mode of 'foo' changed from 0644 (rw-r--r--) to 4755 (rwsr-xr-x)

% /tmp stat foo
  File: foo
  Size: 0               Blocks: 0          IO Block: 4096   regular empty file
Device: 35h/53d Inode: 143782      Links: 1
Access: (4755/-rwsr-xr-x)  Uid: ( 1000/ earnest)   Gid: (  100/   users)
Access: 2020-02-03 16:31:20.745049394 +0000
Modify: 2020-02-03 16:31:20.745049394 +0000
Change: 2020-02-03 16:31:27.305066277 +0000
 Birth: -
```

Comment (AladW) [reply man1/aur-build.1]: >  would you consider adding --ram switch

As I already said: No. This is far out of scope for anything aur-chroot does. If you want that functionality, it's easy enough to add it to a wrapper script that calls aur-chroot. 

> nosuid doesn't prevent building in a tmpfs (/tmp).

Well, there's no reason to document it does, then.

Comment (bartoszek) [reply man1/aur-build.1]: @Earnestly you are right, you can chroot to this path, [but when `makechrootpkg` drops root privileges](https://github.com/archlinux/devtools/blob/4c08847bfa7776abb884fef289a25c1cef9883fb/makechrootpkg.in#L216), suid bit is not preserved resulting in sudo error.
```
==> Installing missing dependencies...
sudo: effective uid is not 0, is /usr/bin/sudo on a file system with the 'nosuid' option set or an NFS file system without root privileges?
==> ERROR: 'pacman' failed to install missing dependencies.
```
[asciicast](https://asciinema.org/a/iGap07zcOXkAkIOy7uVSmsGKY)

Comment (Earnestly) [reply man1/aur-build.1]: Ah so now the real problem shows itself which is attempting to use `sudo` in a `nosuid` mount.  In such a case can't you `-o remount,suid` beforehand since you have root priviledge?

Potentially dependencies can be brought into the chroot before entering it but I'm not sure of that workflow and how it interacts.

Comment (bartoszek) [reply man1/aur-build.1]: > In such a case can't you `-o remount,suid` beforehand since you have root priviledge?
This is what is stated in my `man:arch-build -D` disclaimer we are talking about now.
```
       -D DIR, --directory=DIR
              The base directory for containers. Defaults to /var/lib/archbuild/<chroot-prefix>-<arch>.

              This directory usually contains a /root subdirectory that serves as template for user containers (named after $SUDO_USER, or /copy if unset).

              Mount point for DIR need to have nosuid option disabled.
              # findmnt -o TARGET,OPTIONS -T DIR
              To disable nosuid
              # mount -o remount,suid $(findmnt -no TARGET -T DIR)
```

Earnestly: What I mean to suggest is can this be done outside of aurutils?  Ultimately aurutils is a collection of utilities that are intended to be used with custom scripts, ones such as which can set up chroots, change mount options, or whatever else.

bartoszek: Yep, as mentioned by @AladW, you can create wrapper around `aur-utils` which e.g. creates/destroys costume `tmpfs` and pass it as argument for `-D` switch. But if we rely on user to do heavy lifting, we could at lest mention this in the manual, as this isn't an edge case but a common scenario to use `tmpfs` when building in chroot.

Earnestly: Fair, knowing nothing else I'd agree that it might be worth mentioning in the `NOTES` manual.

AladW: I can live with it being in `NOTES`, mentioning an edge case prominently in the option documentation directly is not something I'd do.

AladW:
I finally want to merge this so it'd be great if this PR only contains https://github.com/AladW/aurutils/pull/666/commits/96cc0b790d19edde3b224ef1ed9c87bb5ced9131 and we leave the rest for later.

PR #667, opened 2020-01-22
Fix missing/unused flags in `aur-{build,chroot}`

bartoszek:

PR #668, opened 2020-01-23
Run shellcheck and option parsing check on pushes and pull requests

rafasc:
I'm willing to merge a PR that adds the `parseopt-consistency` test.
_Originally posted by @AladW in https://github.com/AladW/aurutils/pull/667#issuecomment-577803145_

I also included a small github action that runs it automatically on pushes and pull requests.
For now, only the parseopt check is done.

Later, we can include other tests and the shellcheck run. I left it out of this PR because I still didn't investigate how to cache/reuse the expensive `pacman -Syu`.

Or I can drop the actions commit for a later PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,10 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    container: archlinux/base
+    steps:
+      - run: pacman -Syu --noconfirm base base-devel
+      - uses: actions/checkout@v2
+      - run: test/parseopt-consistency
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,12 @@ jobs:
     runs-on: ubuntu-latest
     container: archlinux/base
     steps:
-      - run: pacman -Syu --noconfirm base base-devel
+      - run: pacman -Syu --noconfirm base base-devel shellcheck
       - uses: actions/checkout@v2
-      - run: test/parseopt-consistency
+
+      - name: Option parsing consistency
+        run: test/parseopt-consistency
+
+      - name: Shellcheck
+        if: success() || failure() # Do not fail-fast
+        run: make shellcheck
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -139,7 +139,7 @@ while true; do
             repo_add_args+=(-p) ;;
         # other options
         --dump-options)
-            printf -- '--%s\n' "${opt_long[@]}"
+            printf -- '--%s\n' "${opt_long[@]}" ${AUR_DEBUG+"${opt_hidden[@]}"}
             printf -- '%s' "${opt_short}" | sed 's/.:\?/-&\n/g'
             exit ;;
         --) shift; break ;;
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -56,7 +56,7 @@ while true; do
             shift; bindmounts_rw+=("$1") ;;
         # other options
         --dump-options)
-            printf -- '--%s\n' "${opt_long[@]}"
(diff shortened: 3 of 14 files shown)


rafasc:
[](https://github.com/rafasc/aurutils/runs/406090833) as expected as mentioned in #667

.github/workflows/ci.yml
```diff
@@ -0,0 +1,10 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  tests:
+    runs-on: ubuntu-latest
```
AladW: Magic 🧙‍♂️

AladW:
I thought `shellcheck` might complain on the unquoted AUR_DEBUG, but looks like it doesn't!

AladW:
> running the test script under an unprivileged user would require we set up one inside the container.

Is that a hard thing to do? I'd never thought `AUR_ASROOT` would be used for something like this 😆

Earnestly:
You can use the `nobody` user for that purpose.  Make sure you create a directory owned by `nobody` to use it.

rafasc:
Added shellcheck.

---
The easiest way to get around the root issue (besides setting  AUR_ASROOT=1) would be to run it directly on the host machine (ubuntu-latest). But we depend on things being sourced from archlinux. c.f `/usr/share/makepkg/util/{parseopt,message,util}.sh`

Alternatively, we could write our own Docker-based github action that sets up the correct environment, including creating the user, setting a repo to be used with other tests, etc.

The problem with this is that the pacman -Syu needs to be cached and I still didn't look into how that can be done. 

Another could be to publish our own archlinux image with the necessary dependencies to the registry. Then change `container: archlinux/base` to our own and remove the `pacman -Syu` step. Then set up something to rebuild it regularly.

To use the `nobody` user we would need to run su -s /bin/bash -c blabla nobody for each step, as they run in a different subshell. We would also run into permission issues because the `nobody` user does not have permission to write in the working directory.

So I went the easy route and disabled the molly-guard.

AladW:
I'm not sure I understand the involved complexity. Compare e.g. https://github.com/AladW/aurutils/blob/ad/next/Dockerfile#L9-L10 - couldn't that replicated in the `.yml` file?

rafasc:
We could do something like this, sure:
``` yaml
    steps:
      - name: Container Setup
        run: pacman -Syu --noconfirm base base-devel shellcheck
      
      - name: Set up unprivileged user
        run:  useradd -m -s /bin/bash archie

      - uses: actions/checkout@v2
      
      - run: chown -R archie $GITHUB_WORKSPACE

      - name: Option parsing consistency
        run: su archie -c test/parseopt-consistency

      - name: Shellcheck
        if: success() || failure() # Do not fail-fast
        run: su archie -c 'make shellcheck'
```
But I do not see the point for this specific job.

What I really dislike is that 50% of the time is spent hitting the mirrors with `run: pacman -Syu --noconfirm base base-devel shellcheck`.

AladW:
Would pointing to a snapshot from `archive.archlinux.org` help?

AladW:
https://github.com/actions/cache 🤔

AladW:
Considering it's already useful in its current state, let's merge this and figure out the details later.

rafasc:
I think we could use `actions/cache` + `archive.archlinux.org`, using the `archive.archlinux.org/repos/week/lastupdate` as the cache key.

Another thing I noticed with issue 672, is that pull requests originating from aladw/aurutils will trigger it twice. Once from the push, and another from the PR.

AladW:
> I think we could use actions/cache + archive.archlinux.org, using the archive.archlinux.org/repos/week/lastupdate as the cache key.

👍 

> Another thing I noticed with issue 672, is that pull requests originating from aladw/aurutils will trigger it twice. Once for the push, and another for the PR.

I noticed the same, wasn't sure if it's normal or not...

PR #672, opened 2020-01-26
aur: respect AUR_EXEC_PATH

### LINKED ISSUE #645: Tests use installed aurutils instead of scripts in the repository
Tests are running the installed aurutils, where we should be testing the scripts in `lib/aur-*` instead.

This could be solved by placing a new `aur` script with the customized `AUR_LIB_DIR` early on `PATH` or by removing the limitations we imposed and make `aur` respect `AUR_LIB_DIR` at runtime, allowing us to set AUR_LIB_DIR appropriately on each test.

AladW:
Defaults to `AURUTILS_LIB_DIR`, which is defined in the Makefile. The variable name is inspired by `GIT_EXEC_PATH`.

Fixes #645

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,14 +3,14 @@ PREFIX ?= /usr
 SHRDIR ?= $(PREFIX)/share
 BINDIR ?= $(PREFIX)/bin
 LIBDIR ?= $(PREFIX)/lib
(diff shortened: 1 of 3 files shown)


AladW:
Done.

PR #675, opened 2020-02-01
Fix aur-fetch of more than one package

dhouck:
There is a fix for this as part of #659, but the discussion on that has been happening for a month, and (at least as a newcomer to the project) it doesn’t look like it’ll be ready to merge soon. In the meantime, #636 is still broken, and it’s only a one-line fix.

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -112,6 +112,8 @@ if (( recurse )); then
 else
    printf '%s\n' "$@"
 fi | while read -r pkg; do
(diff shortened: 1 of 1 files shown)


AladW:
Sorry, I meant to reply earlier. I think it looks fine, I'd just put `unset` at the top of the loop body; in case someone decides to try `GIT_DIR=foo aur-fetch`.

dhouck:
Makes sense; I’ve moved the line there.

PR #677, opened 2020-02-17
aur-sync: fix passing args after '--'

bartoszek:
We have to strip `packagename` from `${@}` when setting `$build_extra_args`. Otherwise calling `aur-sync` with extra flags after `--` won't work.
Rationale:
[asciicast](https://asciinema.org/a/gei6YyhCiP9SqbSCgxdwrf65a)
_`aur-build` is free of the issue, as it doesn't trigger `unused_argv` in `parseopts`_

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -207,7 +207,8 @@ trap 'trap_exit' EXIT
 
 # Append remaining options to the aur-build command-line
 if (( $# )); then
-    build_extra_args+=("$@")
+    # shellcheck disable=SC2207
+    build_extra_args=($(grep -o -- "-- .*" <<<"${@}"))
 fi
 
 # Append contents of ignore file


lib/aur-sync
```diff
@@ -207,7 +207,7 @@ trap 'trap_exit' EXIT
 
 # Append remaining options to the aur-build command-line
 if (( $# )); then
-    build_extra_args+=("$@")
+    mapfile -t -d' ' build_extra_args < <(grep -oP ".*?\K-- .*" <<<"$@")
```
bartoszek: This could be also [shorter](https://github.com/AladW/aurutils/commit/33eaf878ea2ad0b20af25e0708175d07593f631b) but will need a shellcheck mask.

lib/aur-sync
```diff
@@ -207,7 +207,8 @@ trap 'trap_exit' EXIT
 
 # Append remaining options to the aur-build command-line
 if (( $# )); then
-    build_extra_args+=("$@")
+    # shellcheck disable=SC2207
+    build_extra_args=($(grep -o -- "-- .*" <<<"${@}"))
```
Earnestly: ಠ_ಠ

bartoszek: :feelsgood: challenge accepted #678

PR #678, opened 2020-02-17
aur-sync: fix positional arguments parsing.

bartoszek:
Consume positional arguments in one go instead of carrying them around, this fixes #677 and makes script feels sturdier @Earnestly

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -205,9 +205,19 @@ tmp_view=$(mktemp -d --tmpdir "aurutils-$argv0-view.XXXXXXXX") || exit
 
 trap 'trap_exit' EXIT
 
+# Extract 'pkgname' from command-line.
+unset pkgname
+while true; do
+    case "$1" in
+        --) shift; break ;;
+        *) pkgname+=("$1"); shift ;;
+    esac
+done
+
 # Append remaining options to the aur-build command-line
 if (( $# )); then
     build_extra_args+=("$@")
+    shift $#
 fi
 
 # Append contents of ignore file
@@ -223,7 +233,7 @@ if (( rotate )); then
     fi
 fi
 
-if ! (( $# + update )); then
+if ! (( ${#pkgname[@]} + update )); then
     error '%s: no targets specified' "$argv0"
     exit 1
 fi
@@ -245,8 +255,8 @@ else
     exit 13
 fi
 
-{ if (( $# )); then
-      printf '%s\n' "$@"
+{ if [[ -v pkgname[@] ]] ; then
+      printf '%s\n' "${pkgname[@]}"
   fi
 
   if (( update )); then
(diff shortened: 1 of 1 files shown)


lib/aur-sync
```diff
@@ -245,14 +255,13 @@ else
...
-      printf '%s\n' "$@"
-  fi
+if [[ -v pkgname[@] ]] ; then
+    printf '%s\n' "${pkgname[@]}" >argv
+fi
 
-  if (( update )); then
-      aur vercmp --quiet <db_info
-  fi
-} >argv
```
bartoszek: Breaking the grouping emphasize redirection.

AladW:
The issue in #677 makes zero sense to me and I don't feel like watching a 9 minute video to get explanations. Please give some simple reproducible use case. 

Also note I'm unwilling to hack around `parseopts`. If there's an actual issue, it should either be fixed in `parseopts`, or `parseopts` should be replaced with something else.

lib/aur-sync
AladW: No reason to do this at all.

bartoszek:
`parseopts` is working as it supposed to, problem lays here.
https://github.com/AladW/aurutils/blob/c4b218f8177ca447916569e19b02fed036ca3db6/lib/aur-sync#L210
If you call `aur sync pkgname -- --testflag` you get `build_extra_args=("pkgname" "--" "--testflag") which is getting passed to `aurbuild` and breaks options propagation.
In #677 I was suggesting to sanitize it with simple grep call, but this solution looks a bit junky, as pointed by @Earnestly, so I've made more solid patch which address the issue as a whole.

lib/aur-sync
```diff
@@ -205,9 +205,19 @@ tmp_view=$(mktemp -d --tmpdir "aurutils-$argv0-view.XXXXXXXX") || exit
 
 trap 'trap_exit' EXIT
 
+# Extract 'pkgname' from command-line.
+unset pkgname
+while true; do
```
rafasc:
```suggestion
while (( $# )) ; do
```
Otherwise when processing `pkgname` as the last argument, on the next loop `shift` will exit with error causing the `sync` to terminate as it has `set -o errexit`

bartoszek: Thanks.

rafasc:
@AladW 
The issue is easily demonstrated with this silly example:
`lib/aur-sync -d trash --no-view aurutils -- -- -h`

If we are passing `-h` to `makepkg` It should do nothing except show the help. In master it is completely ignored. This PR fixes this.

The underlying issue is that `sync` did not forward non-option before. And when support for that was added the use case where non-options are a mix of package names and options to be forwarded was not considered. I.e. `sync aurutils -- -- -h` will try to call `aur build something something >aurutils< -- -h`

lib/aur-sync
bartoszek: _revert_

AladW:
After a long series of experiments I've decided to stop supporting arguments passed to `aur-build` from `aur-sync`. 

If there are any makepkg options you feel are missing in `aur-sync`, please file a seperate issue/PR for those.

Issue #679, opened 2020-02-22
Install: Scripts not being linked?


JhonnyJason:
Hello guys!

For any command like:
`aur-search ` or `aursearch` - the binary is appearently not available in the usual `$PATH`s
As of `/usr/bin/` for example.

I can see the script files being properly installed to `/usr/lib/aurutils/` - but they are appearently not linked as binary to be available after install.

I can do it manually and then it works.

Quick fix is:

```sh
# ln -s /usr/lib/aurutils/* /usr/bin/
```

Cheers!

=====

1) Do you use the latest release of aurutils and its dependencies? If so,
is the issue reproducible from the master branch?

yes

2) Does the package conform to 
[`PKGBUILD(5)`](https://www.archlinux.org/pacman/PKGBUILD.5.html) and the 
[AUR package guidelines](https://wiki.archlinux.org/index.php/Arch_User_Repository#Submitting_packages)?

IDK - it's your package^^

3) Does the package provide the correct metadata on the 
[AUR RPC interface](https://aur.archlinux.org/rpc.php)?

IDK - it's your package^^

4) Does the package build with `makepkg -s` or `extra-x86_64-build` ?

yes

5) Is the problem reproducible, and not due to a misconfiguration of
`pacman`, `makepkg`, `sudoers`, `gpg` or others?

Probably - I cannot imagine what could be misconfigured...

JhonnyJason:
I just noticed that only `aur` is installed properly as binary and that the suggested usage is:

```sh
$ aur [command]

available commands:
build		graph		repo-filter	sync
chroot		jobs		rpc		vercmp
depends		pkglist		search		vercmp-devel
fetch		repo		srcver
```

I could only figure out by looking at the Makefile xD
So issue is more about available documentation ;-) 
I cannot even type `man aur` if I don't know that the only valid binary to be executed is `aur`.

I found this tool here https://disconnected.systems/blog/archlinux-repo-in-aws-bucket/#aurutils-building-and-managing-packages

And then looking at the scripts it would suggest the binaries are rather called `aur-search` than `aursearch`. xD

Cheers!

AladW:
Install instructions from blog posts aren't supported. Apparently it documents an old version of aurutils (`1.5`) and hasn't been updated since.

PR #681, opened 2020-02-22
Readme Update to be more clear about usage

JhonnyJason:
#679


JhonnyJason:
Ah another update...^^

PR #682, opened 2020-02-22
Readme Update to be more clear about usage

JhonnyJason:
#679

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Upgrade notices are posted to the
 
 The main binary is `aur`.
 
-All further capabilities are called by specifying the command as an option
+All further capabilities are called by providing its specific command as the first argument
 ```sh
 $ aur <command>
 ```
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Upgrade notices are posted to the
 
 The main binary is `aur`.
 
-All further capabilities are called as commands of 
+All further capabilities are called by specifying the command as an option
 ```sh
 $ aur <command>
 ```
@@ -48,7 +48,7 @@ available user commands:
 
 ```
 
-For most complete documentation check the [manual](https://github.com/JhonnyJason/aurutils/blob/master/man1/aur.1) ;-)
+For most complete documentation check the [manual](man1/aur.1) ;-)
 
 ```sh
 $ man aur
--- a/README.md
+++ b/README.md
@@ -26,9 +26,33 @@ Upgrade notices are posted to the
 
 ## USAGE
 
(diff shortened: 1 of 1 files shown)


AladW:
Copying the command-line output to the readme doesn't make much sense to me. Usage is mentioned in `aur(1)`, including several examples, which is mentioned explicitely.

JhonnyJason:
Humm IMO It does make sense a lot - it's not the point where redundancy is to be fought against xD

One could even automate to print the output into the Readme ;-)

Even print the whole manual into the Readme in a format where it is easier to be read. (Maybe, but maybe it would be even too much then)
Best option would be to have an even more explanatory whole documantation in the Readme. Of course that would take time and effort^^.

This would be a fast fix. And it even has an advantage - 1:1 relation to how it looks in action.

JhonnyJason:
The argument about user effort. I would debate that in a longer discussion thread. In very short: it's not just a matter of effort also a matter of strategy.

For now it's just me who had the experience of being confused for a day, crawling through multiple ressources and don't know what is really going on until - I luckiliy figured it out.

I'm sure I'm not the only one who has ran into this trouble^^

So hopefully the discussion is about how to adjust the readme, and not if adjust at all?

Earnestly:
The usage section included in the README explicitly calls for one to read the manual and provides a link to it.

AladW:
aurutils was never meant to be an all-encompassing, minimal effort solution to AUR packages. It instead gives users tools for common AUR and packaging tasks, so they can focus on whatever specific AUR workflow they have in mind rather than worry about abstruse technical details. Each of these tools are documented in their own man pages, with an overview provided in `aur(1)`. 

Copy pasting terminal output for the "main binary" (`aur` is really only a wrapper for commands in lib, with no other functionality) as here doesn't lead to the conceptual understanding those man pages provide and the next steps would then be to copy paste local repository creation instructions and `aur sync` commands with no further thought. If you want to do that, there's other projects like `trizen` or `aurto` that are more suited.

Side-note: It would be nice if https://github.com/AladW/aurutils/blob/master/man1/aur.1 was pre-rendered be but github doesn't support this. Any conversions (to HTML) have to be done manually, leading to possible discrepancies. If automatic conversion is possible similar to what we do with github actions and `shellcheck`, I'm open for ideas. But the current proposal is not something I'm willing to merge for the reasons outlined above.

JhonnyJason:
ok unfortunate^^ - probably the question thread will help others who encounter the same problem.

Cheers! And thanks for the work on those tools btw! Helping me now for my AUR workflows alot ;-)

PR #684, opened 2020-02-24
Remove "See also" from README, add rebuild-detector to man examples

maximbaz:
Something like this @AladW? 🙂 

I've tried to guess the syntax of man files but I'm not actually sure how to check if it's good...?

### DIFF
--- a/man1/aur-build.1
+++ b/man1/aur-build.1
@@ -377,7 +377,10 @@ for details.
 It is sometimes needed to rebuild packages when their dependencies are
 updated, for example in case of dynamic library linking.
 .PP
-You can use https://github.com/maximbaz/rebuild-detector
+You can use
+.UR https://\:github.com/\:maximbaz/\:rebuild-detector
+rebuild-detector
+.UE
 to detect which packages need to be rebuilt.
 .
 .SH CHROOT NOTES
--- a/man1/aur-build.1
+++ b/man1/aur-build.1
@@ -373,6 +373,13 @@ See
 .BR gpg (1)
 for details.
 .
+.SS Rebuilding packages against updated dependencies
+It is sometimes needed to rebuild packages when their dependencies are
+updated, for example in case of dynamic library linking.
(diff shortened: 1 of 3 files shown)


AladW:
Thanks. You can check man pages by running `man -l foo.1` with `foo.1` the file you want to test.

man1/aur.1
```diff
@@ -421,6 +421,15 @@ variable towards it:
     $ export PACMAN=/usr/local/bin/mypacman
 .EE
 .
+.SS Rebuilding packages against updated dependencies
```
AladW: I might leave this section in `aur-build(1)`, if we want to keep it in `aur(1)` I'd write a specific command-line in the `EXAMPLES` section

Earnestly:
Or `man ./foo.1`

man1/aur.1
maximbaz: Moved to the `NOTES` section of `aur-build(1)`, or you have a better location?

man1/aur-build.1
```diff
@@ -373,6 +373,13 @@ See
 .BR gpg (1)
 for details.
 .
+.SS Rebuilding packages against updated dependencies
+It is sometimes needed to rebuild packages when their dependencies are
+updated, for example in case of dynamic library linking.
+.PP
+You can use https://github.com/maximbaz/rebuild-detector
```
AladW: Groff has some special syntax for URIs too:
```
.UR https://\:github.com/\:maximbaz/\:rebuild-detector
.UE .
```

man1/aur.1
AladW: The location looks OK to me

man1/aur-build.1
maximbaz: Nice - seems like even named URLs work, I saw an example in `aur-pkglist`, updated 👍

AladW:
👍

PR #689, opened 2020-03-14
aur-sync: add AUR_CONFIRM_PAGER

### LINKED ISSUE #673: Viewing differences just shows directory listing (without `vifm`)
When aurutils wants me to review a package’s contents, it opens `less` but just shows a list of files. I then need to manually use `:e` to examine them. I have neither `PAGER` nor `AUR_PAGER` set, but the problem persists even when I try different environment variable values. I do not have `vifm` installed.

__Before filing a bug, check the following:__

1) Do you use the latest release of aurutils and its dependencies? If so,
is the issue reproducible from the master branch?

I used the latest release; I’ll try on the master branch in a bit and post an update.

2) Does the package conform to 

I believe so, but my issues is seen across a wide variety of packages and before that should be relevant.

3) Does the package provide the correct metadata on the 

See above

4) Does the package build with `makepkg -s` or `extra-x86_64-build` ?

Yes, but see above

5) Is the problem reproducible, and not due to a misconfiguration of
`pacman`, `makepkg`, `sudoers`, `gpg` or others?

Yes

__If yes to all, create a debug log:__

Doing this created another problem: the pager wasn't used at all; I didn't get a chance to review the files (I had done so before running the command, in this case, but still that’s an issue). You can see at the top that it just lists the files, and does not show their contents.

```
AUR_DEBUG=1 AURDEST=${PWD} aur sync -c btrfs-snapshot > aurutil.log
```
[aurutil.log](https://github.com/AladW/aurutils/files/4136840/aurutil.log)

AladW:
Generalize an additional confirmation prompt for generic viewers set with `AUR_PAGER`, for example for use with GUI file managers without means to exit with status >0 if files were not reviewed or accepted by the user.

Fixes #673

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -321,45 +321,34 @@ fi
 # Link build files in the queue (absolute links)
 parallel -X ln -s "$(pwd -P)"/{} "$tmp_view" :::: "$tmp"/queue
 
-# Explicitly define if exit success is to be trusted (#561)
-trust_exit_success=0
-
-if type -P >/dev/null vifm; then
-    # Avoid directory prefix in printed paths (#452)
-    viewer() ( cd "$1"; vifm -c 'set vifminfo=' -c 'view!' -c '0' - )
-    trust_exit_success=1
-else
-    # shellcheck disable=SC2086
-    viewer() ( cd "$1"; command -- ${PAGER:-less -K -+F} )
-
-    # Trust our fallback
-    if [[ -z $PAGER ]]; then
-        trust_exit_success=1
-    fi
-fi
-
 # Check if dependency graph is valid
 if (( graph )); then
     aur graph "$tmp_view"/*/.SRCINFO >/dev/null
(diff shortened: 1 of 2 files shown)


lib/aur-sync
```diff
@@ -321,45 +321,34 @@ fi
...
-
 # Check if dependency graph is valid
 if (( graph )); then
     aur graph "$tmp_view"/*/.SRCINFO >/dev/null
 fi
 
 if (( view )); then
-    # AUR_PAGER exit codes are respected by definition.
+    # Avoid directory prefix in printed paths (#452)
+    viewer() ( cd "$1"; vifm -c 'set vifminfo=' -c 'view!' -c '0' - )
```
maximbaz: move this to the `elif type -P vifm >/dev/null; then` branch, perhaps?

AladW: Good point, done

PR #691, opened 2020-03-14
aur-fetch: remove AUR_SEEN

AladW:
A downside of `AUR_SEEN` is that removing the git repositories (e.g the `aur-sync` cache) also removes the reference to the "last seen" revision by the user. This PR removes `AUR_SEEN` from `aur-fetch`, and instead extends `--results` such that higher level tools like `aur-sync` can use their own diff mechanisms which do not rely on git repositories. (I will give the implementation for the latter in a later PR.) 

A side benefit is the reduced complexity in `aur-fetch`, such as absence of the seperate "mode" `--confirm-seen`. The log mechanism is otherwise kept for standalone use of `aur-fetch`.

Also extend `--sync=auto` to only make changes, i.e. run `git-rebase` or `git-reset`, if new upstream revisions are available. Otherwise, perform no action.


AladW:
shipit

PR #694, opened 2020-03-15
aur-sync: remove additional build arguments after --

AladW:
This is problematic in `aur-sync` as both program names and option names (for `aur-build`) can be placed after `--`. 

See issue #678.


lib/aur-sync
```diff
@@ -201,11 +201,6 @@ tmp_view=$(mktemp -d --tmpdir "aurutils-$argv0-view.XXXXXXXX") || exit
 
 trap 'trap_exit' EXIT
 
-# Append remaining options to the aur-build command-line
-if (( $# )); then
-    build_extra_args+=("$@")
```
maximbaz: that variable can be entirely removed from the file

AladW: Done

PR #703, opened 2020-03-28
scripts: switch to curl

AladW:
curl can take input from `stdin` through the `-K-` option. Removes the dependency on `wget`.

This affects pretty much all of `aurutils` so there should be no bugs here.

### DIFF
--- a/makepkg/PKGBUILD
+++ b/makepkg/PKGBUILD
@@ -11,11 +11,12 @@ changelog=aurutils.changelog
 sha256sums=('SKIP')
 conflicts=('aurutils')
 provides=("aurutils=${pkgver%%.r*}")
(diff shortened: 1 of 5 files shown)


makepkg/PKGBUILD
```diff
@@ -11,11 +11,12 @@ changelog=aurutils.changelog
 sha256sums=('SKIP')
 conflicts=('aurutils')
 provides=("aurutils=${pkgver%%.r*}")
-depends=('git' 'jq' 'pacutils' 'wget')
+depends=('git' 'jq' 'pacutils')
```
maximbaz: I know that curl comes as transitive dependency of git, but maybe we should add explicitely?

AladW: WFM

AladW:
face

Issue #706, opened 2020-03-30
Fails to retrieve some packages after the switch to curl


terencode:
```
aur sync carla-bridges-win32-git

==> Using [aur] repository
Warning: <stdin>:1: warning: 'url' requires parameter
curl: no URL specified!
curl: try 'curl --help' for more information
```

I wonder if it happens because of the  '-' as mingw-w64-gcc also has the same problem.

When running `AUR_DEBUG=1 aur sync  "carla-bridges-win64-git"`
[aurutils.log](https://github.com/AladW/aurutils/files/4405772/aurutils.log)

rafasc:
A shorter way to trigger this bug is simply: `aur query -t info </dev/null`

`aur sync` triggers this error via `aur-depends` that calls `aur-query` without input under certain conditions. See: https://github.com/AladW/aurutils/blob/master/lib/aur-depends#L46  
We either detect those cases in `depends` or make `query` gracefully exit on empty input.

AladW:
Fixed with #707

PR #707, opened 2020-04-03
Fix "warning: 'url' requires parameter"

rafasc:
Fix for #706 

Leaving this as a draft as I am still uncertain how the wget didn't have this issue.

### DIFF
--- a/lib/aur-query
+++ b/lib/aur-query
@@ -19,7 +19,8 @@ uri_info() {
         else if (NR > 1)
             printf "&arg[]=%s", $0
     } END {
-        printf "\n"
+        if (NR != 0)
+            printf "\n"
     }'
 }
 
@@ -83,6 +84,9 @@ EOF
 fi
 
 # pipeline
-jq -R -r '@uri' | uri | awk '{ printf "url %s\n", $0; }' | curl -K- -fgLsS
+readarray -t urls < <(jq -R -r '@uri' | uri )
+if (( ${#urls[@]} )); then
+    printf 'url %s\n' "${urls[@]}" | curl -K- -fgLsS
+fi
 
 # vim: set et sw=4 sts=4 ft=sh:
--- a/lib/aur-query
+++ b/lib/aur-query
@@ -5,10 +5,6 @@ argv0=query
 AUR_LOCATION=${AUR_LOCATION:-'https://aur.archlinux.org'}
 PS4='+(${BASH_SOURCE}:${LINENO}):${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
-# default options
-declare -i rpc_ver=5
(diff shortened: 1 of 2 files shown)


AladW:
> Leaving this as a draft as I am still uncertain how the wget didn't have this issue.
`wget` actually behaves sensible and exits on empty input (or a newline).

Looking at the code, I noticed some other things:
* Both `--rpc-ver` and `--rpc-url` have no effect because `rpc_url` is set before `rpc_ver` and `rpc_url`, respectively;
* `awk` is not needed in `uri_info`; a while/read loop with a counter does the same;

lib/aur-query
```diff
@@ -19,7 +19,8 @@ uri_info() {
         else if (NR > 1)
             printf "&arg[]=%s", $0
     } END {
-        printf "\n"
+        if (NR != 0)
```
AladW, line 22: `mapfile -t` strips the newline at the end so I guess this isn't needed. (Same if you use while/read and `[[ $foo ]] && bar+=("$foo")`

lib/aur-query
```diff
@@ -83,6 +84,9 @@ EOF
 fi
 
 # pipeline
-jq -R -r '@uri' | uri | awk '{ printf "url %s\n", $0; }' | curl -K- -fgLsS
+readarray -t urls < <(jq -R -r '@uri' | uri )
```
AladW, line 87: I went with a basic while/read approach but I guess this works too.

lib/aur-query
```diff
@@ -83,6 +84,9 @@ EOF
...
+if (( ${#urls[@]} )); then
+    printf 'url %s' "${urls[@]}" | curl -K- -fgLsS
```
AladW, line 89: Misses `\n` after `url %s`

test/issue-706
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+aur query -t info </dev/null
```
AladW, line 2: 🎉 for the test

lib/aur-query
rafasc: This was to address the fact that `aur depends` sometimes calls aur-query without input, (because the entries were already 'seen').

The `uri_info` takes this empty input and unconditionally adds a trailing `\n`. And although mapfile strips the `\n` the 'empty string' remains in the array. Which is then converted into `url<space><nothing><nl>` making curl complain about the lack of a url.

`aur depends carla-bridges-win32-git` (at preset date) and `aur query -tinfo </dev/null ` will trigger this corner case.

lib/aur-query
rafasc: Good catch, thanks. :bowing_man:

rafasc:
@AladW Feel free to force push your solution into this branch if you want.

AladW:
👯‍♂️🏖

PR #708, opened 2020-04-11
aur-build: use --noextract for --pkgver build

AladW:
The `prepare` function is run before the `pkgver` function and may thus influence its output.

Remove `--noprepare` from the preparatory `makepkg` command and use `--noextract` for the main makepkg command instead.


AladW:
:balloon:

PR #711, opened 2020-04-16
aur-sync: store commits reviewed by the user

AladW:
Store "seen" git commit hashes (git `HEAD` after `vifm` or `$AUR_PAGER` exited successfully) in the following directory structure:

```
$XDG_DATA_HOME/aurutils/view
|- pkgname1 # <sha_hash1>
|- pkgname2 # <sha_hash2>
|- ...
```
Compared to `AUR_SEEN` (#552) these hashes are stored even if `$AURDEST` is deleted. (#691)

This approach uses `aur fetch --results` to handle all diff functionality in `aur-sync`, instead of relying on `aur-fetch` to generate patches.

Diffs are only displayed if `HEAD` differs from an existing hash. On first run, hashes should be generated for packages in the local repository (#652), for example with:
```
$ aur sync $(aur repo --list | cut -f1) --print
<inspect files and exit with :q when satisfied>
```
----
Possible improvements:
* [x] Add `--rebuild-all` flag to simplify above command line (#712);
* [ ] Add subdirectories for "new" repositories (#379);

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -9,7 +9,7 @@ XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
 # default options
-chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0 results=0
+chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0
 
 # default arguments (empty)
 chroot_args=() pacconf_args=() repo_add_args=() makepkg_args=() makechrootpkg_makepkg_args=() makepkg_common_args=()
@@ -199,7 +199,6 @@ fi
 # Write successfully built packages to file (#437)
 if [[ -v results_file ]]; then
     : >"$results_file" # truncate file
-    results=1
 fi
 
 # Defaults to /usr/share/devtools/pacman-<prefix>.conf
@@ -312,7 +311,7 @@ while IFS= read -ru "$fd" path; do
     cd_safe "$db_root"
     LANG=C repo-add "${repo_add_args[@]}" "$db_path" "${pkglist[@]}"
 
-    if (( results )); then
+    if [[ -v results_file ]]; then
         printf "build:file://${db_path%/*}/%s\n" "${pkglist[@]}" | tee -a "$results_file"
     fi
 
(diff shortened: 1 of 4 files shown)


AladW:
With #712 the above command
```
aur sync $(aur repo --list | cut -f1) --print
```
can be simplified to
```
aur sync --rebuild-all --print
```

AladW:
Implementing #379 is more tricky. Personally I'm satisfied with the current "flat" layout of:
```
$tmp_view
|- foo.diff
|- bar.diff
|- foo/
|- bar/
```
but it marks no difference between packages that have not changed since the last review, and packages which have not yet been reviewed (i.e. without revision in `XDG_DATA_HOME/aurutils/view`) in the first place.

A number of proposed schemes have been proposed. The following list is how I recall them and with my perceived downsides:
1. Create `new`, `pkg` and `diff` subdirectories
  - Downsides: 
    * Dependencies can no longer be displayed in dependency order (if distributing packages between `new` and `pkg` without duplication) or are redundant (when `new` contains a subset of `pkg`)
    * Better suited for Miller column pagers like `ranger`, although `tree-view` in `vifm` could work.
2. Write `/dev/null` patches
  - Downside: 
    * Clunky with many unreviewed packages 
    * Patches may appear unexpectedly after exiting with `:cq`
3. Mark directories with special characters
  - Downside: 
    * Unclear which special character to choose

I have another approach in mind, which is however specific to `vifm`:
* Use a fixed "$tmp_view" directory, e.g. in `XDG_CACHE_HOME`;
* Use `vifm` bookmarks (`:bmark`, `:delbmarks`) to tag directories which were not reviewed.

Users of other pagers could either:
1. Check if directory names in `$XDG_DATA_HOME/aurutils/view` have a corresponding reference, or
2. rely on an environment variable (e.g. `AUR_PAGER_NEW`, `AUR_PAGER_PKG` with comma-seperated values) to sort packages.

As this is a more complicated/controversial point, perhaps it can be left for a later PR.

AladW:
Any feedback on the commits in this PR?

rafasc:
When cloning a package for the first time we get some fatal errors, that aren't really fatal.
```
fatal: not a git repository: 'rebuild-detector/.git'
Cloning into 'rebuild-detector/.git'...
```
`XDG_DATA_HOME` needs its default fallback, otherwise we get:

```
mkdir: cannot create directory ‘/aurutils’: Permission denied
```

After merging the recent PRs into my tree, I got some other permission denied errors, but I am unable to reproduce them. So I'll use this a few more days. Otherwise I think it is ok. :+1:

AladW:
I'll also report back when I get the chance to test this more extensively.

AladW:
@maximbaz @rafasc seems to work fine after cursory testing. Merge? ✌

PR #712, opened 2020-04-17
aur-sync: add --rebuild-all

AladW:
Flag to rebuild a local repository in full or repopulate `AURDEST` in conjunction with `--print`.

Provisions for `--provides` are not required, because `aur-repo-filter` has no output on empty input.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -67,7 +67,7 @@ trap_exit() {
 }
 
 usage() {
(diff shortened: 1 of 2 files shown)


AladW:
Renamed `--print` to `--no-build` as discussed on IRC. Maybe there could be a short option for it, like `-B` (which is already taken by `--build-command`).

Issue #709, opened 2020-04-14
What does the `--` do in aur-sync?


jwhendy:
I was going to pass `--skippgpcheck` to `aur sync` and found it wasn't a recognized option. I found that I was mis-remembering the `--` from `aur-build`. That said, what is the `--` for in the `aur-sync` documentation?

```
$ man aur-syncSYNOPSIS
       aur sync [-AcfgkLPstTuh] [--] pkgname...
# no -- mentioned after this
```

For comparison:
```
$ man aur-build
SYNOPSIS
       aur build [-d database] [-cfNRsv] [-a queue] [-B string] [--] [makepkg args]
# ...
OPTIONS
       All arguments after -- are passed to makepkg(8), or makechrootpkg  when  -c  is
       specified. The default options are --clean --syncdeps and -c -u, respectively.
```

I admit to not being a `bash` expert, but my quick check indicates that `--` is used to specify the end of options. I'd interpret that as things after `--` are interpreted as packages, but the error mentions `aur-depends`, so I wonder if this is meant to pass options to downstream `aur-*` commands intead?

```
aur-depends: invalid option '--skippgpcheck'
```

Thanks for any insights/clarification!

Earnestly:
The use of `--` is a POSIX convention for utilities that informs the option parser that anything which follows should not be considered as options.

> The argument -- should be accepted as a delimiter indicating the end of options. Any following arguments should be treated as operands, even if they begin with the '-' character. The -- argument should not be used as an option or as an operand.

https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html

jwhendy:
That was my read, but it wasn't clear to me how `--` would be used in the context of `aur-sync`. What would go between `--` and `[pkgname]`?

> Using -- to separate arguments to different utilities, as xinit does, is always mistake.

Does that apply to `aur build [options] [--] [makgpkg args]`, or is that consuming those arguments in a different, more acceptable manner?

I mainly posted here as if the `--` isn't doing anything for `aur-sync`, is it necessary in the docs? Since I'd seen it `aur-build`, I spent time looking. If it does nothing and `--` is always available to do it's thing, perhaps it can be removed from the docs to avoid searching for the meaning.

Earnestly:
It does apply; I believe `aur sync` was fixed recently.

AladW:
There's a missing `--` in https://github.com/AladW/aurutils/blob/master/lib/aur-sync#L254
```
[...]

PR #713, opened 2020-04-20
aur-build: add --margs

AladW:
Replace `--build-command` by the following:

1. Add options to the `makepkg` command-line (both on the host and on the container, depending if -c is specified) with the `--margs` option. Arguments are comma-separated and may be specified multiple times. 
  To avoid quoting issues with `makepkg --config` through `--margs`, `--makepkg-conf` now sets the makepkg configuration used with `makepkg` for host (non-container) builds.
2. Support the `MAKEPKG` and `PACMAN` environment variables, to set the commands for `pacman` and `makepkg` operations.

Revisit of #678 (@bartoszek) and #709. A test case should include `aur sync --margs -h`.

### DIFF
--- a/THANKS
+++ b/THANKS
@@ -1,6 +1,6 @@
 Andrew Gregory <https://github.com/andrewgregory>
 Assaf Gordon <http://savannah.gnu.org/users/agn>
-Robin Broda <https://github.com/coderobe>
+bartoszek <https://github.com/bartoszek>
 Damien Robert <https://github.com/DamienRobert>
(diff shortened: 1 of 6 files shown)


AladW:
Looks like this doesn't work just yet... documentation for `--makepkg-conf` and `--pacman-conf` is also missing.

AladW:
I've left out the `MAKEPKG` and `PACMAN` commit as I would like to eventually use `run_pacman` from makepkg, to handle locks left by `pacman` (concurrent `aur-build` processes).

PR #717, opened 2020-05-08
completion: fix broken completion for aur repo

rafasc:
When we attempt to generate a completion we check if the previous 'word'
matches 'word:' because options which require arguments are marked with
an ':' suffix in the array.

If we find a match, we complete with files by default,
otherwise we show all available options for the subcommand being used.

This worked until aur-repo learned '--repo:'.  Making aur repo <tab>
complete with files, as the subcommand itself matched one of its
options with required arguments.

To prevent that, always complete with options if we are on the second
word.

### DIFF
--- a/completions/bash/aurutils.in
+++ b/completions/bash/aurutils.in
@@ -30,7 +30,7 @@ cat <<'EOF'
     # Fallback to default (files) completion in such cases.
 
     opts=(${default_cmds[${words[1]}]})
-    if [[ ${opts[*]} != *$prev:* ]]; then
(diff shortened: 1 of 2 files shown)


AladW:
Thanks. I guess in an ideal world we could have `--repo` complete actual repositories (with `pacconf` e.g.), but I already have trouble understanding basic completion issues. Unless there's any other ideas let's merge this.

rafasc:
Shouldn't be hard to add. The bash version already has dispatch function to allow making custom completions instead of using the dynamically generated.

But that's content for another PR.

PR #720, opened 2020-05-10
aur-build: pass on --syncdeps to pkgver makepkg

AladW:
Fixes #716

PR #723, opened 2020-05-31
Use makepkg config when updating pkgver

aruhier:
Sometimes packages check the compiler and its options. It can fail
during the pkgver update if the default makepkg.conf (/etc/makepkg.conf)
has been modified.
This commit forces makepkg to use the given makepkg_conf instead.

Example: package checking that the compiler is gcc and default makepkg
uses clang.


AladW:
Part/duplicate of #720 

Please check if that one works as intended.

aruhier:
Indeed, I didn't see this one.

Thanks!

PR #725, opened 2020-07-03
Add option for incrementing pkgrel

zoqaeski:
This allows packages to be rebuilt and detected as upgrades by `pacman -Syu`. It updates the `pkgrel` variable using an awk script.

I'm not sure whether a check for `pkgver` as a conflicting option should also be added, as it seems that running `makepkg --nobuild --nodeps` resets the `pkgrel` back to 1 if the version changes. To play it safe, these modifications update the `pkgrel` first.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -9,7 +9,7 @@ XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
 # default options
-chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0 results=0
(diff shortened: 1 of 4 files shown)


AladW:
Somewhere in 2016 I toyed with a similar idea, using setconf(1) which 
was made purposely for this rather than an awk script. I noticed the 
following problems:

* It's not clear what format versioning bumps should have, as +1 is used 
by AUR maintainers for packaging changes. +0.1 is something Arch 
spinoffs tend to do, but nothing prevents AUR maintainers from doing the 
same. Special characters like ~ also have no effect here (unlike 
Debian/dpkg).

* pkgrel can contain shell variables which are only populated at 
runtime, as long as it evaluates to an integer. This is usually in 
exceptionally bad PKGBUILDs, so less of an issue.

If you control pkgrel, i.e. in PKGBUILDs you maintain yourself (or are 
reasonably confident on the pkgrel scheme used by the AUR maintainer), 
it's simple to create an aur(1) script that bumps pkgrel before calling 
aur-build. I use this approach in combination with rebuild-detector. But 
for the reasons above, it's not general enough to make it part of 
aur-build itself.

PR #726, opened 2020-07-04
Create Dockerfile

AladW:
Useful for local testing outside of Arch.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,15 +1,16 @@
 FROM archlinux/base:latest
-LABEL Alad Wenter <alad@archlinux.org>
+LABEL Name=aurutils
 
-RUN pacman -Syu --noconfirm base base-devel
-RUN pacman -Syu --noconfirm jq pacutils git wget shellcheck vim vifm devtools bash-completion parallel
+# packages, time zone
+RUN pacman -Syu --noconfirm --needed base base-devel
+RUN pacman -Syu --noconfirm jq pacutils git wget shellcheck vim vifm devtools bash-completion man-db man-pages
 RUN ln -fs /usr/share/zoneinfo/Europe/Berlin /etc/localtime
 
 # unprivileged user
 ARG USERNAME=archie
 RUN useradd -m -s /bin/bash $USERNAME
-RUN install -o $USERNAME -dm755 /home/$USERNAME/.local
-RUN install -o $USERNAME -dm755 /home/$USERNAME/.local/bin
+RUN install -o $USERNAME -Dm755 /home/$USERNAME/.local && \
+    install -o $USERNAME -dm755 /home/$USERNAME/.local/bin
 ENV PATH "/home/${USERNAME}/.local/bin:$PATH"
 
 # locales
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+
+RUN pacman -Syu --noconfirm base base-devel
+RUN pacman -Syu --noconfirm jq pacutils git wget shellcheck vim vifm devtools bash-completion parallel
+RUN ln -fs /usr/share/zoneinfo/Europe/Berlin /etc/localtime
+
+# unprivileged user
+ARG USERNAME=archie
+RUN useradd -m -s /bin/bash $USERNAME
+RUN install -o $USERNAME -dm755 /home/$USERNAME/.local
+RUN install -o $USERNAME -dm755 /home/$USERNAME/.local/bin
```
maximbaz, line 12: Every `RUN` creates a new image layer, I would suggest optimizing stuff a little bit, e.g. merging these two `RUN` into one, and above two `pacman` installations - don't go crazy trying to minimize the number of layers, but in this case I just think there is no value to create two layers, one without `bin` folder yet created and another one with.

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM archlinux/base:latest
+LABEL Alad Wenter <alad@archlinux.org>
+
+RUN pacman -Syu --noconfirm base base-devel
```
maximbaz, line 4: I would add `--needed` flag as well, in case some of the stuff is already part of `archlinux/base`

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN pacman -Syu --noconfirm jq pacutils git wget shellcheck vim vifm devtools bash-completion parallel
```
maximbaz, line 5: `parallel` is not needed anymore?

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# locales
+RUN echo en_US.UTF-8 UTF-8 > /etc/locale.gen && locale-gen
+ENV LANG en_US.UTF-8
+
+# sudoers
+RUN echo $USERNAME ALL=\(root\) NOPASSWD: /usr/bin/pacman > /etc/sudoers.d/$USERNAME-pacman \
+    && chmod 0440 /etc/sudoers.d/$USERNAME-pacman
+
+# man pages
+RUN pacman -Syu --noconfirm man-db man-pages
```
maximbaz, line 24: Would also suggest merging this one with the `pacman` call above

docker/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN install -o $USERNAME -dm755 /home/custompkgs
+RUN sudo -u $USERNAME repo-add $REPO_DB
+
+ARG PACMAN_CONF=/etc/pacman.conf
+RUN perl -0777 -pi -e 's/#(\[custom\])\n#(SigLevel.*)\n#(Server.*)\n/$1\n$2\n$3\n/' $PACMAN_CONF
+RUN pacsync custom
+
+# aurutils package
+RUN cd /home/archie \
+    && sudo -u archie git clone https://github.com/AladW/aurutils \
```
maximbaz, line 37: Is your plan for this Dockerfile to test that master branch of `aurutils` can be built, or to assist with development? For the former what you made is all good, for the latter maybe `aurutils` need to be a volume instead (so that a person can mount _their_ code to test, maybe before submitting a PR), and as a last line add something like `CMD ["/bin/bash"]` so that users can test the resulting `aur` binary?

AladW: The volume idea is a good one, I'll look into it 🙂

AladW:
Updated 🙂

PR #729, opened 2020-07-18
aur-repo: do not try to extract an empty database

### LINKED ISSUE #727: aur-repo: bsdtar errors on empty repository
Simple way to reproduce:
```
aur sync -d custom <package not in custom>
```
```
bsdtar: */desc: Not found in archive
bsdtar: Error exit delayed from previous errors.
```

AladW:
Fixes #727

### DIFF
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -8,7 +8,8 @@ PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 modifier=local table_format=0
 
 db_table() {
-    bsdtar -Oxf "$2" '*/desc' | awk -v table="$1" '
+    # do not extract an empty database (#727)
+    (( $(bsdtar -tf "$2" | wc -l) )) && bsdtar -Oxf "$2" '*/desc' | awk -v table="$1" '
     function print_previous() {
         table ? format = "%s\t%s\t%s\t%s\n" : format = "%1$s\t%4$s\n"
 


AladW:
@rafasc `bsdcat` was used before, do you remember the reason for using `bsdtar -Oxf`?

maximbaz:
I guess to be able to pass the pattern, i.e. `'*/desc'` in `bsdtar -Oxf "$2" '*/desc'`?

rafasc:
> I guess to be able to pass the pattern, i.e. `'*/desc'` in `bsdtar -Oxf "$2" '*/desc'`?

I wish I explained it on the commit message.

`bsdcat` includes binary "junk" that is not interesting for us, although not very noticable when directly used on the terminal.
see: `bsdcat /var/lib/pacman/sync/core.db | less`

Using `bsdtar` allows us to get only the contents of `desc` files.

lib/aur-repo
```diff
@@ -8,6 +8,9 @@ PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 modifier=local table_format=0
 
 db_table() {
+    # do not extract an empty database (#727)
+    (( $(bsdtar -tf "$2" | wc -l) )) || return
+
     bsdtar -Oxf "$2" '*/desc' | awk -v table="$1" '
```
rafasc, line 14: Maybe we should put the output in a tempfile and only pass it to awk if it is non-empty. 

The advantage being that in case of failure we would at keep the input that caused the error.

Earnestly: awk can be used to work with both empty and non-empty inputs

AladW: Yeah, the issue here is with `bsdtar` erroring out if the archive is empty; awk would work regardless.

AladW:
I'll merge this to remove as it blocks 3.0.0, I'll happily look at another PR with a different/better approach.

PR #733, opened 2020-08-18
Make sure db_table exits successfully on empty db

maximbaz:
Follow-up to #729, seems like on an empty DB even though it is not being extracted, the function still doesn't return successfully, and so you cannot build your first package. Having `return 0` is a simple way to ensure the function always succeeds.

### DIFF
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -56,6 +56,7 @@ db_table() {
     END {
         print_previous()
     }'
+    return 0
 }
 
 usage() {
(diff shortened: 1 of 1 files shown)


AladW:
Maybe it's more clear to put the `return 0` at the top, in the sense of:
```
(( $(bsdtar -tf "$2" | wc -l) )) && return 0

bsdtar ...
```

AladW:
Thanks!

AladW (approved):

PR #743, opened 2020-10-03
aur-build: extend --user to local builds

AladW:
This allows in particular to build packages with one user, and GPG sign them with another. This takes advantage of `aur-build` signing packages separately from `makepkg` in `/var/tmp`. `gpg` and other commands such as `repo-add` are unaffected by this flag and run as the local user.

To use `--syncdeps` with the build user, an appropriate sudoers rule should be added. This is similar to the existing `sudo -u build` approach documented in `aur-build`. If desired, `--syncdeps` can be implemented separately by the user, e.g. to allow an `sudo aur-build` invocation with dependency resolution.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -71,7 +71,7 @@ if ! parseopts "$opt_short" "${opt_long[@]}" "${opt_hidden[@]}" -- "$@"; then
 fi
 set -- "${OPTRET[@]}"
 
-unset db_name db_path db_root makepkg_conf results_file queue
+unset db_name db_path db_root makepkg_conf makepkg_user results_file queue
 while true; do
     case "$1" in
         # build options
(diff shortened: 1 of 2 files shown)


lib/aur-build
```diff
@@ -201,6 +202,14 @@ if [[ -v results_file ]]; then
     : >"$results_file" # truncate file
 fi
 
+if [[ -v makepkg_user ]]; then
+    id -- "$makepkg_user" >/dev/null
+
+    makepkg() {
+        command sudo -u "$makepkg_user" makepkg "$@"
```
maximbaz, line 209: I don't know if it's possible that this line becomes recursive, but if you are not sure either, maybe it would be safer to resolve the full path to `makepkg`, or just hardcode `/usr/bin/` ?

maximbaz (approved):

AladW:
It's fine because makepkg() is a function which is not exported to sudo. 
Even the "command" here is not required but I added it for clarity.

lib/aur-build
AladW: It's fine because `makepkg()` is a function not exported to `sudo`. Even the `command` here is not required but I added it for clarity.

AladW:
As I figured out you can already do this with `--build-command` - just create a wrapper script that runs `sudo -u <user> makepkg`:
```
#!/bin/sh --
user=$1
shift
sudo -u "$user" makepkg "$@"
```

PR #748, opened 2020-10-06
aur-build: do not print results to stdout

AladW:
As https://github.com/AladW/aurutils/commit/306ff632557d37dc70125b98d24ea45bdbb80517 but for `aur-build`.

Also remove some redundancy for the "Running..." messages.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -351,7 +351,7 @@ while IFS= read -ru "$fd" path; do
     LANG=C repo-add "${repo_add_args[@]}" "$db_path" "${pkglist[@]}"
 
     if [[ -v results_file ]]; then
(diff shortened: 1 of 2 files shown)


lib/aur-build
```diff
@@ -26,6 +26,11 @@ db_replaces() {
     }'
 }
 
+run_msg() {
+    printf 'Running %s\n' >&2 "$*"
```
maximbaz, line 30: This is a bit unexpected order of arguments for my taste, personally I would prefer to put `>&2` either right after `printf`, or before everything, or after everything, but not between arguments to `printf` 🙂

maximbaz (approved):

AladW:
🚀

PR #749, opened 2020-10-06
aur-build: propagate --pacman-conf to makepkg for local builds

AladW:
Noticed this as my package tests would not use their custom set cache directory: `makepkg -s` runs `pacman` with a few hardcoded arguments; in particular, `--config` is not supported. The only way around this is either use a chroot, or set `PACMAN` to a `pacman --config` wrapper. As former is already possible, this PR implements latter.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -4,6 +4,7 @@
 set -o errexit
 shopt -s extglob
 argv0=build
+runtime_dir=${XDG_RUNTIME_DIR:-/run/user/$UID}/aurutils
 startdir=$PWD
 XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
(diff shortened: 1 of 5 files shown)


lib/aur-build
```diff
@@ -222,11 +225,24 @@ if (( chroot )); then
     printf 'Running aur-chroot %s\n' >&2 "${args[*]}"
     aur chroot "${args[@]}"
 else
+    # makepkg -s does not support setting pacman.conf (#)
```
maximbaz, line 228: Did you want to link to a github issue here?

maximbaz (approved):
Looks good, just one typo

lib/aur-build
AladW: I thought of filing an issue first, but I went straight to the PR. Removed the #

AladW:
Thanks 👍

PR #751, opened 2020-10-10
tests: add chroot, issue 741, 378, 216

AladW:
@maximbaz `package-test-chroot` doesn't quite seem to work, perhaps you could take a look? My small VM doesn't have the space for an arch container and its copy...


maximbaz:
> @maximbaz package-test-chroot doesn't quite seem to work, perhaps you could take a look?

I pushed the fix, it took quite a while on my laptop, but the script succeeded in the end 😉

maximbaz (approved):

PR #754, opened 2020-10-14
aur-build: support MAKEPKG environment variable

AladW:
Also includes a fix for supporting `PACMAN` with `--pacman-conf`.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -128,7 +128,6 @@ while true; do
             makepkg_common_args+=(--rmdeps) ;;
         -s|--syncdeps)
             makepkg_common_args+=(--syncdeps) ;;
-        # makepkg options (config)
         --makepkg-conf)
             shift; makepkg_conf+=(--config "$1")
             chroot_args+=(--makepkg-conf "$1") ;;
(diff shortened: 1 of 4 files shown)


lib/aur-build
```diff
@@ -203,6 +202,11 @@ elif ! [[ -w $db_path ]]; then
     exit 13
 fi
 
+# Custom makepkg command
+if [[ ${MAKEPKG-} ]]; then
```
maximbaz, line 206: although this seems to work, is it the same as using `${MAKEPKG:-}`, and if so, lets maybe use this syntax instead as it is more common? unless `${MAKEPKG-}` has a different behavior, which I can't find documentation for...

maximbaz (approved):
👍

lib/aur-build
AladW: The difference between `-` and `:-` is that the value following it would be substituted if `$MAKEPKG`is empty, or empty and unset, respectively. It doesn't really matter in this case, I only added the `-` if the script is run with `bash -u` and we could leave it out entirely.

maximbaz: Ah I see, interesting - I don't mind either way, leave whatever you prefer, just interesting to know there is an actual difference 👍

PR #755, opened 2020-10-14
aur-build: use libmakepkg for packagelist

AladW:
Basic performance comparison on a low-performance VM:
```
[archie@vbox makepkg]$ time makepkg --packagelist
/home/archie/projects/aurutils/makepkg/aurutils-git-3.0.0rc.r7.gc91de35-1-any.pkg.tar.zst
    
real    0m2.412s
user    0m1.766s
sys     0m0.580s

[archie@vbox makepkg]$ time ../makepkg-fast --packagelist
/home/archie/projects/aurutils/makepkg/aurutils-git-3.0.0rc.r7.gc91de35-1-any.pkg.tar.zst
    
real    0m0.064s
user    0m0.051s
sys     0m0.008s
```

Where `makepkg-fast` is a script using libmakepkg/util/pkgbuild.sh print_all_package_names() directly if called with --packagelist.

---

A possible issue is that the PKGBUILD is not linted if the build is skipped due to matching version in the local repository. In this specific case, `lint_pkgbuild` could be called (which apparently is a cause for the slow-down above).

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -47,9 +47,12 @@ usage() {
     exit 1
 }
 
-source /usr/share/makepkg/util/util.sh
+source /usr/share/makepkg/util/config.sh
 source /usr/share/makepkg/util/message.sh
+source /usr/share/makepkg/util/option.sh
 source /usr/share/makepkg/util/parseopts.sh
+source /usr/share/makepkg/util/pkgbuild.sh
+source /usr/share/makepkg/util/util.sh
 
 if [[ ! -v NO_COLOR ]] && [[ ! -v AUR_DEBUG ]]; then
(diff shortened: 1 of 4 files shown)


maximbaz (approved):
Cool!

AladW:
I've been thinking on this some more, and the only scenario I can see regarding unlinted PKGBUILDs is (practically speaking) when a package was built with an older version of makepkg, and the PKGBUILD is no longer valid with a newer version of makepkg. In that case, the version from `print_all_package_names` would still be correct; the package would just need adaptations to build again with the newer makepkg version.

Maybe there's other possible scenarios, but for now I think it's a reasonable idea.

PR #756, opened 2020-10-14
aur-chroot: use --update/--build

AladW:
Enable functionality with `--update`/`-U` (create or update chroot) and `--build`/`-B` (build package inside chroot), instead of disable it with `--no-prepare` and `--no-build`, respectively.

---
I always found the selective disabling of functionality in `aur-chroot` awkward, and the full program can be enabled by golfing `-UB`.


maximbaz (approved):
Makes sense 👍

PR #763, opened 2020-10-25
aur-query: add AUR_QUERY_RPC, AUR_QUERY_RPC_SPLITNO

AladW:
This replaces the `--rpc-ver` and `--rpc-url` options (to allow easy propagation to all utils using `aur-query`), and allows to set the amount of packages before a `GET` request is split to multiple URIs, respectively.


maximbaz (approved):
Makes sense

PR #764, opened 2020-10-25
aur-build, aur-chroot: rename --prefix to --suffix

AladW:
This option affects the suffix `extra` in `pacman-extra.conf`, and should thus be called `--suffix`.


maximbaz (approved):
haha good catch 🙃

PR #767, opened 2020-10-26
aur-sync: use XDG_RUNTIME_DIR for view directory

AladW:
With miller column file managers (in particular, ranger), saving the directory with symlinked build files in `TMPDIR` caused a lot of irrelevant files to appear in the left-most column. Using `XDG_DATA_HOME` allows to use this left-most column to switch to viewed revisions in `XDG_DATA_HOME/aurutils/view`.

To avoid conflicts, the parent process ID is suffixed to the directory name.

Later commits may move this directory (and possibly additional symlinks to directories in `XDG_DATA_HOME/aurutils/view`) to the more suitable `XDG_RUNTIME_DIR`.

Before:
Before

After:
After

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -6,6 +6,7 @@ argv0=sync
 XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
 XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 XDG_DATA_HOME=${XDG_DATA_HOME:-$HOME/.local/share}
+XDG_RUNTIME_DIR=${XDG_RUNTIME_DIR:-/run/user/$UID}
 AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/$argv0}
 PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
@@ -205,9 +206,11 @@ while true; do
     shift
 done
 
-tmp=$(mktemp -d --tmpdir "aurutils-$argv0.XXXXXXXX") || exit
-tmp_view=$(mktemp -d --tmpdir "aurutils-$argv0-view.XXXXXXXX") || exit
+# Directory for package symlinks
+mkdir -p "$XDG_RUNTIME_DIR"/aurutils
 
+tmp=$(mktemp -d --tmpdir "aurutils-$argv0.XXXXXXXX") || exit
+tmp_view=$(mktemp -d "$XDG_RUNTIME_DIR/aurutils/view.XXXXXXXX") || exit
 trap 'trap_exit' EXIT
 
 # Directory for git checksums (revisions viewed by the user, #379)


AladW:
@FichteFoll

maximbaz:
I can't say I'm super fond of this PR... 🤔 Even if you move the directory, whatever is on the left most column isn't any more _relevant_ to your task of reviewing diffs than some garbage in `/tmp`... And you are sacrificing some out-of-the-box functionality which forces you to think about extra things, such as suffixing folder with process ID or remembering to have a trap that cleans up the directory on exit... 

What if you just nest it deeper in `/tmp` and keep the rest as it is?

```bash
mkdir -p /tmp/aurutils
tmp_view=$(mktemp -d '/tmp/aurutils/aurutils-$argv0-view.XXXXXXXX')
```

FichteFoll:
I personally wasn't bothered by the left column showing things I didn't care about until you mentioned it yesterday. Probably because I considered the left column to be of no interest to me due to being used to miller columns. Either way, I have since tried out 2 column mode for reviewing aur packages and that turned out quite nicely. Here's a screenshot of how that looks after launching `aur sync …`:

2020-10-26_20-27-03_aurutils_ranger

The core part of that is `AUR_PAGER="ranger --command aur"` with the following in `rc.conf`:
```
alias aur chain aur_mark_new; flat 1; scout -vpr ^\.SRCINFO$|\.(asc|bz2|deb|gitignore|gz|log|rpm|sig|tgz|xz)$; set column_ratios 2,5
```
`aur_mark_new` is a custom command that I already mentioned in https://github.com/AladW/aurutils/issues/379#issuecomment-711426955

I believe this is a better solution for the problem and it doesn't matter where the directory is located for that.

AladW:
```
I believe this is a better solution for the problem and it doesn't matter where the directory is located for that.
```
Looks great!

```
What if you just nest it deeper in /tmp and keep the rest as it is?
```

In that case I would stuff it to `XDG_RUNTIME_DIR/aurutils` as I planned to originally, with no extra directories. In the end these are files the user interacts with on runtime, so I think it makes more sense than `/tmp`.

lib/aur-sync
```diff
@@ -205,9 +206,10 @@ while true; do
     shift
 done
 
-tmp=$(mktemp -d --tmpdir "aurutils-$argv0.XXXXXXXX") || exit
-tmp_view=$(mktemp -d --tmpdir "aurutils-$argv0-view.XXXXXXXX") || exit
+mkdir -p "$XDG_RUNTIME_DIR"
 
+tmp=$(mktemp -d --tmpdir "aurutils-$argv0.XXXXXXXX") || exit
+tmp_view=$(mktemp -d "$XDG_RUNTIME_DIR/aurutils/view.XXXXXXXX") || exit
```
maximbaz, line 212: I don't think this will work, you need to `mkdir -p $XDG_RUNTIME_DIR/aurutils` above or maybe just use `aurutils-view.XXXXXXXX` instead of subfolders `aurutils/view.XXXXXXXX`, depending if you still want to "hide" the folders from a higher level:

```
$ echo $XDG_RUNTIME_DIR
/run/user/1000

$ mktemp -d "$XDG_RUNTIME_DIR/aurutils/view.XXXXXXXX"
mktemp: failed to create directory via template ‘/run/user/1000/aurutils/view.XXXXXXXX’: No such file or directory
```

AladW: Yeah I did `mkdir -p $XDG_RUNTIME_DIR/aurutils` in the end 🙂

maximbaz (approved):

PR #781, opened 2020-11-02
aur-chroot: add --create

AladW:
Yet another rewrite of `aur-chroot`... 

* Set the default `pacman.conf` for `aur-build --chroot` to `/etc/aurutils/pacman-<db>.conf`. To accommodate this, the `Makefile` creates `$(PREFIX)/etc/aurutils`. Users will no longer have to write their config in `/usr/share` and can use a `pacman.conf` per local repository with no additional effort (apart from specifying the repository with `-d` or `AUR_REPO`).

* Create containers with `aur chroot --create`, and allow to specify the target packages on the command-line. This way all required packages can be specified in one step, e.g. to include `ccache` in the container. This additional step is only required when using `aur chroot` directly, or if other packages than `base-devel` or `multilib-devel` are required; `aur build` runs `aur chroot --create` on each `--chroot` build. (This is possible because `aur-chroot` only runs `mkarchroot` if there is no existing container in the target path, which still defaults to `/var/lib/aurbuild/<arch>`.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ PREFIX ?= /usr
 SHRDIR ?= $(PREFIX)/share
 BINDIR ?= $(PREFIX)/bin
 LIBDIR ?= $(PREFIX)/lib
+ETCDIR ?= /etc
 AURUTILS_LIB_DIR ?= $(LIBDIR)/$(PROGNM)
 
 .PHONY: shellcheck install build completion aur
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -23,6 +24,7 @@ install-aur: aur
 
 install: install-aur
 	@install -Dm755 lib/aur-* -t '$(DESTDIR)$(LIBDIR)/$(PROGNM)'
+	@install -dm755 aurutils  -t '$(DESTDIR)$(ETCDIR)'
```
maximbaz, line 27: `install: target directory not allowed when installing a directory`

Must be `@install -dm755 '$(DESTDIR)$(ETCDIR)/aurutils'` it seems

Makefile
```diff
@@ -3,6 +3,7 @@ PREFIX ?= /usr
 SHRDIR ?= $(PREFIX)/share
 BINDIR ?= $(PREFIX)/bin
 LIBDIR ?= $(PREFIX)/lib
+ETCDIR ?= $(PREFIX)/etc
```
maximbaz, line 6: remove `$(PREFIX)`, otherwise this is `/usr/etc/aurutils`

maximbaz (approved):
Besides these comments, seems to work well 👍 Prepare to confuse people yet again about the config being now in a different location 😁

AladW:
Thanks!

PR #786, opened 2020-11-07
aur-chroot: add base-devel to multilib container

codicodi:
[`multilib-devel`](https://www.archlinux.org/groups/x86_64/multilib-devel/) group is pretty small and does not constitute a usable container. It is intended to be installed on top of regular `base-devel`. Relevant devtools code: https://github.com/archlinux/devtools/blob/20200407/archbuild.in#L7-L14


AladW:
Thanks!

AladW (approved):

PR #788, opened 2020-11-09
aur: add --version output

AladW:
If wanting to support different `aurutils` versions in the same script, having the version directly from `aur` can be useful. (e.g. if aurutils is installed manually through the `Makefile`)


Makefile
```diff
@@ -5,13 +5,15 @@ BINDIR ?= $(PREFIX)/bin
 LIBDIR ?= $(PREFIX)/lib
 ETCDIR ?= /etc
 AURUTILS_LIB_DIR ?= $(LIBDIR)/$(PROGNM)
+AURUTILS_GIT_VER := $(shell git describe --tags)
```
maximbaz, line 8: This won't work when building from a tarball, right...

Maybe make it `?=` and in PKGBUILD pass `pkgver` as `make` parameter?

AladW: Good point

AladW: https://github.com/AladW/aurutils/pull/788/commits/de351889d62feb30cb2f978b52415448cc1e3097 adds a fallback that doesn't require changes in the PKGBUILD. Thoughts?

maximbaz: This I'm fine with, the small downside comparing to passing pkgver in PKGBUILD is that now you need to remember to adjust version before each next release 😉

maximbaz (approved):

Makefile
AladW: Sure (and I'll probably forget to update it at some point...) but on the upside people installing from a tarball directly will also have a functional `aur --version` by only running `make` 🙂

AladW:
Thanks for the quick review, I think I'll release `3.1.1` right away 🎉

PR #789, opened 2020-11-09
aur: fix substitution of AUR_LIB_DIR

AladW:
(눈_ლ)


makepkg/PKGBUILD
```diff
@@ -25,7 +25,7 @@ pkgver() {
 
 build() {
     cd aurutils
-    make DESTDIR="$pkgdir"
+    make
```
maximbaz, line 28: intentional?

maximbaz (approved):

makepkg/PKGBUILD
AladW: Yeah, it's the same in the AUR PKGBUILD - `DESTDIR` does nothing in the build step.

PR #790, opened 2020-11-09
aur-sync: add --provides-deep

### LINKED ISSUE #592: `sync`: `--provides` fails to filter out unneeded transitive (make) dependency
EDIT: Using `master` branch.

With `mongodb-bin` available,

```bash
$ aur sync unifi --provides
==> Using [aur] repository
  -> mongodb: (none) -> 4.0.12-1
  -> python2-scons: (none) -> 3.1.1-1
  -> unifi: 5.10.25-1 -> 5.10.26-1
    dependency mongodb satisfied by mongodb-bin
```
still tries to install `python2-scons`, which should not be needed if `mongodb-bin` is fulfilling the `mongodb` dependency of `unifi`.

If a dependency is satisfied via `--provides`, I would expect the leaves of that dependency tree to also be pruned. In this case, it seems like only `mongodb` is being pruned, but not its (make) dependency `python2-scons`.

[aurlog.txt](https://github.com/AladW/aurutils/files/3489231/aurlog.txt)

AladW:
Allows to take transitive dependencies into account.

Fixes #592

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -6,7 +6,6 @@ argv0=sync
 XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
 XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 XDG_DATA_HOME=${XDG_DATA_HOME:-$HOME/.local/share}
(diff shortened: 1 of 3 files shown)


maximbaz (approved):
Haven't tested but the code looks fine 🙂

PR #792, opened 2020-11-14
aur-repo: set repository extension with AUR_DBEXT

### LINKED ISSUE #700: Issues if --root points to a file system which doesn't support sym links
I came across the following issue: When you specify the  path to the repository via the `--root` option and it points  to a mounted file system which doesn't support sym links you run into the following error

```
==> ERROR: '/tmp/tmp.r2xrUgTpZk/repo.db' does not have a valid database archive extension.
```

The cause of this is that the `.db` extension is hard coded within aurutils. Therefore it is not possible to specify the database name with `repo.db.tar.gz`.

https://github.com/AladW/aurutils/blob/d3e8fa5b80da2e4df9c49f00e7634142cc0dcf0c/lib/aur-build#L183

This is okay if there is a symlink from `repo.db` to `repo.db.tar.gz`. But a samba share for example doesn't support sym links.

My suggestion would be to only append the extension if the specified database name has none. In any other case leave it as it is.

AladW:
1. Fix a bug with `--results` where the results file would be written to $PWD/package instead of $PWD, when using relative paths.

2. Fix #700 by not hardcoding `.db` in `aur-sync`, `aur-build` and `aur-repo`.

3. Add `aur repo --status` for repository selection in `aur-build`. This also generalizes `aur repo --path` (now removed).

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -12,7 +12,7 @@ PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 chroot=0 no_sync=0 overwrite=0 sign_pkg=0 run_pkgver=0
 
 # default arguments (empty)
-chroot_args=() pacconf_args=() repo_add_args=() makepkg_args=() makechrootpkg_makepkg_args=() makepkg_common_args=()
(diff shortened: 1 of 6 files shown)


maximbaz (approved):
👍

PR #815, opened 2021-05-28
CI: Change image from archlinux/base to archlinux

xenrox:
Fixes broken CI.


AladW:
Yeah, that was a false positive from shellcheck. Should all work now.

PR #819, opened 2021-06-17
Losslessly optimize mascot image with optipng -o9

98WuG:
[OptiPNG is a PNG optimizer that recompresses image files to a smaller size, without losing any information.](http://optipng.sourceforge.net/)
```
$ optipng -o9 06Nitori1.png
[...]
Output file size = 318311 bytes (62362 bytes = 16.38% decrease)
```


AladW (approved):

AladW:
Nice, thanks.

PR #822, opened 2021-07-12
use the proper xdg cache path for the order file instead of config path, since it's a temporary file

BrainDamage:
use the proper xdg cache path for the order file instead of config path, since it's a temporary file

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -243,7 +243,7 @@ view_db=$XDG_DATA_HOME/aurutils/view
 mkdir -p "$view_db"
 
 # Default to showing PKGBUILD first in patch. (#399)
-orderfile=$XDG_CONFIG_HOME/aurutils/$argv0/orderfile
+orderfile=$XDG_CACHE_HOME/aurutils/$argv0/orderfile
 mkdir -p "${orderfile%/*}"
 
 if [[ ! -s $orderfile ]]; then


Earnestly:
I wonder how far `XDG_STATE_HOME` applies to files like this.

BrainDamage:
I wasn't aware of it, it could be a good fit as well.
Personally I'd be fine with either state/cache/data since those at least clearly are paths meant to be written by the program runtime instead of more permanent configuraion

AladW:
When I chose XDG_CONFIG_HOME, what I had in mind was people modifying orderfile so they could choose what stuff they want to prioritize in the diffs. So they're written by the program runtime as a default value, rather as something that's cached. If there's ambiguity in that, maybe something different should be done.

On Tue, 13 Jul 2021 13:06:50 -0700
BrainDamage ***@***.***> wrote:

> I wasn't aware of it, it could be a good fit as well.
> Personally I'd be fine with either state/cache/data since those at least clearly are paths meant to be written by the program runtime instead of more permanent configuraion
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/AladW/aurutils/pull/822#issuecomment-879366449


-- 
alad ***@***.***>

AladW:
Since there's been no new opinions, I think this can be closed.

PR #848, opened 2021-11-05
lib/aur-build: Queue existing packages for adding to the repo

CyberShadow:
Fixup for b6e2db4d90830686d8d220b939553044dc9573ed

Sorry, I could only actually test that patch today, looks like it was missing a tiny addition.


AladW:
Cheers

Issue #860, opened 2021-11-25
Support or documentation for non-AUR packages


dhouck:
[Suggested label: question and/or enhancement]

This is a great tool for building Arch packages and putting them in a repository for Pacman.  Despite the name, this is useful for more than just the AUR; is there a way to use most of this same infrastructure for non-AUR packages?  As far as I can tell, there is not, but primarily because `aur-depends` is called unconditionally by `aur-sync` and I have to drop down to `aur-build`, and do things like `git pull`ing the packages manually.

As a specific example, both for concreteness and to forestall the “just submit it to the AUR” objection:

I like the [Terminus font](http://terminus-font.sourceforge.net/ "Font homepage"), which is available as a package in [`community/terminus-font`](https://archlinux.org/packages/community/any/terminus-font/).  However, it supports a number of variant character appearances, and I like a few of the options more than the defaults.  I could do what [several other people](https://aur.archlinux.org/packages/?O=0&K=terminus-font "AUR search page") have done and upload a custom configuration to the AUR, but nobody else would use it and it would be harder to make sure I stayed in sync with upstream, and wouldnʼt actually make things any easier.

So unless I end up switching to Gentoo where [the equivalent package](https://packages.gentoo.org/packages/media-fonts/terminus-font) has USE flags for this kind of thing, Iʼm better off installing packages that donʼt come from the AUR.  But Iʼd still like to use as much of this as still applies, which is a lot.

__Before filing a bug, check the following:__

1) Do you use the latest release of aurutils and its dependencies? If so,
is the issue reproducible from the master branch? Yes
**Other questions deleted as not relevant**

__If yes to all, create a debug log:__with

```
AUR_DEBUG=1 aur <name> <arguments> >aurutils.log
```
N/A

AladW:
> As far as I can tell, there is not, but primarily because aur-depends is called unconditionally by aur-sync and I have to drop down to aur-build, and do things like git pulling the packages manually.

You don't need to git pull manually, because there's `aur-fetch` - which `aur sync` uses as well. The only unique aspect about `aur sync` is its file inspection mechanism. Since that one is already decoupled, I planned to move it to a seperate script (`aur-view`).

> As a specific example, both for concreteness and to forestall the “just submit it to the AUR” objection:

One specific aim of the modular nature of `aurutils` is that you can use it for other tasks than just to AUR, so I don't object.

dhouck:
[...]

PR #862, opened 2021-11-27
aur-fetch: add --existing

AladW:
Possible options to make `aur-fetch` more suitable for non-AUR workflows. Related discussion: #860

### DIFF
--- a/lib/aur-fetch
+++ b/lib/aur-fetch
@@ -7,7 +7,7 @@ XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[1]}(): }'
 
 # default options
-recurse=0 sync=no
+clone=1 recurse=0 sync=no
 
 results() {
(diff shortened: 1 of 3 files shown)


dhouck:
There are uses for `--clone-url` and it makes sense to add it, but since my example use case for creating #860 involved official packages, I just want to point out that for official packages this wouldnʼt work.

All packages are cloned from the same repository, just with different branches, for reasons that I assume would make sense if I found the discussion about moving from svn to git.  For that youʼd need to either call `asp checkout $pkg`, or pass more complicated arguments to `git clone` to set the branch and not download more of the repo than necessary.

Given that a lot of other things also need to be different (eg. if you use `asp`, you need to run `asp update` before anything that `aur-fetch` does), and that `aur-fetch` shouldnʼt try cloning such packages unless explicitly requested (since if theyʼre just needed as dependencies theyʼll already be in the official repositories as binary packages), I donʼt know if itʼs worth supporting this workflow more directly.

AladW:
Thanks for the comment. In that case I'd only consider `--existing`. Setting the clone URL through `AUR_LOCATION` makes the assumptions for it (i.e. no branches) more explicit.

Issue #863, opened 2021-11-27
aur-view: decouple package inspection from `aur-sync`


AladW:
The only unique functionality in `aur-sync` is its package inspection mechanism, with support for `vifm` and storage of viewed commit IDs. This is already decoupled from the rest of `aur-sync` (the IDs are even stored in `$XDG_CACHE_HOME/aurutils/view`, and should be simple to move to a new script, `aur-view`.

Related to custom workflows such as #840 and #860, where `aur-sync`, apart from its inspection mechanism, is ill-suited.

PR #865, opened 2021-11-29
aur-sync: move inspection mechanism to aur-view

AladW:
#863 

A possible issue is what do with the build order. It is displayed in `aur-view` (as part of the same directory as the symbolic links to package directories), but cannot be modified as a copy. Thus it should either be modifiable, or not shown in the first place. `chmod -w` to emphasize the read-only aspect might be confusing.

A man page for `aur-view` is also missing.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -357,12 +357,14 @@ fi
 
 # Verify dependency tree (#20)
 if (( graph )); then
-    readarray -t queue_a < "$tmp"/queue
-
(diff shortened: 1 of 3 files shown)


AladW:
I see two options to address the build order file:
1. Store everything as an array
2. Create a subdirectory `$pkg` and save the file above it

Latter would break things for people who assume the `view.??????` path (e.g. for vifm configuration), so I guess it'd need a minor version bump.

AladW:
I went with the first approach, optionally taking inputs from `stdin` by specifying `-` or `/dev/stdin`.

PR #876, opened 2021-12-08
aur-sync: apply --ignore to repository targets

AladW:
This is comparable to having the ignored packages in a separate repository.

Also remove the big WARNING for ignored packages; just print them to stderr.


AladW:
A perhaps unfortunate consequence of `grep -w` is that users could now place `pkgver-pkgrel` in their ignore file, and have it removed from the `aur repo` output. The `aur-depends` output would still be filtered on `pkgname` exclusively.

PR #882, opened 2021-12-23
Add & fix dependency selection

### LINKED ISSUE #845: aur-sync: missing dependencies error when trying to resolve cyclic dependencies with --no-checkdepends
So, I have the following situation:

- package A depends on package B and C
- package C has a checkdepends on package B

When using `--no-checkdepends` aurutils filters out package B as it is a checkdepends of C, but then the `aur sync` call fails because it can't find the dependency B for package A.

adrien1018:
- Add `--no-depends`, `--no-makedepends`, `--no-checkdepends` options to aur-graph
- Use `aur graph --no-checkdepends` when `--no-check` is given to aur-sync.
- The dependency selection code in aur-depends was incorrect (it would fetch unneeded dependencies); this is now fixed.

Also fixes #845.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+aur
+completions/bash/aur
+completions/zsh/_aur
--- a/lib/aur-depends
+++ b/lib/aur-depends
@@ -54,6 +54,9 @@ tr_ver() {
 chain() {
     local a num sub
 
+    select_args=("$1" "$2" "$3")
+    shift 3
+
     aur query -t info "$@" > json/0 || exit
     num=$(count json/0)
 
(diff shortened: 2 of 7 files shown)


AladW:
While technically correct, I don't think the 1.5x increase in `aur-graph` for an edge case is worth it. Moreso when it's only about avoiding cycles, and `aur-graphs` does not check those.

For filtering `checkdepends` in every iteration of `aur-depends/chain`, I'm also curious if, in practice, this actually leads to less requests being made.

adrien1018:
For `aur-depends`, It is not only about request optimization. Take `aur depends ruby-activemodel --no-checkdepends` as an example. The current version will output 30 packages while in fact only `ruby-activemodel` should be outputted (all the others are something like depends of checkdepends), results in `aur sync ruby-activemodel --nocheck` building unnecessary packages.

A possible use case of the `aur-graph` options is when sorting pre-existing packages before building them using `aur build --nocheck` (which is what I do in my custom repo). Surely it is not that useful regarding aur-sync, but it would still be a nice feature to have.

AladW:
OK. In that case I want to think some more about the exact implementation (which preferably limits the inflation of `aur-graph`). In the meantime, `aur-depends` and `aur-graph` are unlikely to change so anyone interested can already use this PR.

`--nocheck`, `--table` might be also be useful for `aur-graph`.

AladW:
See #936 for `aur-depends` changes

PR #884, opened 2021-12-26
aur-build.1: Remove mention of `keyserver-options`

FichteFoll:
The option has been deprecated and is now just called `auto-key-retrieve`.

>  --keyserver-options {name=value}
> …
>    auto-key-retrieve
>      This is an obsolete alias for the option auto-key-retrieve.  Please do not use it; it will be removed in future versions..


AladW:
Cheers

PR #919, opened 2022-02-11
aur-query: add set operations

AladW:
This also adds the `--symdiff` flag and properly makes `--any` return the union, instead of the symmetric difference, of results.

- [x] Add `jq` for intersection
- [ ] Add `jq` for union (with unique entries)


AladW:
@Earnestly thoughts on moving the `jq` filters to `aur-query`? This way, `aur-search` does not have to deal with implementation details of aurweb (i.e. set operations not being available from the aurweb API), and could perform set operations just by using the right `aur-query` options.

Earnestly:
To the extent that my implementation returns aurweb-styled results, specifcally their API: `{"resultcount": ..., "results": [ ... ]}`, I think that `aur-query` should probably do such filtering.  This would make it seem transparent to any user (such as `aur-search`) expecting to parse aurweb json.

Earnestly:
I believe all unique entries can be done by `jq -Mcs 'add | .results as $r | {"resultcount": ($r | length), "results": $r}'`

Earnestly:
A potential test case for the intersection sorting might be querying the real aurweb and comparing the web results with `aur search`:

```
% aur search -r mkdocs material | jq -r '.results[].Name' | sort
mkdocs-material
mkdocs-material-extensions
mkdocs-material-pymdownx-extras

% curl -s 'https://aur.archlinux.org/packages?K=mkdocs+material' | hq 'tbody > tr > td:first-child > a' text | sort
mkdocs-material
mkdocs-material-extensions
mkdocs-material-pymdownx-extras
```

AladW:
What I did is leave the JSON Lines format for `aur-query` as default (in case any consumers rely on this fact), and add `--all` and `--any` flags. `aur-search`, as a higher-level tool, then defaults to `aur-query --any` for `--raw` (alias `--json`).

I've also added the proposed test. For `aur query --any`, `jq add` works, but `resultcount` also needs to be updated (and duplicate entries removed). This is probably a special case of what is done for `--all`.

AladW:
```bash
jq -Mjcs 'map(.results) | unique | add as $r
            | {
                "resultcount": ($r | length),
                "results": $r
            }'
```
seems to work.

Earnestly:
`jq -s add` appears to work on its own.  I get, in gron-like format:

```
.resultcount = 628
.results[0].Description = "Extract data from gcm and szs files. Contains gcmdump, yaz0dec, and rarcdump"
[...]
.results[627].Version = "0.3.0-3"
.type = "search"
.version = 5
```

The `.type` and `.version` fields are something I missed from the intersection stuff, which is rebuilding the results from scratch.

lib/aur-query
```diff
@@ -43,15 +43,17 @@ rpc_info() {
...
     local rpc_url="$AUR_QUERY_RPC"
     local rpc_ver="$AUR_QUERY_RPC_VERSION"
-    local splitno="$AUR_QUERY_RPC_SPLITNO_POST"
+    # FIXME: data-urlencode appends across multiple urls, resulting in
+    #        N / splitno requests with the same contents (#920)
+    #local splitno="$AUR_QUERY_RPC_SPLITNO_POST"
 
     # Write opening and closing quotes with \x22 (hexadecimal)
-    awk -v rpc="$rpc_url" -v version="$rpc_ver" -v splitno="$splitno" '{
-        if (NR % splitno == 0 || NR == 1) {
```
Earnestly: `!((NR-1) % limit) { ... }`

tests/issue/918
```diff
@@ -0,0 +1,20 @@
...
+set -e -o pipefail
+tmp=$(mktemp -d)
+trap 'rm -rf -- "$tmp"' EXIT
+
+term1=mkdocs
+term2=material
+cd "$tmp"
+
+# RPC query (set intersection)
+aur search --raw "$term1" "$term2" | jq -r '.results[].Name' | sort >"$tmp"/a
```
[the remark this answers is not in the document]
AladW: Using this approach, I used a randomized test with 10 packages.

lib/aur-query
AladW: Makes sense

AladW:
Maybe I'm missing something, but I can't reproduce it:
```bash
$ AUR_QUERY_RPC_SPLITNO_POST=1 aur query -t info aurutils aurutils-git | jq -s add | gron | grep .Name
json.results[0].Name = "aurutils";
```
```bash
$ AUR_QUERY_RPC_SPLITNO_POST=1 aur query -t info aurutils aurutils-git | jq -c | gron -s | grep .Name
json[0].results[0].Name = "aurutils";
json[1].results[0].Name = "aurutils-git";
```

Earnestly:
The `-s` is quite necessary for `jq`

AladW:
That is one possibility, but the main use is when info requests are split, i.e. by 150 (GET) or 730 (POST). `jq` can deal with the output regardless, but it's an option regardless (and allows experiments like checking if `packages-meta-ext-v1.json`* can be replicated locally).

* which, coincidentally has no `resultcount`, `type` or `version`.

Earnestly:
For duplicates within a single object array you'd have to use something like `.results | unique_by(.)` for the whole object, or `.Name` for a specific key within.

AladW:
As aurweb does not respect its own specified limits (i.e. https://gitlab.archlinux.org/archlinux/aurweb/-/blob/master/conf/config.defaults#L24) trying to make `aur-query` work with the maximum number of arguments is a futile effort. I've replaced the `union` filter with `jq -Mjcs add`, which indeed works fine with `-t search`.

AladW:
I'm still having random issues with `jq -s add`:
```bash
aur query -t search bacula-dir-mariadb ros-melodic-mbf-costmap-nav kaption helium-ledger-cli python-jamo-git qdc kwin-scripts-sticky-window-snapping-git maestral-qt exec-helper-docs deepin-wine-plugin | jq -s add
```
```json
{
  "resultcount": 2,
  "results": [
    {
      "Description": "Deepin Wine plugin virtual package",
      "FirstSubmitted": 1524900490,
      "ID": 814453,
      "LastModified": 1604648028,
      "Maintainer": "Skywol",
      "Name": "deepin-wine-plugin-virtual",
      "NumVotes": 4,
      "OutOfDate": null,
      "PackageBase": "deepin-wine-plugin-virtual",
      "PackageBaseID": 131921,
      "Popularity": 0.598254,
      "URL": "http://www.deepin.org",
      "URLPath": "/cgit/aur.git/snapshot/deepin-wine-plugin-virtual.tar.gz",
      "Version": "5.1.13-1"
    },
    {
      "Description": "Deepin Wine plugin",
      "FirstSubmitted": 1524899634,
      "ID": 814455,
      "LastModified": 1604648148,
      "Maintainer": "Skywol",
      "Name": "deepin-wine-plugin",
      "NumVotes": 4,
      "OutOfDate": null,
      "PackageBase": "deepin-wine-plugin",
      "PackageBaseID": 131920,
      "Popularity": 0.598254,
      "URL": "http://www.deepin.org",
      "URLPath": "/cgit/aur.git/snapshot/deepin-wine-plugin.tar.gz",
      "Version": "5.1.13-1"
    }
  ],
  "type": "search",
  "version": 5
}
```

CyberShadow:
If you just want to pick the first objects in an array with a field having some value, you could try e.g. `group_by(.ID) | map(.[0])`.

Otherwise a self-contained example of the problem would help :)

PR #939, opened 2022-02-24
query: abort if no query type is specified

### LINKED ISSUE #938: line 144: curl_config: command not found
Problem when running `aur query linux`.
```
$ aur query linux
/usr/lib/aurutils/aur-query: line 144: curl_config: command not found
```
I cannot find the curl_config package in aur or arch repos, no curl_config executable exists in the curl package either.

AladW:
Fixes #938

### DIFF
--- a/lib/aur-query
+++ b/lib/aur-query
@@ -77,7 +77,7 @@ trap_exit() {
 }
 
 usage() {
-    printf 'usage: %s [-t [info|search] ] [-b by] [-aer] <package...>\n' "$argv0" >&2
+    printf 'usage: %s -t [info|search] [-b by] [-aer] <package...>\n' "$argv0" >&2
     exit 1
 }
 
@@ -134,6 +134,8 @@ elif [[ $arg_type == "info" ]] && (( AUR_QUERY_RPC_POST )); then
 
 elif [[ $arg_type == "info" ]]; then
     curl_config() { jq -R -r '@uri' | rpc_info; }
+else
+    usage
 fi
 
 # generate curl configuration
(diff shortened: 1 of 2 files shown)


lib/aur-query
```diff
@@ -77,7 +77,7 @@ trap_exit() {
 }
 
 usage() {
-    printf 'usage: %s [-t [info|search] ] [-b by] [-aer] <package...>\n' "$argv0" >&2
+    printf 'usage: %s -t [info|search] [-b by] [-aer] <package...>\n' "$argv0" >&2
```
Earnestly, line 80: `usage: info|search [...]` would imply `info` or `search` is required.  Does it make sense to just make one of the the default and return to `[-t info|search]`?

AladW: I've thought of making `search` the default, but it's difficult to argue that either default makes more sense than the other one. For `aur-search`, the default is `search` (clearly).

Earnestly: In that case a more usual synopsis might look like: `usage: aur query [-aer] [-b by] info|search package...`

AladW: I've made some observations regarding default arguments which I think lead to an overall better design: https://github.com/AladW/aurutils/pull/944

With those changes, `aur query <args>` returns:
```
{"error":"No request type/data specified.","resultcount":0,"results":[],"type":"error","version":5}
```

AladW: About positional arguments for the type, not sure about that, I've used `-t <type>` for a pretty long time. I guess it could be considered for further tools (`aur-sync--ninja` uses positional arguments exclusively.)

PR #947, opened 2022-02-27
depends, search, sync: add `--optdepends`

### LINKED ISSUE #945: aur-depends: take `optdepends` into account
Some packages like https://aur.archlinux.org/packages/deal-ii encode optional compile-time dependencies as `optdepends`, which can be an annoyance if you're just looking to build with all features enabled. `aur-depends` could take these dependencies into account to side-step the issue. `aur search --info` should also display `optdepends`.

AladW:
Fixes #945


AladW:
I wonder if there's some way to incorporate `select_deps` in `aur-depends` directly into the `jq` filter. 

https://github.com/AladW/aurutils/blob/master/lib/aur-depends#L39-L47

`jq --arg` maybe.

Issue #907, opened 2022-02-05
`build`: improve unattended builds


FichteFoll:
The current situation of having unattended builds of many packages with a build user (as documented in `aur-build(1)` and `aur-chroot(1)`) is unsatisfactory. The setup suffers from file permission conflicts between the main/current user and the build user, different sync and seen caches and potentially different `AUR_PAGER` setups, which would all need to be synced or worked around/ignored.

An alternative to the separate build user would be to increase or disable the sudo authentication timeout, potentially temporarily for the build duration, which is probably better to work with (and what I will consider picking up for now) but not ideal either because it's more general in scope than a dedicated build user with appropriate configuration.

---

The simplest approach to this is probably to run `aur` commands itself with elevated permissions, but that's sort of the "golden hammer" solution and some de-elevation of privileges is needed on 3 to 4 occasions (depending on package signing):

- https://github.com/AladW/aurutils/blob/master/lib/aur-build#L308-L313 for the packagelist
- https://github.com/AladW/aurutils/blob/master/lib/aur-build#L353-L355 for pkgver
- https://github.com/AladW/aurutils/blob/master/lib/aur-build#L387-L388 for the actual makepkg command
- gpg for signing

quoting @AladW:
> In short, to make a root `aur-build` work 
> 1. drop permissions for user commands (with `setpriv`) instead of elevating permissions for root commands 
> 2. replace `makepkg -s` (which elevates permissions) with `makepkg --printsrcinfo` (as user, to retrieve a list of dependencies) and `pacinstall` (as root) 
> 3. avoid prefixing commands with `setpriv` or `sudo` by redefining commands (as is already done for `$PACMAN_AUTH`, `$MAKEPKG`), if a build user is specified on the command-line (for example with `-U <build_user>`)

Another option (suggested by @Earnestly) would be a daemon that performs tasks that require permissions via IPC.

Related: #852

---

*Discussed on IRC from 2022-02-05 14:51 UTC+1 to 18:15 UTC+1.*

AladW:
Note that there are potential differences between `aur-chroot` and regular builds, because `aur-chroot` (devtools) already drops to an unprivileged user with `sudoers` access to pacman for the build. In particular, chroot builds are fixed to use `makepkg -s`.

AladW:
For host builds, I think the simplest approach is to add two helpers: 
1. `aur-build--privileged`, which swallows `--syncdeps`, `--rmdeps`, and uses `setpriv`
[...]

PR #951, opened 2022-03-01
chroot: do not exit on chroot creation

AladW:
https://github.com/AladW/aurutils/issues/907#issuecomment-1055780519


AladW:
I'm going to split out the commits that re-exec as root (breaking existing sudoers rules) to a different PR.

Issue #952, opened 2022-03-01
`sync`: `--ignore` should probably ignore packages provided via command-line


Earnestly:
The `aurvcs.7` manual makes suggestion to provide packages via command-line, but at the same time it's not unreasonable for people to also ignore some of those if some are temporarily failing.

While it is possible to filter out the broken packages before providing them to the command-line itself this sort of behaviour is a little surprising.

If explicit listing on the command-line is the way to override an ignore from an ignore-file, perhaps a better approach is to use `[--no-ignore=pkg,...]...`

Essentially there's nothing functionally wrong with the current design but it is surprising.

AladW:
The reason I went with this approach in 07ba7819d3f2b640d80f305789b3f217e611a36c is that
1. packages on the `aur-sync` command-line already take precedence for other purposes (`--provides`, `--no-ver-argv`)
2. it allows to quickly check an ignored package for updates (e.g. a fix for a regression) without modifying the (more permanent) `--ignore-file`

And (FWIW) `pacman` has the same behavior. I haven't thought on how it affects the workflow `aurvcs`, though.

AladW:
Something like
```diff
diff --git a/man7/aurvcs.7 b/man7/aurvcs.7
index 577de4e..cb42396 100644
--- a/man7/aurvcs.7
+++ b/man7/aurvcs.7
@@ -30,6 +30,7 @@ Any option arguments are forwarded to
 .SS Program source
 .EX
     #!/bin/bash
+    set -e
     XDG_CACHE_HOME=${XDG_CACHE_HOME:\-$HOME/.cache}
     AURDEST=${AURDEST:\-$XDG_CACHE_HOME/aurutils/sync}
     AURVCS=${AURVCS:\-.*\-(cvs|svn|git|hg|bzr|darcs)$}
@@ -43,19 +44,24 @@ Any option arguments are forwarded to
         grep \-Fxf \- <(printf \(aq%s\en\(aq *) | xargs \-r aur srcver
     }
 
-    db_tmp=$(mktemp)
-    trap \(aqrm \-rf "$db_tmp"\(aq EXIT
+    # Scratch space for intermediary results.
+    tmp=$(mktemp \-d)
+    trap \(aqrm \-rf "$tmp"\(aq EXIT
 
     # Retrieve a list of the local repository contents. The repository
     # can be specified with the usual aur\-repo arguments.
-    aur repo \-\-list "$@" >"$db_tmp"
[...]

PR #953, opened 2022-03-02
aurvcs.7: implement sample code with `aur-view`

AladW:
https://github.com/AladW/aurutils/issues/952#issuecomment-1057192219


lib/aur-srcver
```diff
@@ -92,12 +89,26 @@ for n in "$@"; do
 done
 wait
 
+# Check if targets are available
+dirs=("$tmp"/*/)
```
Earnestly, line 93: nullglob?

PR #964, opened 2022-03-12
examples: rewrite `sync-devel`, `sync-list`

AladW:


AladW:
`sync-list` turned out to be exceedingly complicated. I guess it demonstrates that 1. a more simple approach is preferable 2. limitations of the `aur-repo` command-line.

PR #969, opened 2022-03-22
aur-fetch: add --no-pager to the git log command

nl6720:
Invoking the pager needlessly interrupts/delays the command by forcing interactive action to close the pager.


AladW:
:+1:

PR #973, opened 2022-03-26
examples: add `chroot-batch`

### LINKED ISSUE #907: `build`: improve unattended builds

AladW:
Issue #907

### DIFF
--- a/examples/chroot-batch
+++ b/examples/chroot-batch
@@ -1,0 +1,111 @@
+#!/bin/bash
+[[ -v AUR_DEBUG ]] && set -o xtrace
+set -o errexit
+shopt -s nullglob
+argv0=chroot-batch
+startdir=$PWD
+PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+
+# assumption: pkgbase == dirname
+queue=$1 db_name=$2 db_root=$3 build_user=${4:-$SUDO_USER} overwrite=${5:-0}
+build_args=(-a "$queue" -d "$db_name" --root "$db_root")
+
+# resolve symlinks
+db_path=$(realpath -e -- "$db_root/$db_name".db)
+files_path=$(realpath -e -- "$db_root/$db_name".files)
+
+# re-exec as root
+if (( EUID != 0 )); then
+    exec sudo --preserve-env=GNUPGHOME,SSH_AUTH_SOCK,AUR_DEBUG "${BASH_SOURCE[0]}" "$@"
+fi
+
+# run user/unpriviliged commands with setpriv(1)
+runas() { 
+    setpriv --reuid "$1" --regid "$1" --init-groups --reset-env -- "${@:2}"
+}
+
+# PKGEXT in container may differ from host
+chroot_makepkg_conf=$(aur chroot --path)/etc/makepkg.conf
+build_args+=(--makepkg-conf "$chroot_makepkg_conf")
+
+# 1. check which packages in the queue need to be built
+mapfile -t build_dry_run < <(
+    runas "$build_user" aur build "${build_args[@]}" --dry-run --pkgver 2>/dev/null)
+wait "$!"
+
+if (( overwrite == 0 )); then
+    printf '%s\n' "${build_dry_run[@]}" | awk '/^exist:/ {print}'
+
+    mapfile -t pkgbase_targets < <(
+        printf '%s\n' "${build_dry_run[@]}" | awk -F: '/^build:/ {print $2}')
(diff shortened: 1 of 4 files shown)


AladW:
Following https://github.com/AladW/aurutils/issues/907#issuecomment-1070958818 I would also add `aur-sync --build-user`,  but something like:
```bash
# Specify user for aur-build (#907)
if [[ -v build_user ]]; then
    ninja() {
        sudo -u "$build_user" ninja "$@"
    }
    aur() { # only applies to build--ninja, build below
        sudo -u "$build_user" aur "$@"
    }
fi

# Build dependency tree with ninja (#908)
if (( AUR_SYNC_USE_NINJA )); then
    ...
```
results in 3 `sudo` calls with `AUR_SYNC_USE_NINJA=1`, namely
```
sudo -u "$build_user" aur sync--ninja "$AURDEST" >build.ninja
sudo -u "$build_user" env NINJA_STATUS='[%s/%t] ' ninja -k "$keep_going"
sudo -u "$build_user" env NINJA_STATUS='[%s/%t] ' ninja -nC /var/empty -f "$PWD"/build.ninja
```
i.e. potentially 3 prompts instead of 1.

examples/chroot-batch
```diff
@@ -0,0 +1,80 @@
...
+
+repo_add_and_sign() {
+	# ensure the passphrase is not stored as an environment variable
+	# by saving it in a local variable inside a function
+	local passphrase pkglist pkgbase
+	read -p 'GPG passphrase for signing packages: ' passphrase
+
+	#global var_tmp build_user db_path
+	for pkgbase in "$@"; do
+		cd "$startdir"
```
Earnestly, line 57: Perhaps `cd -- "$startdir"/"$pkgbase"`?

examples/chroot-batch
```diff
@@ -0,0 +1,80 @@
...
+		cd "$pkgbase"
+		PKGDEST="$var_tmp" aur chroot --build --cargs -c,-u # makechrootpkg -cu
+
+		cd "$var_tmp"
+		pkglist=(*) siglist=()
```
Earnestly, line 62: May need `nullglob` here.

examples/chroot-batch
AladW: cf. c7964918fd9f82931c2c6ee2af6b7153206acbda

examples/chroot-batch
AladW: I'm trying to think of a scenario where `pkglist` would be empty, but generally yes.

AladW:
The assumption `dirname == pkgbase` could be dropped if the `--dry-run` output is null-separated (there are two relevant paths, one for the package in the local repository, the other for the `PKGBUILD` directory). This assumption is fulfilled when using `aur-sync` and `aur-fetch`, so I'll leave this aside for now.

PR #977, opened 2022-04-02
aur-build: drop privileges with `setpriv(1)`

AladW:
There are three cases:

1. If `UID == 0`, drop privileges for user commands (e.g. `repo-add`, `gpg`, `makepkg`)
2. If `UID > 0`, raise privileges for root commands (`pacman`)
3. Otherwise, run the command as-is (e.g. `aur-chroot`)

This approach and `gpg-set-passphrase(1)` renders `examples/chroot-batch` obsolete, demonstrative purposes excepted. It can replaced by a corresponding section in `aur-build(1)`.


AladW:
This turned out a lot more work than anticipated. On the flip-side, using `setpriv(1)` exposed two long-standing bugs in `aur-repo`.

lib/aur-build
```diff
@@ -203,24 +216,42 @@ while true; do
     shift
 done
 
+# mollyguard for makepkg
+if (( UID == 0 )) && { [[ ! -v build_user ]] || [[ ! -v AUR_ASROOT ]]; }; then
```
Earnestly, line 220: The second condition is always true, instead: `&& ! { [[ -v build_user ]] || [[ -v AUR_ASROOT ]]}` https://en.wikipedia.org/wiki/De_Morgan%27s_laws

lib/aur-build
```diff
@@ -373,7 +407,7 @@ while IFS= read -ru "$fd" path; do
                 (( dry_run )) && printf '%s:%s:%s\n' build "$pkgbase" "file://$pkgpath"
             fi
         # pkgbase may differ from pkgname; prefix to package path with --full
-        done < <(PKGDEST="$db_root" aur build--pkglist --full "${pkglist_args[@]}")
+        done < <(as_user PKGDEST="$db_root" aur build--pkglist --full "${pkglist_args[@]}")
```
Earnestly, line 410: As `setpriv` runs the command without a shell are you sure `as_user PKGDEST=...` works as intended, should it perhaps not be: `PKGDEST=... as_user ...`?  The `as_user` function uses `env` but the `"$@"` comes after the terminating `--`.

makepkg/aurutils.changelog
```diff
@@ -1,3 +1,23 @@
+## 9.1
+
+`aur-build` now uses `runuser(1)` to run unprivileged commands (`makepkg`, `gpg`,
```
Earnestly, line 3: Mentions `runuser` here but uses `setpriv`

AladW: https://github.com/AladW/aurutils/commit/1b1d86496612cb785332ad468fd6780ac61af6bb

lib/aur-build
AladW: These are the possible combinations:
```
$ foo=1 env printenv foo
1

$ env foo=1 printenv foo
1

$ env foo=1 -- printenv foo
env: ‘--’: No such file or directory

$ env -- foo=1 printenv foo
1

$ env -C ~ -- foo=1 printenv foo
1

$ env -- -C ~ foo=1 printenv foo
env: ‘-C’: No such file or directory
```
so it's perhaps unexpected, but it works

lib/aur-build
AladW: If both `build_user` and `AUR_ASROOT` are set, the second condition is false, no?

Earnestly: `(not A) or (not B)` is always true regardless of truthiness of A or B.

Note that with bash you can use a single `[[` with short-circuiting or/and and even `( expression )` notation (as distinct from subshell).

E.g. If you want to error out if UID is 0 and neither build_user or ABS_ASROOT is set then:

```sh
if [[ $UID == 0 && ! (-v build_user || -v ABS_ASROOT) ]]; then ...
```

As such this will be false if either build_user or ABS_ASROOT is set.

If you want this if statement to be true if both build_user and ABS_ASROOT is set, or if both are unset then you'll need more conditions:

```sh
if [[ $UID == 0 && ( ! (-v build_user || -v ABS_ASROOT) || (-v build_user && -v ABS_ASROOT)) ]]; then
```

With this it will be false if either build_user is set of ABS_ASROOT is set, but not both.

AladW: > (not A) or (not B) is always true regardless of truthiness of A or B.

I made a truth table:

img

Demonstration:

```bash
# A true, B true
user_id=0
build_user=
AUR_ASROOT=

(( user_id == 0 )) && { [[ ! -v build_user ]] || [[ ! -v AUR_ASROOT ]]; }
echo $?

# A true, B false
user_id=0
build_user=
unset AUR_ASROOT

(( user_id == 0 )) && { [[ ! -v build_user ]] || [[ ! -v AUR_ASROOT ]]; }
echo $?

# A false, B true
user_id=0
unset build_user
AUR_ASROOT=

(( user_id == 0 )) && { [[ ! -v build_user ]] || [[ ! -v AUR_ASROOT ]]; }
echo $?

# A false, B false
user_id=0
unset build_user
unset AUR_ASROOT

(( user_id == 0 )) && { [[ ! -v build_user ]] || [[ ! -v AUR_ASROOT ]]; }
echo $?
```
```
1
0
0
0
```
So the idea is that the `setpriv` logic requires both `AUR_ASROOT` and `build_user`, otherwise there's an error. With the Morgan, this would be (note the `&&`)
```
(( user_id == 0 )) && ! { [[ -v build_user ]] && [[ -v AUR_ASROOT ]]; }
```

Earnestly: Hm, fair enough.  The final construction is nice, but I still think you're better off using `[[` instead of `((` (which enables arbitrary code evaluation in < 5.1 bash).

AladW: Do you have some more information on `[[`, `((` and code evaulation?

Earnestly: http://mywiki.wooledge.org/BashProgramming/05#Arithmetic_Expansion

It also applies to `test -v`, see also https://lists.gnu.org/archive/html/bug-bash/2021-10/msg00175.html

AladW: Thanks, both changes are in 9.2 now.

PR #983, opened 2022-04-09
`aur`: load environment variables

AladW:
This is more convenient than changing environment variables globally (e.g. in ~/.bashrc, which is sourced by ~/.bash_profile), while not being as complex as a fully-fledged configuration system.

- [x] add to `aur.in`
- [x] document in `aur(1)`

### DIFF
--- a/aur.in
+++ b/aur.in
@@ -1,4 +1,5 @@
 #!/bin/bash -
+readonly XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 readonly argv0=aur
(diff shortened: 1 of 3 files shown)


AladW:
Conceptually this is similar to `makepkg.conf`, although `makepkg` sources its configuration instead of specifically exporting variables. In particular, `makepkg.conf` can contain arrays, while the environment file proposed here cannot.

Regardless, after adding another `AUR_` environment variable with https://github.com/AladW/aurutils/commit/4b5916e802941785e21b2767a0418891bb4208a7, this looks like a worthwhile addition.

Issue #837, opened 2021-10-09
`sync`: pick up a "provides" package from the command line


CyberShadow:
Consider: <code>aur sync [bitlbee-libpurple-git](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=bitlbee-libpurple-git) [bitlbee-discord-git](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=bitlbee-discord-git)</code>

`bitlbee-libpurple-git` provides `bitlbee`, and `bitlbee-discord-git` depends on `bitlbee`, so one would expect that `bitlbee-discord-git` is built against `bitlbee-libpurple-git`. However, aurutils pulls in the [`bitlbee`](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=bitlbee) package instead.

Not sure if this is a defect or maybe there is a good reason for this. However, it is at least inconsistent with the behavior when `bitlbee-discord-git` is available in a repository; then, aurutils does not pull in `bitlbee`, but just uses `bitlbee-discord-git` which provides it.

AladW:
https://gitlab.archlinux.org/archlinux/aurweb/-/issues/307

AladW:
With RPCv6, this should be as simple as:
```
{ aur query -t search -b provides - <argv | jq -r '.results[].Name'
  ...
} > filter
```

AladW:
This should work (at the cost of a large file download):
```
$ aur pkglist --info | jq -r '.[] | select(.Provides[]? | contains("libcubeb.so")) | .Name'
```

VorpalBlade:
This is also an issue with .so provides on the command line. Example command which triggered issues:

```
aur sync -c qtforkawesome qtutilities c++utilities syncthingtray
```

qtforkawesome fails to build because (according to alad on IRC):
> &lt;alad&gt; Vorpal, oh, I see now for 1): error: target not found: libqtforkawesome.so - this would likely by solved by using aur-graph to resolve the dependencies a second time, using the .SRCINFO

AladW:
`provides` on the command line are addressed with 9680df6ae9318c5dcbf52a7ee3ac3690521944e5

AladW:
So, with `aur-format` merged you can create a `shlibs` file as follows:
```bash
aur pkglist --info | aur format -f '%n\t%P\n' --delim=, \
[...]

PR #990, opened 2022-04-25
[Feature] Better zsh completion

VorpalBlade:
This is still work in progress. It is a draft PR.
* There are lots of TODO comments.
* Many parts are not well tested.
* The history is likely to be messy and also likely to end up squashed, reworked etc

Still, feedback is welcome.


VorpalBlade:
Commit [ad4ce1f](https://github.com/AladW/aurutils/pull/990/commits/ad4ce1f49fbf6f772ce3763bec5a8b37d404f6d1) is a reasonable example of what updating the completion file in the future when the command line interface changes would entail. Not difficult (generally) but another thing you will end up having to remember to do.

Note: That commit also includes a bug fix to the definition of -U/--user in aur-build (the format should be `--flag=:description of the argument to the flag:completion_function`. The middle section was missing (but is required when you have a completer function. Depending on shell settings the middle value may be displayed to the user to aid them.

AladW:
Sorry I've been busy lately. I can give some comments on the TODOs:
> # TODO: It is not clear to me how mode flags interact with source flags.
> # Specifically --search and --info also overwrite mode. Are those exlusive with mode flags?

The `aur-pkglist` command-line is problematic because behavior depends on the order of the command-line flags (e.g. `--info --quiet` vs. `--quiet --info`). That's something to be fixed in a future version.

> # TODO: Undocumented?
`aur-pkglist --plain` should be documented or the flag left out

> # TODO: "[...] or any other field generated by repo-add"
I left this in because I don't know what fields `repo-add` might add in the future. I don't know of a way to retrieve all possible fields programmatically.

> # TODO: Possibly be smarter about positional arguments based on what search_by is specified
For an info search `__aur_list_pkgs` could be used, a regular search (where the term does not have to match any existing packages) should not be completed.

> # TODO: This should maybe be only local packages (i.e. {aur repo -l | cut -f1})?
Good point. It seems odd that `aur repo -q` implies `aur repo -u`, and does not apply to `-l`.

AladW:
> Commit [ad4ce1f](https://github.com/AladW/aurutils/pull/990/commits/ad4ce1f49fbf6f772ce3763bec5a8b37d404f6d1) is a reasonable example of what updating the completion file in the future when the command line interface changes would entail. Not difficult (generally) but another thing you will end up having to remember to do.

Looks reasonable to me

VorpalBlade:
> The aur-pkglist command-line is problematic because behavior depends on the order of the command-line flags (e.g. --info --quiet vs. --quiet --info). That's something to be fixed in a future version.

That would be difficult to express using _arguments in zsh and would require manual parsing. And since it is likely to change in the future there doesn't seem to be much point in doing so. I'm just going to not try to handle that.

> `aur-pkglist --plain` should be documented or the flag left out

Yes the man page needs to be updated! (Hint hint!) I added `'--plain[list output one package per line, default behaviour]'` for now. That seems accurate to me. But if it becomes a hidden option it should be removed.

> I left this in because I don't know what fields repo-add might add in the future. I suppose one could retrieve all fields programmatically by printing all %...% headers in a repository .db.

I found (at least) one flag that was actually missing in the man page already: optdepends. So yeah. I haven't done a complete check though. Automatically retrieving that list would be nice. Probably best if it is implemented as a hidden option that the completion code can call to get it, one option per line.

> Good point. It seems odd that aur repo -q implies aur repo -u and does not apply to -l.

`aur repo -luq` should work instead of what I wrote I believe (can't test it as I don't have any outdated packages on my system right now). That should (I believe) be the list of all things that would make sense to ignore?

AladW:
#993 fixes the remaining inconsistencies between tab-completion and documented options.

> It's a compromise between the "most relevant" ignore targets (AUR upgrades which you may not want) and a slight delay on the tab completion. Either way, in https://github.com/AladW/aurutils/pull/992 -q applies to both --list and --upgrades.

Another thing to note is that `--ignore` can be used as a crude way to handle `provides`. See https://github.com/AladW/aurutils/issues/837. Then `aur repo -ql` is preferable over `aur repo -qu`

VorpalBlade:
As far as I can see there is only one issue remaining, and that is what do do about the undocumented (but not hidden) `--buildscript` flags to `aur srcver` and `aur build`.

AladW:
`aur-srcver(1)`: https://github.com/AladW/aurutils/commit/4feddcf9d12032523ad7eda48775e972705af86e
`aur-repo(1)`: https://github.com/AladW/aurutils/commit/4feddcf9d12032523ad7eda48775e972705af86e
`aur-build(1)`: https://github.com/AladW/aurutils/pull/993/commits/16f28102928e6f57e408049b576f62c39a3c5a0e (#993)

VorpalBlade:
I don't know anything else left to do. It is large enough that I would be surprised if there *aren't* any bugs left. But nothing I have run into in daily use, nor in focused testing.

As for possible improvements:
* It would be nice to be able to complete the comma separated lists of arguments to makepkg. And that wouldn't even be difficult (using the `_sequence` helper bundled with zsh). Except `_makepkg` is an lazy loaded function defined in a file that is not named the same as the function (it is defined in `_pacman`). I don't know of a way to ensure it would be available when the user has not yet tab completed any command from the `_pacman` file.

The `_arguments` helper that I use throughout the file has a... somewhat dense? syntax. In addition the official documentation is verbose and dense. Tutorials and looking at existing code is a much easier way to get into it. The general form is however not bad: `<optional-repeat-marker>(<exclusivity-spec>)<flag>[<help-string>]:<argument description>:<completer/action for argument>`. The difficulty comes from all the details and variations (such as a leading space in front of the argument completer having a semantic meaning!). I'm happy to answer any questions you may have though, and I expect to be around and able to help for the foreseeable future.

In addition, different users will have zsh configured differently (that is one of the key selling points of zsh after all!). For example, I use zsh4humans where the argument description is generally not shown (since it conflicts with some other features: I get a tmux window popup with a fuzzy search of the completions available using `fzf`). On the grml config (used by the Arch live CD for example) it is shown however. This obviously makes testing a bit annoying.

AladW:
Thanks, the effort is much appreciated. I'll give it a whirl soon and then merge it for the next release. Sorry for the late reply.

Regarding possible improvements, I wonder how to handle multiple repositories best. That is, `aur repo -lq` with no further arguments assumes a single local repository, and will error out if multiple are available (asking for disambiguation). 

What complicates matters is that the completion output then depends on the order of arguments, i.e. if a local repository was specified with `-d` prior to, say, --ignore, then `aur repo -d ... -lq` is proper. If not (--ignore was specified after `-d` or `-d` was not specified at all) then the completion command would fail with more than one local repository.

VorpalBlade:
> [...] I wonder how to handle multiple repositories best. That is, aur repo -lq with no further arguments assumes a single local repository, and will error out if multiple are available (asking for disambiguation).

I have not thought about this use case.

> What complicates matters is that the completion output then depends on the order of arguments

It is of course possible to to custom code and parsing of the command line. Still if ignore is passed before the user specifies the repo I don't see a solution other than error out gracefully with a `_message` (something that may or may not be displayed to the user depending on zsh settings, I don't see them in my setup for example).

Right now I will not be able to look into this though, I have come down with a quite nasty flu.

AladW:
Merged!
img

Issue #995, opened 2022-05-22
`search`: duplicate package name with `NO_COLOR`


cyrinux:
Hi,

with aurutils-git, `NO_COLOR=true` is broken.

the package name is duplicate.

```
$ NO_COLOR=true aur search -n -k NumVotes arch-secure-boot
aur/arch-secure-bootaur/arch-secure-boot 1.3.2-1 (+2 0.00%)
    UEFI Secure Boot for Arch Linux + btrfs snapshot recovery

$ aur search -n -k NumVotes arch-secure-boot
aur/arch-secure-boot 1.3.2-1 (+2 0.00%)
    UEFI Secure Boot for Arch Linux + btrfs snapshot recovery
```

Regards

Earnestly:
Seems like missing newline in no_color codepath

cyrinux:
Hmm, i'm using kitty, with color enable, i can see the get the link when mouse over `aur/arch-secure-boot` and with NO_COLOR=true i can get the link on the second duplicate `aur/arch-secure-boot`.

Earnestly:
Yes, you should it's always used.  Not quite sure why it's only doubling the name, as it's not missing elements that would otherwise interfere with printf(1)'s implicit looping feature.

AladW:
The issue appears to be that the package name is printed a second time when `ALL_OFF` is undefined i.e. if `colorize()` was not run. The function should return in that case:
```bash
if [[ ! -v ALL_OFF ]]; then
    printf "%s" "$mesg"
    return  # <---
fi
```
That said, I'm not sure `NO_COLOR` should disable the link since it's only about "color" escapes (which are e.g. annoying when debugging with `set -x`).

AladW:
Fixed with #996

PR #996, opened 2022-05-24
search: print package name once with NO_COLOR

AladW:
Issue #995

Alternative: always print the hyperlink, even with NO_COLOR=1. The `hyperlink` function is then reduced to:
```bash
hyperlink() {
    local OSC8="\e]8" ST="\e\\"
    printf "$OSC8;;%s$ST%s$OSC8;;$ST" "$1" "$2"
}
```

### DIFF
--- a/lib/aur-search
+++ b/lib/aur-search
@@ -13,13 +13,8 @@ query_args=()
 
 hyperlink() {
     local uri=$1 mesg=$2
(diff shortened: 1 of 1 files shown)


AladW:
Maybe not, but at least in the alternative example it identifies the escape sequences.

PR #1001, opened 2022-05-31
search: pass color escapes through printf

cruegge:
On dumb terminals (e.g. Emacs shell-mode), `message.sh` does not set the color escapes via `tput`, so they don't contain acutal `0x1b` bytes, but `\e` strings that are expected to be used in `printf` format strings. For hyperlinks, that was not done, leading to literal `\e` in the output.


cruegge:
I pushed a fix for the shellcheck error.

cruegge:
But thinking about it, simply using the `%b` escape in `hyperlink()` to interpret escape sequences would probably also work, and be simpler.

cruegge:
I pushed the alternative version to https://github.com/cruegge/aurutils/tree/alternative, if you want to test. Let me know if I should update the PR.

AladW:
I've been trying this with emacs `shell`, but with '%b' I still get:
```
]8;;https://aur.archlinux.org/packages/zyc\aur/zyc]8;;\ 0.2.0-1 (+0 0.00%) 
    A GUI for searching and selecting parts and footprints for use in SKiDL.
```
With '%s':
```
]8;;https://aur.archlinux.org/packages/zyc\\e[1m\e[34maur/\e[0m\e[1mzyc]8;;\ 0.2.0-1 (+0 0.00%) 
    A GUI for searching and selecting parts and footprints for use in SKiDL
```
Maybe OCS8 sequences should be disabled entirely for dumb terminals?

cruegge:
Emacs recently added support for OSC8, but for some reason it's not active by default. `M-:` `(add-to-list 'comint-output-filter-functions #'comint-osc-process-output)` should do it. So I would very much like hyperlink support to stay, either by default or as opt-in using an env var or a command line argument.

AladW:
OK, I'll leave in the OSC8 sequences then. Applied manually with https://github.com/AladW/aurutils/commit/5b243aeb8604eda71ec7b4b34cb391f1168653e1

Thanks for the report :+1:

cruegge:
Thanks.

PR #1008, opened 2022-07-03
fetch: improve compatibility with aurutils <=9.6

AladW:
* Do not ignore system configuration, i.e. in $HOME/.gitconfig
* Only run git-fetch by default
* Add --auto flag (shorthand for --sync=auto)
* Add -f flag (shorthand for --discard), i.e. aur fetch -Sf
* Remove deprecation notices


AladW:
Hopefully this will be the last rework of `aur-fetch` in a while...

PR #1010, opened 2022-07-10
build: do not use generated configs for elevated commands

AladW:
When using `sudoers(5)` to remove password prompts for select commands, `pacman --config=/tmp/...` command-lines are variable and thus require whitelisting `/usr/bin/pacman` (opening up `sudo pacman -U` and `sudo pacman --config`), or `--config` with arbitrary arguments (e.g. `--config=/tmp/bad.conf`).

Use a different mechanism to limit upgrades to the local repository, which uses fixed command-lines:

1. Propagate new databases with `pacsync custom`
2. Check if any installed packages are to be upgraded with `pacman -Quq`
3. Restrict the targets to the local repository with `pacsift --repo=custom`
4. Finally, install upgrades with `pacinstall --no-confirm`

For replacements, it suffices to use `pacinstall --resolve-replacements=none`. Compare the (unpublished) pacman commit:

https://github.com/eli-schwartz/pacman/commit/80b118b48b38dd55224d33d54867c533c23b0040

WIP/experimental: pacutils 0.11.1 removes prompt messages with `--no-confirm`, and this only really makes sense if the above `makepkg` commit is merged upstream (`aur build -rs`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ ETCDIR ?= /etc
 AURUTILS_LIB_DIR ?= $(LIBDIR)/$(PROGNM)
 AURUTILS_VERSION ?= $(shell git describe --tags || true)
 ifeq ($(AURUTILS_VERSION),)
-AURUTILS_VERSION := 10
+AURUTILS_VERSION := 11
 endif
 
 .PHONY: shellcheck install build completion aur
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -27,12 +27,6 @@ args_csv() {
     printf '%s' "${str%,}"
 }
 
-db_replaces() {
-    bsdcat "$1" | awk '/%REPLACES%/ {
-        while(getline && NF != 0) { print; }
-    }'
-}
-
 diag_moved_packages() {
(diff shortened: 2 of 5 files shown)


AladW:
With a hypothetical `pacsift --upgrades --repo=...` (https://github.com/andrewgregory/pacutils/issues/62), this would be something like:
```bash
    if (( chroot )) || (( no_sync )); then
        continue
    else
        # Propagate database to pacman
        sudo pacsync "$db_name" # custom.db
        sudo pacsync "$db_name" --dbext=.files # custom.files

        # Verify if any host packages can be upgraded
        mapfile -t targets < <(pacsift --repo="$db_name" --upgrades)

        # Restrict upgrades to the local repository
        if (( ${#targets[@]} )); then
            msg >&2 'Upgrading installed local repository packages'

            if ! printf '%s\n' "${targets[@]}" | sudo pacinstall --resolve-replacements=none --no-confirm; then
                error 'Failed to upgrade local repository'
                exit 1
            fi
        fi
    fi
```

Issue #1012, opened 2022-07-14
`build`: one way of reducing sudo waits


greg-minshall:
i've been migrating, and re-migrating, to a new machine.  (i run arch linux, in case that turns out to be relevant.)  i've seen various discussions here of how to deal with `sudo` (like #907).

one approach i'm playing with now is to do installation of the AUR packages in a many-phased fashion (using a Makefile).  the input to the whole thing is, of course, all the AUR packages i need from the previous machine.  i then use `aur depends` (or the meta file) to produce the transitive closure of `depends` and `makedepends`, as well as whatever `pacman` packages are needed.  after than, i start my phased install; certainly this list isn't optimal, but here are my current (or, almost current) phases.

0. download all the packages (just the `PKGBUILDs`, whatever `aur-fetch` does)
1. [should do] make sure all keys listed in any `validpgpkeys` in the downloaded `PKGBUILDs` are installed (using `gpg` `--locate-keys` and `--recv-keys`)
2. install all `pacman` packages needed by any of our packages (including `depends`, `makedepends`)
3. [messy -- building and `sudo`]  if any packages that are `makedepends` *themselves* have `makedepends`, build and synchronize those latter `makedepends` to `pacman` (`sudo`) and then `pacman -S` (`sudo`) them  (trying to split this into build/install phases seems problematic, given that some of *these* packages may themselves have `makedepends`)
4. process the remaining `makedepends`:
  - build all the remaining `makedepends`
  - [`sudo`] install these packages by synchronizing the repository to `pacman`, and then `pacman -S` them
5. process the rest of the packages
  - build all the remaining packages
  - [`sudo`] install these packages by synchronizing the repository to `pacman`, and then `pacman -S` them

i know this looks like a hack, but it *seems* to work (in the sense of reducing the number of `sudo` prompts, and, if a `sudo` prompt times out, getting back to the prompt is fairly quick [except maybe in step 2?]), at least on my fairly simple system.  BUT, to make it work, i had to add two new flags to `aur-sync`, one of which needed to be added to `aur-build`.

- `--no-sync`: this flag is added to `aur-sync` and, if specified, passed on to `aur build`, which prevents `aur-build` from (oh, i see it's changed in `10`; congrats on, and thanks for, the release) doing any `pacsync` or `pacman` operations for its packages.  [it is weird to call `aur sync --no-sync`, but `aur-sync` provides a lot of functionality to download and build packages.]
[...]

PR #1014, opened 2022-07-20
lib: add aur-build--sync helper

AladW:
Instead of prefixing several pacsync/pacman commands with `sudo`, wrap the commands in an `build--sync` helper. When configurating `sudoers` rules, this is more granular than blanket `pacsync` and `pacman` rules: only upgrading local repository targets is whitelisted.

Compared to the previous approach, `--pacman-conf` does not apply to local repository upgrades. This makes it harder to inadvertendly shoot yourself in the foot by passing on a potentially world-writeable file to a privileged process.

A helper can also be overridden through `AUR_EXEC_PATH`, and called by other scripts that wish to replicate a restrained local repository upgrade.

----
The idea of a separate helper to upgrade the local repository originates from issue #1012.

### DIFF
--- a/lib/aur-build--sync
+++ b/lib/aur-build--sync
@@ -1,62 +1,32 @@
-#!/bin/perl -T
-# This script can be used in NOPASSWD rules (sudoers) or similar (doas/setuid
-# wrappers) to upgrade packages in a local repository without password prompt.
-# Use basic measures from perlsec(1perl) to limit security impact.
-use strict;
-use warnings;
-use v5.20;
-my $argv0 = 'build--sync';
+#!/bin/bash
+# build--sync - helper for upgrading local repository
+set -e
+[[ -v AUR_DEBUG ]] && set -o xtrace
+argv0=build--sync
+PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
 
-# minimal path
-$ENV{PATH} = "/bin:/usr/bin";
-delete @ENV{qw(IFS CDPATH ENV BASH_ENV)};
+if (( ! $# )); then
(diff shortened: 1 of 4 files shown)


lib/aur-build--sync
```diff
@@ -0,0 +1,62 @@
...
+# This script can be used in NOPASSWD rules (sudoers) or similar (doas/setuid
+# wrappers) to upgrade packages in a local repository without password prompt.
+# Use basic measures from perlsec(1perl) to limit security impact.
+use strict;
+use warnings;
+use v5.20;
+my $argv0 = 'build--sync';
+
+# minimal path
+$ENV{PATH} = "/bin:/usr/bin";
```
Earnestly, line 11: It may not be necessary here but there is a `getconf(3)` attribute called `PATH` or `_CS_PATH` (as used by `command -p`) which is supposed to represent the default system `PATH` where all POSIX utilties can be found.

AladW: > It may not be necessary here but there is a getconf(3) attribute called PATH or _CS_PATH (as used by command -p) which is supposed to represent the default system PATH where all POSIX utilties can be found.

`perl` has a `POSIX` module, so it might be queried this way.

> However consider that many of these security practices are something the sudo/doas tool should be responsible for and so if a user configures those to allow certain things it might not be nice to undermine them later; unintuitively.

Good point. I could introduce a `--reset-env` option or (conversely) `--preserve-env` to address this.

I initially came to `perl -T` after seeing older versions of `opendoas` [would not reset PATH](https://github.com/advisories/GHSA-59xc-99wg-3hgf). `perl -T` forces you to think on which data is trusted and which isn't, and for that alone I find it useful.

AladW: In the end I went with a simple `bash` script and left things to the user.

PR #1016, opened 2022-07-28
Support extending zsh completion for third party commands

VorpalBlade:
This adds support for extending zsh completion for third party aur commands. I have tested it with my own local `aur-rm`.

Some of the helpers (completing packages and repositories) have been made into auto loadable standalone functions so that they are accessible by third party completion scripts. This defines a new stable API and might thus need some thought about if that would present a problem.

### DIFF
--- a/completions/Makefile
+++ b/completions/Makefile
@@ -1,9 +1,14 @@
 PREFIX ?= /usr
 SHRDIR ?= $(PREFIX)/share
-.PHONY = bash install-bash zsh install-zsh
+.PHONY = bash install-bash install-zsh
+
+ZSH_SITE_FUNCTIONS = \
+	zsh/_aur \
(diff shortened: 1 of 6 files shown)


AladW:
I'd skip mentioning it's a "stable API". This gives peoples expectations, and they will complain when we have to change it for a technical reason. In contrast, if the API is there and some migration notices are given, people will simply adjust their scripts.

VorpalBlade:
> I'd skip mentioning it's a "stable API". This gives peoples expectations, and they will complain when we have to change it for a technical reason. In contrast, if the API is there and some migration notices are given, people will simply adjust their scripts.

Done. I also reworded the documentation a bit.

AladW:
:+1:

PR #1025, opened 2022-09-05
`chroot`: limit `sudo`-specific flags to `sudo`

### LINKED ISSUE #1024: `sudo` flag `--preserve-env` cannot be used in alternative `AUR_PACMAN_AUTH`
While aurutils supports alternative command prefixes to run commands as root via `AUR_PACMAN_AUTH` the `aur-chroot` command contains the sudo-specific [`--preserve-env` flag](https://github.com/AladW/aurutils/blob/5c53970b5480ec670ee0ba27640e8d1aac45b842/lib/aur-chroot#L172) that prevents alternative commands from being used defeating the purpose of `AUR_PACMAN_AUTH`. Is there any workaround for this issue? As far as I can tell this is the only instance in aurutils where a sudo-specific flag is used so it would be nice if it could be avoided somehow.

AladW:
Fixes #1024

### DIFF
--- a/lib/aur-build--sync
+++ b/lib/aur-build--sync
@@ -1,7 +1,9 @@
 #!/bin/bash
+# shellcheck disable=SC2086
 # build--sync - helper for upgrading local repository
-set -e
+set -o errexit
 [[ -v AUR_DEBUG ]] && set -o xtrace
+AUR_PACMAN_AUTH=${AUR_PACMAN_AUTH:-sudo}
 argv0=build--sync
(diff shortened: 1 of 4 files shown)


AladW:
As it turns out, `sudoers` can set which variables are preserved set per command:
```
Defaults!/usr/bin/makechrootpkg env_keep += "PKGDEST GNUPGHOME"
```
Perhaps more disturbingly is the following from `sudoers(5)`

> **If the command matched is ALL, the SETENV tag is implied for that command;** this default may be overridden by use of the NOSETENV tag.

In other words `aur-chroot` has an implicit dependency on `ALL`, unless an explicit rule is added for `/usr/bin/makechrootpkg`. At minimum this should be documented in `aur-chroot(1)`.

PR #1054, opened 2023-03-15
`srcver`: accept arbitrary paths to PKGBUILD directories

### LINKED ISSUE #1052: `srcver`: match `build` command-line
Both `aur-build` and `aur-srcver` assume a top-level directory with `PKGBUILD` subdirectories.  `aur-srcver` accepts these subdirectories as arguments on the command-line. 

`aur-build` is more flexible however in that it accepts an argument file with (relative or absolute) paths, with running from the current directory (`$PWD`) as a special case. `aur-srcver` should match this behavior if possible. 

Example use case: running `aur-srcver` to update `pkgver` in the same fashion as `aur build --pkgver --dry-run` in aurutils v10.

AladW:
Closes #1052


AladW:
`aur-srcver(1)` needs an update as well

PR #1064, opened 2023-05-09
Add support to pass --no-sync option to aur-build command

SuperBFG7:
This enables the usage of aur-sync as normal user without sudo permission.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -112,14 +112,14 @@ fi
 opt_short='d:D:k:U:AcfLnorRSTuv'
 opt_long=('bind:' 'bind-rw:' 'database:' 'directory:' 'ignore:' 'root:'
           'makepkg-conf:' 'pacman-conf:' 'chroot' 'continue' 'force'
-          'ignore-arch' 'log' 'no-confirm' 'no-ver' 'no-graph' 'no-ver-argv'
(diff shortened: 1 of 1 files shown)


AladW:
`aur-sync` uses `aur-build --syncdeps` to install `dependencies`, which requires those dependencies to be visible to `pacman`. `--no-sync` prevents that from happening. Maybe you're looking for `--no-build` instead.

SuperBFG7:
I do want to build the package, but skip the sync of the new package to the local DB after the build.
My use case is: running `aur sync -u --no-check` unattended to regularly update the packages, so they will be available for the next system upgrade.
Yes, it will fail if a package introduces a new dependency, but most of the time it will work and I don't have to give passwordless sudo access to the build user.

Are you sure about the conflict with --syncdeps? To me it seems the --syncdeps is independent and continues to work even with the --no-sync option passed to aur-build.

AladW:
> Yes, it will fail if a package introduces a new dependency, but most of the time it will work and I don't have to give passwordless sudo access to the build user.

If you have package A that depends on B, and B gets a new version, how would you build package A against the newly updated B without pacman sync?

In any event, it's a small change with a use case, so it looks fine to merge.

SuperBFG7:
Thank you for merging it.
The dependency case will indeed fail as well. I haven't thought about this, as I only use a few AUR packages and don't really have dependencies between them (until now).

PR #1074, opened 2023-05-22
aur-chroot.1: fix typo

abouvier:
The path `/var/lib/aurbuild/root` does not exist. But maybe `/var/lib/aurbuild/<arch>/root` is more appropriate?


AladW:
The man page still mentions `i686`, so `<arch>` sounds better than `x86_64`

AladW:
:+1:

PR #1100, opened 2023-06-26
`sync-rebuild`: add `--vcs`

AladW:
This allows to regenerate `.SRCINFO` for packages with a `pkgver` function, so that a correct comparison between `.SRCINFO` and local repository package can be made.


AladW:
The approach is wrong, the `pkgver()` function needs to run and `makepkg --printsrcinfo` does not do this. Use `aur-srcver` instead

PR #1116, opened 2023-09-29
`build--pkglist`: load PKGBUILD by absolute path

### LINKED ISSUE #1115: `build--pkglist`: bogus behavior if anything in $PATH contains a PKGBUILD
```
$ grep pkgname= PKGBUILD
pkgname=sccache-git

$ aur build--pkglist
/home/operator/pkgbuild/sccache-git/sysz-1.4.3-0-any.pkg.tar.zst
```

This happens because I've got a directory in $PATH that happens to contain a (non-executable) PKGBUILD, and `aur-build--pkglist` does this:

https://github.com/AladW/aurutils/blob/98d3b58f91f3c1be33d9264f9caf7496c4d38ada/lib/aur-build--pkglist#L59

In contrast, makepkg does this ([src](https://gitlab.archlinux.org/pacman/pacman/blob/0dfe5c96/scripts/makepkg.sh.in#L1135-1138)):

```sh
	if [[ ${BUILDFILE:0:1} != "/" ]]; then
		BUILDFILE="$startdir/$BUILDFILE"
	fi
	source_buildfile "$BUILDFILE"
```

Might be worth doing something similar, just to reduce the average WTF/hour of the world :-)

intelfx:
This avoids doing the wrong thing if a PKGBUILD happens to be somewhere in $PATH, in which case a `source PKGBUILD` will load it instead of the one in $PWD.

Fixes #1115.

### DIFF
--- a/lib/aur-build--pkglist
+++ b/lib/aur-build--pkglist
@@ -53,10 +53,13 @@ while true; do
     shift
 done
 
+# source follows $PATH, compute absolute path to PKGBUILD (#1115)
+buildscript="$(realpath --strip -- "${buildscript-PKGBUILD}")"
+
 # Sourcing the PKGBUILD should be done without set -e or other modes
 # to match makepkg behavior (e.g. aur/cemu, aur/nicotine-plus-git)
 # shellcheck disable=SC1090
-source_safe "${buildscript-PKGBUILD}"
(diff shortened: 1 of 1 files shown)


AladW:
I'd use `realpath` instead of some heuristic copied from `makepkg`

PR #1117, opened 2023-10-04
readme: Better explain where to find setup instructions

frazar:
First of all, thank you for your work! I use `aurutils` every day.

I wanted to contribute to smooth the adoption process for newcomers. The previous `USAGE` section of `README.md` had a few minor shortcomings (in my opinion):

1. It links to the source code of the [`aur(1)`](man1/aur.1) man page, which is hardly readable. 
2. It refers to `aur(1)` for installation instructions, but these are reported in a section at the very end of the page. Therefore, they can be missed easily.

This commit tries to address the above, by;

1. Adding the explicit bash command to open the manual.
2. Adding the name of the manual section where the instructions are located, so that they can be found easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,11 @@ Upgrade notices are posted to the
 
 ## USAGE
 
-Documentation is included as groff manuals. [`aur(1)`](man1/aur.1) contains a
-general overview, instructions on creating a local repository, and
-several examples.
+Documentation is included as [man pages](https://wiki.archlinux.org/title/Man_page)
+with groff typesetting. They provide a general overview of the various utilities
+with several examples.
+Detailed instructions on how to set up a local repository can be found in the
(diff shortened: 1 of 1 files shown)


AladW:
Point 2. is fine, but I don't think basic `man` usage is something for the aurutils `README` to explain. 

It would be nice if there were some HTML rendering though. Unfortunately it can't be done on github: https://github.com/github/markup/pull/1196#issuecomment-716809366

frazar:
I agree with your observation. But, to be honest, it was not immediately clear to me that "groff manual" meant "the man page". 

Do you want me to proceed with removing the example `man aur` command?

AladW:
Sure. Optionally you could add a link to https://wiki.archlinux.org/title/Man_page for the unlikely event someone didn't hear about man pages when stumbling upon the aurutils README.

frazar:
Great! Should be ready for merging now.

AladW:
Looks great, thank you.

Issue #909, opened 2022-02-05
chroot: replace devtools functionality


AladW:
Without having to rely on `devtools` and its various quirks, `aur-chroot` could be extended in multiple ways. For example:

* Handle `CacheDir` reliably (#809)
* Move the complete build process to the container (GPG keyrings included), instead of retrieving sources on the host
* Support `overlayfs` (#705)
* Support custom `systemd-nspawn` arguments

For a prototype, see https://github.com/AladW/aurutils/blob/chroot-standalone/lib/aur-chroot

PR #1125, opened 2023-12-03
`build`, `chroot`: add `-I <pkg>`, and also `aur build --cargs` just in case

### LINKED ISSUE #1124: `chroot`, `build`: add a flag to install a package into the build chroot
It would be nice if `aur build` (and, by extension, `aur chroot`) had a way to pass an `-i <pkg>` flag to makechrootpkg (yes, `aur chroot` has `--cargs`, but `aur build` has nothing).

This could be useful to install extra packages (e. g. ccache or sccache) into the temporary chroot when `-c -T` is used.

intelfx:
Add support for the `-I|--install <pkg>` flag to both `aur build` and `aur chroot` and pass this flag through to `makechrootpkg` (in makechrootpkg, it is a newly added flag `-i`).

Furthermore, add `aur build --cargs` (and pass this flag through to `aur chroot`) just in case there are future `makechrootpkg` flags we will not have exposed.

Depends on https://gitlab.archlinux.org/archlinux/devtools/-/merge_requests/214.

Fixes #1124.

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -86,7 +86,7 @@ opt_long=('arg-file:' 'chroot' 'database:' 'force' 'root:' 'sign' 'gpg-sign'
           'makepkg-conf:' 'bind:' 'bind-rw:' 'prevent-downgrade' 'temp'
           'syncdeps' 'clean' 'namcap' 'checkpkg' 'makepkg-args:' 'user:'
           'margs:' 'buildscript:' 'null' 'dbext:' 'install:' 'cleanbuild'
-          'cargs:' 'makechrootpkg-args:')
+          'cargs:' 'makechrootpkg-args:' 'cargs-no-default' 'makechrootpkg-args-no-default')
 opt_hidden=('dump-options' 'ignorearch' 'noconfirm' 'nocheck' 'nosync' 'repo:'
             'results:' 'results-append:' 'status')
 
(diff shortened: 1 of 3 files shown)


intelfx:
Welp -- nevermind, `makechrootpkg -I` does a totally different thing, namely installs a package _file_...

I guess devtools needs a new flag for this usecase

intelfx:
Alright, this now depends on https://gitlab.archlinux.org/archlinux/devtools/-/merge_requests/214.

lib/aur-build
```diff
@@ -147,6 +148,8 @@ while true; do
             chroot_args+=(--temp) ;;
         -I|--install)
 			shift; chroot_args+=(--install "$1") ;;
+		--cargs|--makechrootpkg-args)
```
AladW, line 151: The indentation seems off here

AladW:
https://github.com/aurutils/aurutils/pull/1125/commits/c8658e5fd581789869101f1c5306909315f67d57 and https://github.com/aurutils/aurutils/pull/1125/commits/47d9b36ed2baf129e7db68c91b8653c8cefdcfd0 look fine for merging though (formatting aside). The `changelog` should also be updated.

intelfx:
@AladW yeah, assuming I won't be able to convince devtools maintainers -- if you are sympathetic to the use-case described here, maybe this could be the final straw to reimplement makechrootpkg functionality within aurutils? What do you think?

intelfx:
I can go through #909 and copy-write mkarchroot and makechrootpkg as e. g. `aur-chroot--create` and `aur-chroot--build`, applying points from #909 where relevant, and then we can turn the rest of it into separate issues...

AladW:
I don't think it makes sense to copy makechrootpkg and mkarchroot entirely, since I don't want to copy all of the legacy design and implementation issues. Some ideas can be recycled though. See the `chroot-standalone` branch.

AladW:
Since a devtools replacement is in the works (#1155), I think this PR can be closed in favor of that one.

PR #1133, opened 2024-01-25
Fix formating of aur-view(1) manual

FranklinYu:
The `.BR` macro seems to eliminate all spaces between words.


AladW:
Thanks :+1:

PR #1137, opened 2024-02-12
`sync`: early `chroot` check

AladW:


abouvier:
I get this:
```
$ aur sync --chroot --database=myrepo cantata
Error:
    aur-chroot could not find a pacman.conf(5) file for container usage. Before
    using aur-chroot, make sure this file is created and valid. See OPTIONS in
    aur-chroot(1) for configuration details.
    
    The following file paths were checked:
        '/etc/aurutils/pacman-x86_64.conf'
        '/usr/share/devtools/pacman.conf.d/aurutils-x86_64.conf'
```
It seems that `/etc/aurutils/pacman-myrepo.conf` is not even checked, because `--suffix myrepo` is not sent to `aur-chroot` (only `aur-build` send this to `aur-chroot` apparently).

abouvier:
You should add `chroot_args+=(--suffix "$db_name")` somewhere :p

abouvier:
Even after these modifications, the suffix is still not transmitted to `aur-chroot`.

abouvier:
Oops sorry, I was using the old patch. It's working now ;)

PR #1155, opened 2024-05-03
Replace devtools functionality

chrooti:
This is an attempt at solving (part of) https://github.com/aurutils/aurutils/issues/909.

This PR takes inspiration from the branch linked in the issue and a few arch projects to replace devtools with pure systemd-nspawn.

A few things:
- --namcap and --checkpkg will need a separate implementation (should it be done in this PR?)
- I've left a few XXX comments where I thought there could be improvements
- I absolutely don't think this is ready for merge in the current state, I'm looking for pointers on what I've already done in order to know if something is fundamentally wrong/missing

### DIFF
--- a/lib/aur-build
+++ b/lib/aur-build
@@ -52,6 +52,10 @@ get_local_upgrades() {
     return "${PIPESTATUS[0]}"
 }
 
+auth_cmd() {
+    ${AUR_PACMAN_AUTH:-sudo --preserve-env} "$@"
+}
+
 trap_exit() {
     if [[ ! -v AUR_DEBUG ]]; then
         rm -rf -- "$tmp"
@@ -122,8 +126,7 @@ while true; do
             shift; pacman_conf=$1
             chroot_args+=(--pacman-conf "$1") ;;
         --pkgver)
-            run_pkgver=1; makepkg_args+=(--noextract)
-            chroot_build_args+=(--margs '--holdver') ;;
+            run_pkgver=1; makepkg_args+=(--noextract) ;;
         --root)
             shift; db_root=$1
             repo_args+=(--root "$1") ;;
@@ -141,13 +144,14 @@ while true; do
             shift; chroot_args+=(--user "$1") ;;
         -N|--namcap)
             chroot_args+=(--namcap) ;;
+        # XXX: remove this if and when aur chroot--standalone is made standard
         --checkpkg)
             chroot_args+=(--checkpkg) ;;
+        # XXX: remove this if and when aur chroot--standalone is made standard
         -T|--temp)
             chroot_args+=(--temp) ;;
         # makepkg options (common)
         -A|--ignorearch|--ignore-arch)
-            chroot_build_args+=(--ignorearch)
             makepkg_common_args+=(--ignorearch) ;;
         -n|--noconfirm|--no-confirm)
             makepkg_common_args+=(--noconfirm) ;;
@@ -163,7 +167,6 @@ while true; do
         -L|--log)
             makepkg_args+=(--log) ;;
         --nocheck|--no-check)
-            chroot_build_args+=(--nocheck)
             makepkg_args+=(--nocheck) ;;
         # XXX: cannot take arguments that contain commas (e.g. for file paths)
         --makepkg-args|--margs)
@@ -196,6 +199,13 @@ while true; do
     esac
     shift
 done
+
+# If --pkgver is enabled, VCS sources are updated with makepkg -od.
+# Therefore they do not need to be updated a second time with makepkg
+# --verifysource in makechrootpkg.
+if (( chroot )) && (( run_pkgver )); then
+    makepkg_args+=(--holdver)
+fi
 
 # Assign environment variables
 db_ext=${db_ext:-$AUR_DBEXT} db_name=${db_name:-$AUR_REPO} db_root=${db_root:-$AUR_DBROOT}
@@ -331,7 +341,7 @@ if (( chroot )); then
     # Update pacman and makepkg configuration for the chroot build
     # queue. A full system upgrade is run on the /root container to
     # avoid lenghty upgrades for makechrootpkg -u.
-    aur chroot "${chroot_args[@]}" --create --update
+    auth_cmd aur "${AUR_BUILD_CHROOT:-chroot--standalone}" "${chroot_args[@]}" --create --update
 fi
 
 # Early check for `makepkg` buildscript
@@ -385,7 +395,8 @@ while IFS= read "${read_args[@]}" -ru "$fd" path; do
 
     if (( create_package )); then
         if (( chroot )); then
-            PKGDEST="$var_tmp" aur chroot "${chroot_args[@]}" --build "${chroot_build_args[@]}"
+            PKGDEST="$var_tmp" AUR_CHROOT_UID=$(id -u) \
+                auth_cmd aur "${AUR_BUILD_CHROOT:-chroot}" "${chroot_args[@]}" --build "${chroot_build_args[@]}" -- "${makepkg_common_args[@]}" "${makepkg_args[@]}"
         else
             #shellcheck disable=SC2086
             PKGDEST="$var_tmp" LOGDEST="${LOGDEST:-$PWD}" \
@@ -443,9 +454,9 @@ while IFS= read "${read_args[@]}" -ru "$fd" path; do
         continue
     else
         #shellcheck disable=SC2086
-        ${AUR_PACMAN_AUTH:-sudo} pacsync "${sync_args[@]}" "$db_name"
+        auth_cmd pacsync "${sync_args[@]}" "$db_name"
         #shellcheck disable=SC2086
-        ${AUR_PACMAN_AUTH:-sudo} pacsync "${sync_args[@]}" "$db_name" --dbext=.files
+        auth_cmd pacsync "${sync_args[@]}" "$db_name" --dbext=.files
 
         # Retrieve upgrade targets in local repository. May error in case of
         # conflicts or dependency errors.
@@ -455,7 +466,7 @@ while IFS= read "${read_args[@]}" -ru "$fd" path; do
         if (( ${#targets[@]} )); then
             printf >&2 "%s: upgrading packages in repository '%s'\n" "$argv0" "$db_name"
             #shellcheck disable=SC2086
-            printf '%s\n' "${targets[@]}" | ${AUR_PACMAN_AUTH:-sudo} pacman "${sync_args[@]}" -S --noconfirm -
+            printf '%s\n' "${targets[@]}" | auth_cmd pacman "${sync_args[@]}" -S --noconfirm -
         fi
     fi
 done
--- a/lib/aur-chroot--standalone
+++ b/lib/aur-chroot--standalone
@@ -1,0 +1,346 @@
+#!/bin/bash
+# aur-chroot - build packages with systemd-nspawn
+[[ -v AUR_DEBUG ]] && set -o xtrace
+set -o errexit
+argv0=chroot
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+machine=$(uname -m)
+
+# default arguments
+directory=/var/lib/aurbuild/$machine
+makepkg_args=()
+makepkg_user=build
+
+# default options
+update=0 build=0 create=0 status=0 run_namcap=0
+
+args_csv() {
+    # shellcheck disable=SC2155
+    local str=$(printf '%s,' "$@")
+    printf '%s' "${str%,}"
+}
+
+# XXX: a missing makepkg.conf usually indicates a missing devtools, whereas
+# a missing pacman.conf usually indicates the local repository was not configured
+diag_makepkg_conf() {
+    echo >&2 'Error:'
+
+    cat <<EOF | pr -to 4 >&2
+aur-$argv0 could not find a makepkg.conf(5) file for container usage. Before
+using aur-$argv0, make sure this file is created and valid. See OPTIONS in
+aur-$argv0(1) for configuration details.
+
+The following file paths were checked:
+EOF
+    printf '%s\n' "${@@Q}" | pr -to 8 >&2
+}
+
+diag_pacman_conf() {
+    echo >&2 'Error:'
+
+    cat <<EOF | pr -to 4 >&2
+aur-$argv0 could not find a pacman.conf(5) file for container usage. Before
+using aur-$argv0, make sure this file is created and valid. See OPTIONS in
+aur-$argv0(1) for configuration details.
+
+The following file paths were checked:
+EOF
+    printf '%s\n' "${@@Q}" | pr -to 8 >&2
+}
+
+usage() {
+    printf 'usage: %s [-BU] [--create] [-CDM path] [package...]\n' "$argv0"
+    exit 1
+}
+
+opt_short='C:D:M:x:ABNTU'
+opt_long=('directory:' 'pacman-conf:' 'makepkg-conf:' 'build' 'update'
+          'create' 'bind:' 'bind-rw:' 'user:' 'makepkg-args:'
+          'ignorearch' 'namcap' 'checkpkg' 'temp' 'makechrootpkg-args:'
+          'margs:' 'cargs:' 'nocheck' 'suffix:')
+opt_hidden=('dump-options' 'status')
+
+if opts=$(getopt -o "$opt_short" -l "$(args_csv "${opt_long[@]}" "${opt_hidden[@]}")" -n "$argv0" -- "$@"); then
+    eval set -- "$opts"
+else
+    usage
+fi
+
+# _chroot functions are "jitted" in a script inside the chroot and executed there
+# in order to run complex commands and keep things tidy
+# idea comes from https://gitlab.archlinux.org/archlinux/devtools/-/blob/master/src/makechrootpkg.in
+
+_chroot_create() {
+    # XXX: use localedef to generate locales directly (outside of nspawn)
+    locale-gen
+
+    pacman-key --init
+    pacman-key --populate
+}
+
+# https://gitlab.archlinux.org/archlinux/devtools/-/blob/master/src/makechrootpkg.in?ref_type=heads#L235
+_chroot_run_namcap() {
+    sudo pacman -S --needed --noconfirm namcap
+    for pkgfile in PKGBUILD /pkgdest/*; do
+        namcap "$pkgfile" 2>&1 | tee "/logdest/${pkgfile##*/}-namcap.log"
+    done
+}
+
+unset bindmounts_ro bindmounts_rw makepkg_conf pacman_conf suffix
+while true; do
+    case "$1" in
+        -x|--suffix)
+            shift; suffix=$1 ;;
+        -B|--build)
+            build=1 ;;
+        -U|--update)
+            update=1 ;;
+        --create)
+            create=1 ;;
+        --status)
+            status=1 ;;
+        # makepkg options (`--build`)
+        -A|--ignorearch)
+            makepkg_args+=(--ignorearch) ;;
+        --nocheck)
+            makepkg_args+=(--nocheck) ;;
+        # XXX: cannot take arguments that contain commas (e.g. for file paths)
+        --makepkg-args|--margs)
+            shift; IFS=, read -a arg -r <<< "$1"
+            makepkg_args+=("${arg[@]}") ;;
+        -C|--pacman-conf)
+            shift; pacman_conf=$1 ;;
+        -D|--directory)
+            shift; directory=$1 ;;
+        -M|--makepkg-conf)
+            shift; makepkg_conf=$1 ;;
+        --bind)
+            shift; bindmounts_ro+=("$1") ;;
+        --bind-rw)
+            shift; bindmounts_rw+=("$1") ;;
+        -N|--namcap)
+            run_namcap=1 ;;
+        # XXX: remove this if and when aur chroot--standalone is made standard
+        --checkpkg)
+            printf >&2 'The --checkpkg option is deprecated in the new aur chroot standalone implementation:'
+            printf >&2 ' running checkpkg as part of aur chroot will not be supported anymore'
+            ;;
+        # XXX: remove this if and when aur chroot--standalone is made standard
+        -T|--temp)
+            printf >&2 'The -T|--temp option is deprecated in the new aur chroot standalone implementation:'
+            printf >&2 ' there is no more difference between temporary and non-temporary chroots.'
+            ;;
+        --user)
+            shift; makepkg_user=$1 ;;
+        # XXX: remove this if and when aur chroot--standalone is made standard
+        --makechrootpkg-args|--cargs)
+            shift
+            printf >&2 'The --makechrootpkg-args|--cargs option is deprecated in the new aur chroot standalone implementation:'
+            printf >&2 ' makechrootpkg is not used as a backend anymore.'
+            ;;
+        --dump-options)
+            printf -- '--%s\n' "${opt_long[@]}" ${AUR_DEBUG+"${opt_hidden[@]}"}
+            printf -- '%s' "${opt_short}" | sed 's/.:\?/-&\n/g'
+            exit ;;
+        --) shift; break ;;
+    esac
+    shift
+done
+
+# XXX: default paths can be set through the Makefile (`aur-chroot.in`)
+etcdir=/etc/aurutils shrdir=/usr/share/devtools
+
+# The pacman configuration in the chroot may contain a local repository that
+# is not configured on the host. Therefore, $db_name is only used for the
+# default paths below when specified on the command-line or through `AUR_REPO`.
+if [[ -v suffix ]]; then
+    default_pacman_paths=("$etcdir/pacman-$suffix.conf"
+                          "$etcdir/pacman-$machine.conf"
+                          "$shrdir/pacman.conf.d/$suffix.conf"
+                          "$shrdir/pacman.conf.d/aurutils-$machine.conf")
+
+    default_makepkg_paths=("$etcdir/makepkg-$suffix.conf"
+                           "$etcdir/makepkg-$machine.conf"
+                           "$shrdir/makepkg.conf.d/$suffix.conf"
+                           "$shrdir/makepkg.conf.d/$machine.conf")
+else
+    default_pacman_paths=("$etcdir/pacman-$machine.conf"
+                          "$shrdir/pacman.conf.d/aurutils-$machine.conf")
+
+    default_makepkg_paths=("$etcdir/makepkg-$machine.conf"
+                           "$shrdir/makepkg.conf.d/$machine.conf")
+fi
+
+# Change the default /usr/share/devtools/pacman-extra.conf in aur-chroot to
+# /etc/aurutils/pacman-<repo>.conf or /etc/aurutils/pacman-<uname>.conf in
+# aur-build, and pass it on to aur-chroot (#824, #846)
+for def in "${default_pacman_paths[@]}"; do
+    if [[ -f $def ]] && [[ ! -v pacman_conf ]]; then
+        pacman_conf=$def
+        break
+    fi
+done
+
+# The same as above but for /etc/aurutils/makepkg-<repo>.conf or
+# /etc/aurutils/makepkg-<uname>.conf. If the file is not found, fallback to
+# makepkg.conf files in /usr/share/devtools.
+for def in "${default_makepkg_paths[@]}"; do
+    if [[ -f $def ]] && [[ ! -v makepkg_conf ]]; then
+        makepkg_conf=$def
+        break
+    fi
+done
+
+# No pacman configuration is available for the container, or it points to a
+# non-existing file. Print a matching diagnostic and exit.
+if [[ ! -v pacman_conf ]]; then
+    diag_pacman_conf "${default_pacman_paths[@]}"
+    exit 2
+elif [[ ! -f $pacman_conf ]]; then
+    diag_pacman_conf "$pacman_conf"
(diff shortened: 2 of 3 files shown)


lib/aur-chroot
```diff
@@ -49,6 +49,17 @@ EOF
     printf '%s\n' "${@@Q}" | pr -to 8 >&2
 }
 
+add_mount() {
```
AladW, line 52: All of these calls should already be handled by systemd-nspawn

lib/aur-chroot
```diff
@@ -205,36 +211,73 @@ while read -r key _ value; do
...
     # specified on the command-line.
     # available packages depend on the configured pacman configuration
     if (( $# )); then
         base_packages=("$@")
-
-    # XXX: use pacini to not process Include directives in pacman.conf
-    # (not supported by devtools)
     elif [[ $(pacini --section=multilib "$pacman_conf") ]] && [[ $machine == "x86_64" ]]; then
-        base_packages=('base-devel' 'multilib-devel')
+        base_packages=('base' 'base-devel' 'multilib-devel')
```
AladW, line 230: Chroots only assume base-devel, not base

lib/aur-chroot
```diff
@@ -205,36 +211,73 @@ while read -r key _ value; do
...
+if (( create )) && [[ ! -d $directory/root ]]; then
     # default to base-devel or multilib-devel, unless packages are
     # specified on the command-line.
     # available packages depend on the configured pacman configuration
     if (( $# )); then
         base_packages=("$@")
-
-    # XXX: use pacini to not process Include directives in pacman.conf
-    # (not supported by devtools)
     elif [[ $(pacini --section=multilib "$pacman_conf") ]] && [[ $machine == "x86_64" ]]; then
```
AladW, line 229: Since we don't care about devtools now, we could use `pacconf` instead of `pacini`

lib/aur-chroot
```diff
@@ -260,12 +304,34 @@ if (( print_path )); then
...
+        --setenv=PKGDEST=/pkgdest \
+        --setenv=LOGDEST=/logdest \
+        --setenv=SRCDEST=/srcdest \
+        --setenv=SRCPKGDEST=/srcpkgdest \
+        --setenv=BUILDDIR="$PWD" \
+        --setenv=MAKEPKG_CONF="$makepkg_conf" \
+        --setenv=SOURCE_DATE_EPOCH="${SOURCE_DATE_EPOCH:-$(date +%s)}" \
+        "${bindmounts_ro[@]/#/--bind-ro=}" \
+        "${bindmounts_rw[@]/#/--bind=}" \
+        makepkg --cleanbuild --force --syncdeps --noconfirm "${makepkg_args[@]}" -- "$@"
```
AladW, line 334: I wouldn't hardcode any makepkg arguments here. It's worth considering to even leave out the `makepkg` command entirely, so that people could use `aurutils` installed inside the chroot, e.g. `aur chroot aur build ...`

lib/aur-chroot
```diff
@@ -205,36 +211,73 @@ while read -r key _ value; do
...
-        # shellcheck disable=SC2086
-        $AUR_PACMAN_AUTH install -d "$directory" -m 755 -v
-    fi
+    # create directories for pacman
 
-    if [[ ! -d $directory/root ]]; then
-        # shellcheck disable=SC2086
-        $AUR_PACMAN_AUTH mkarchroot -C "$pacman_conf" -M "$makepkg_conf" "$directory"/root "${base_packages[@]}"
-    fi
+    $AUR_PACMAN_AUTH install -d "$directory"/root -m 755 -v
```
AladW, line 237: There's so many AUR_PACMAN_AUTH calls here, maybe the entire script should run/rexec as root, similar to what devtools does.

lib/aur-chroot
chrooti: Agree about the makepkg arguments. About removing makepkg I'm a bit concerned: wouldn't this radically change the semantics of `aur chroot` (and make a bunch of flags useless?), are you okay with that?

lib/aur-chroot
chrooti: I tried, but it looks like I have to add a lot of envvar to preserve_env. Not sure if that's worth it.

lib/aur-chroot
chrooti: I implemented this by using pacman-conf (less dependencies) even though I don't know the differences between the two.

AladW: The main difference is that pacconf has `--raw` ($repo and $arch are not replaced, standard values are not included in the output). But it might not be a requirement for `aur-chroot`

lib/aur-chroot
AladW: A possible approach is to not do it in the script (and limit the elevation tool to `sudo --preserve-env`), but rexec with `$AUR_PACMAN_AUTH env ... aur-chroot`. Thoughts?

AladW:
The changes look up. To allow for a transition period, perhaps make it `aur-chroot-standalone` at first?

lib/aur-chroot
AladW: Yeah, maybe changing the semantics could be left for a future PR. For syncdeps, aur-build also does not include those flags by default. They could propagated if say, the user does `aur-build --syncdeps`

lib/aur-chroot
```diff
@@ -260,12 +267,32 @@ if (( print_path )); then
...
+
+    printf "u %q $(id -u) \"makepkg user\" /build /bin/bash\n" "$makepkg_user" | \
+        $AUR_PACMAN_AUTH systemd-sysusers --root "$directory"/root --replace=/usr/lib/sysusers.d/build.conf -
+
+    $AUR_PACMAN_AUTH systemd-nspawn \
+        -D "$directory"/root \
+        --chdir="$PWD" \
+        --user="$makepkg_user" \
+        --volatile=overlay \
+        --setenv=PKGDEST=/pkgdest \
```
AladW, line 286: These variables should match bindmounts_rw, eg. `--setenv=PKGDEST=${PKGDEST:-$PWD}:/pkgdest"`. The variables could be assigned in the script to avoid repetition.

chrooti: Does makepkg support colon separated paths in environment variables? How does it work? Does it use the first one that exists? Nice to know btw (couldn't find it in the documentation)

The reason why I implemented it like this is that:
- I need to bind mount all of PKGDEST LOGDEST SRCDEST SRCPKGDEST
- I don't know which ones are going to be pointing at the same directory (by default $PWD is used for all 4)
- So I mount all the paths (which might be the same path 4 times) in 4 different fixed locations and pass those to makepkg

Sorry for the mansplaining :) But given that, why do I have to pass the path outside the chroot to the makepkg variables too?

lib/aur-chroot
chrooti: I think I'm missing something: whatever we do we need to keep track of the variables anyway and keep the list updated. I can't spot the difference between using `env` or `$AUR_PACMAN_AUTH` directly (in this case).

chrooti:
With the last commit I think I've made the whole `--margs` mechanism redundant (at least in `aur chroot`). I haven't removed it yet for backward compatibility but it might be worth it (it's not a huge change).

I agree with renaming it into aur chroot standalone (I'll do it as a last step). I'd like your input on:
- what to do with namcap/checkpkg? Should I keep them commented? Remove them entirely from `aur chroot`?
- how would a user decide if `aur-chroot` or `aur-chroot--standalone` should be used from `aur build`? envvar?

AladW:
* `namcap` seems simple enough to implement, but AFAICT `checkpkg` has no relevance to an AUR repo setup. If someone wants it, they could make a request.
* An envvar like `AUR_BUILD_CHROOT` could work, provided `aur-chroot` and `aur-chroot-standalone` expose the same interface on the command-line.

lib/aur-chroot
AladW: That was a mistake on my behalf: it should be `--setenv=PKGDEST=${PKGDEST:-$PWD}`. In any case, `aur-build` sets PKGDEST to a temporary directory, so that needs to be propagated to `makepkg` (if possible, to avoid manually moving packages like `devtools` does.)

chrooti: Sorry, I feel like an idiot: I'm setting environment variables to directories with fixed names because e.g. `/pkgdest` is the path *inside* the container (i.e. the mount target), the mappings are added to the `bindmounts_rw` variable just above the `systemd-nspawn` call.
Then, the path *inside* the container will always be /pkgdest, but outside the container the location will be the correct one (i.e. PWD or the PKGDEST variable)
I have actually tested the integration with `aur build` and it looks like its working as intended, please enlighten me :D

AladW:
That sounds reasonable. In line with the other naming, I'd call it `aur-chroot--helper` (two hyphens at the end).

lib/aur-chroot
chrooti: I went and implemented the `sudo --preserve-env` and rexec with a list of variables. At least there is only one place where we need to keep track of that :)

chrooti:
I went for a slightly different solution inspired once again by makechrootpkg: `declare -f` as a poor man's git. The mechanism is convoluted but the script code itself is still quite readable (IMHO).

lib/aur-build
```diff
@@ -341,9 +341,8 @@ fi
 
 if (( chroot )); then
     # Update pacman and makepkg configuration for the chroot build
-    # queue. A full system upgrade is run on the /root container to
-    # avoid lenghty upgrades for makechrootpkg -u.
-    aur chroot "${chroot_args[@]}" --create --update
+    # queue.
+    $auth_cmd aur "${AUR_BUILD_CHROOT:-chroot--standalone}" "${chroot_args[@]}" --create --update
```
Earnestly, line 345: Can `$auth_cmd` be turned into a function instead?

lib/aur-build
```diff
@@ -397,8 +396,11 @@ while IFS= read "${read_args[@]}" -ru "$fd" path; do
 
     if (( create_package )); then
         if (( chroot )); then
-            PKGDEST="$var_tmp" aur "${AUR_BUILD_CHROOT:-chroot--standalone}" \
-                "${chroot_args[@]}" --build "${chroot_build_args[@]}" -- "${makepkg_common_args[@]}" "${makepkg_args[@]}"
+            PKGDEST="$var_tmp" \
```
Earnestly, line 399: I don't like these column fixed \ line continuations, is it really necessary to so strongly wrap longer commands?

lib/aur-chroot
```diff
@@ -296,16 +312,39 @@ if (( build )); then
...
     # allow makepkg_user to install packages (e.g. --syncdeps)
-    $AUR_PACMAN_AUTH tee "$directory/root/etc/sudoers.d/makepkg_user" \
+    tee "$directory/root/etc/sudoers.d/makepkg_user" \
         < <(printf '%q ALL = NOPASSWD: /usr/bin/pacman' "$makepkg_user") \
         1> /dev/null
-    $AUR_PACMAN_AUTH chmod 440 "$directory/root/etc/sudoers.d/makepkg_user"
+    chmod 440 "$directory/root/etc/sudoers.d/makepkg_user"
+
+    {
+        printf 'makepkg'
```
Earnestly, line 325: Fwiw:
```sh
printf 'makepkg %s || exit\n' "${*@Q}"` > "$directory"/root/aur-chroot--helper
```

chrooti: There had to be a way, thanks for showing me! :)

lib/aur-build
chrooti: can you give me an example of how you would format this one (and perhaps others)?

Earnestly: I wouldn't both with automated systems. It's up to alad really but I personally tend to just keep the command as long as it needs to be.

chrooti: For the time being I indented it as the line immediately below (1 line for envvars, 1 line for the rest)
Is it an acceptable tradeoff for you? Also waiting to hear from alad on this.

AladW: I have no strong opinion about this, but the tradeoff seems fair enough.

AladW:
`declare -f` seems reasonable for now, especially when the script needs to live inside the chroot.

lib/aur-chroot
```diff
@@ -124,6 +147,9 @@ while true; do
     shift
 done
 
+printf >&2 $'You\'re using a new version of aur-chroot.\n'
```
Earnestly, line 150: In diagnostic messages try to avoid contractions. Consider something like `Set AUR_BUILD_CHROOT=chroot if experiencing problems with this newer implementation of aur-chroot.`

However it seems like this kind of message which will eventually be removed in time so I wonder if it's needed at all in the code instead of being part of the release notes.

lib/aur-chroot
```diff
@@ -103,18 +120,24 @@ while true; do
         --bind-rw)
             shift; bindmounts_rw+=("$1") ;;
         -N|--namcap)
-            makechrootpkg_args+=(-n) ;;
+            run_namcap=1 ;;
+        # XXX: remove this if and when aur chroot--standalone is made standard
         --checkpkg)
-            makechrootpkg_args+=(-C) ;;
+            echo \
```
Earnestly, line 126: Use `printf` instead of `echo`

lib/aur-chroot
```diff
@@ -184,88 +210,147 @@ elif [[ ! -f $makepkg_conf ]]; then
...
+    rm "$directory"/root/aur-chroot--helper
+}
+
+trap cleanup EXIT
+
 # create new container, required for update/build steps
-if (( create )); then
+if [[ ! -d $directory/root ]]; then
+    if (( !create )); then
+        printf >&2 '%s: %q is not a directory\n' "$argv0" "$directory"/root
```
Earnestly, line 248: Don't use `%q` as it can encode the argument in a manner which will not be identical to what the user provided causing confusion.

lib/aur-chroot
```diff
@@ -184,88 +210,147 @@ elif [[ ! -f $makepkg_conf ]]; then
...
-
-    if [[ ! -d $directory/root ]]; then
-        # shellcheck disable=SC2086
-        $AUR_PACMAN_AUTH mkarchroot -C "$pacman_conf" -M "$makepkg_conf" "$directory"/root "${base_packages[@]}"
-    fi
+    # # create both the root container and a directory for pacman db
+    install -d "$directory"/root/var/lib/pacman -m 755
+
+    # # XXX: allow setting machine-id from host (and provide default) to enable .nspawn file usage
+    systemd-firstboot \
```
Earnestly, line 268: Are the line continuations really necessary here?

lib/aur-chroot
```diff
@@ -184,88 +210,147 @@ elif [[ ! -f $makepkg_conf ]]; then
...
+    printf 'pacman -Syu --noconfirm --config %q %s || exit' "$pacman_conf" "${*@Q}" >> "$directory"/root/aur-chroot--helper
 fi >&2
 
-# print path for processing by other tools (e.g. makepkg --packagelist)
-if (( print_path )); then
-    realpath -- "$directory"/root
-    exit $?
-fi
+if (( create )) || (( update )); then
+    systemd-nspawn \
```
Earnestly, line 291: Again, why can't this just be one line?

lib/aur-chroot
```diff
@@ -184,88 +210,147 @@ elif [[ ! -f $makepkg_conf ]]; then
...
+        "$PWD"
+        "${PKGDEST:-$PWD}:/pkgdest"
+        "${LOGDEST:-$PWD}:/logdest"
+        "${SRCDEST:-$PWD}:/srcdest"
+        "${SRCPKGDEST:-$PWD}:/srcpkgdest"
+    )
+
+    # create makepkg_user
+    systemd-sysusers --root "$directory"/root --replace=/usr/lib/sysusers.d/build.conf - \
+        < <(printf "u %q %s \"makepkg user\" /build /bin/bash\n" "$makepkg_user" "$AUR_CHROOT_UID")
```
Earnestly, line 312: Take advantage of literal quoting when it's appropriate:
```sh
printf 'u %q %s "makepkg user" /build /bin/bash\n'
```

lib/aur-chroot
```diff
@@ -184,88 +210,147 @@ elif [[ ! -f $makepkg_conf ]]; then
...
+
+    # allow makepkg_user to install packages (e.g. --syncdeps)
+    tee "$directory/root/etc/sudoers.d/makepkg_user" \
+        < <(printf '%q ALL = NOPASSWD: /usr/bin/pacman' "$makepkg_user") \
```
Earnestly, line 316: `%q` is a form of encoding that produces strings according to bash rather than sudo. The `User` grammar for sudo doesn't allow single quotes but uses similar mechanism for escaping problem characters:

Consider the `sudoers` manual:

```
User_List ::= User |
              User ',' User_List

User ::= '!'* user name |
         '!'* #user-ID |
         '!'* %group |
         '!'* %#group-ID |
         '!'* +netgroup |
         '!'* %:nonunix_group |
         '!'* %:#nonunix_gid |
         '!'* User_Alias

[...]

A user  name,  user-ID,  group,  group-ID,  netgroup,  nonunix_group  or
nonunix_gid  may  be enclosed in double quotes to avoid the need for es‐
caping special characters.  Alternately, special characters may be spec‐
ified in escaped hex mode, e.g., \x20  for  space.   When  using  double
quotes, any prefix characters must be included inside the quotes.

[...]

Quotes around group names are optional.  Unquoted  strings  must  use  a
backslash  (‘\’)  to  escape  spaces and special characters.  See “Other
special characters and reserved words” for a  list  of  characters  that
need to be escaped.
```

Here quotes are `"` which `%q` won't use, but it might use `\`. So I would be careful about mixing encoding schemes and making sure they accord to one another.

(N.B. sudoers doesn't seem to say how to handle `"` inside a quoted string, whether `\"` or closing, escaping and re-opening is needed (`"\""`))

lib/aur-chroot
chrooti: > However it seems like this kind of message which will eventually be removed in time so I wonder if it's needed at all in the code instead of being part of the release notes.

I've changed the code according to the first part of this comment. On the second part though I feel like this is a pretty big change and nobody really reads changelog. One line sounds like a fair price to avoid N issues with people complaining their build broke.

lib/aur-chroot
chrooti: thanks for taking the time to explain this!

lib/aur-chroot
AladW: Usually with these kind of major changes I put a message in the `.install` file of the (AUR) PKGBUILD and write some announcement on the forums and AUR comments. That being said, I'd still keep the old `aur-chroot` as default for the time being. Then if it's been sufficiently been tested by people using chroot, it can be made the default.

lib/aur-build
```diff
@@ -329,9 +339,8 @@ fi
 
 if (( chroot )); then
     # Update pacman and makepkg configuration for the chroot build
-    # queue. A full system upgrade is run on the /root container to
-    # avoid lenghty upgrades for makechrootpkg -u.
-    aur chroot "${chroot_args[@]}" --create --update
+    # queue.
```
AladW, line 342: Was removing the second part of the comment intentional?

makepkg/PKGBUILD
```diff
@@ -15,7 +15,6 @@ depends=('git' 'pacutils' 'curl' 'bash' 'perl' 'perl-json-xs')
 makedepends=('git')
 optdepends=('bash-completion: bash completion'
             'zsh: zsh completion'
-            'devtools: aur-chroot'
```
AladW, line 18: I'd only remove this when `aur-chroot--standalone` becomes the default implementation

man1/aur-chroot.1
```diff
@@ -144,20 +144,10 @@ are checked in order:
 .br
 .BI /usr/share/devtools/makepkg.conf.d/<machine>.conf
 .
-.SS makechrootpkg options
```
AladW, line 147: I'd leave the man page unchanged until `chroot--standalone` becomes the default implementation

completions/zsh/_aur
```diff
@@ -48,7 +48,6 @@ _aur_build_sync_args=(
     '(-n --no-confirm)'{-n,--no-confirm}'[do not wait for user input when installing or removing build dependencies]'
     '(-r --rmdeps)'{-r,--rmdeps}'[remove dependencies installed by makepkg]'
     '(-S --sign --gpg-sign)'{-S,--sign,--gpg-sign}'[sign build packages and the database with gpg]'
-    '(-T --temp)'{-T,--temp}'[build in a temporary container]'
```
AladW, line 51: I'd leave in the completions until `chroot--standalone` becomes the default implementation

AladW:
Could you rebase this to one commit for `aur-build` changes, and another for the `aur-chroot--standalone` addition?

lib/aur-build
chrooti: since we're still using the old chroot I'm going to add it back

chrooti:
I made one extra for aur-sync, hopefully it's alright.

chrooti:
@AladW sorry for pinging but I see you have answered issues and opened a PR... are you still interested in merging this?

AladW:
I'm still interested, I just got some deadlines that prevent me from looking into more complex PRs. Will give it a look in the next week or two.

lib/aur-build
```diff
@@ -385,7 +395,8 @@ while IFS= read "${read_args[@]}" -ru "$fd" path; do
 
     if (( create_package )); then
         if (( chroot )); then
-            PKGDEST="$var_tmp" aur chroot "${chroot_args[@]}" --build "${chroot_build_args[@]}"
+            PKGDEST="$var_tmp" AUR_CHROOT_UID=$(id -u) \
+                auth_cmd aur "${AUR_BUILD_CHROOT:-chroot}" "${chroot_args[@]}" --build "${chroot_build_args[@]}" -- "${makepkg_common_args[@]}" "${makepkg_args[@]}"
```
AladW, line 399: See L129; the additional arguments after `--` will be interpreted by the existing `aur--chroot` as arguments for `base_packages`.

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+    # XXX: use localedef to generate locales directly (outside of nspawn)
+    locale-gen
+
+    pacman-key --init
+    pacman-key --populate
+}
+
+# https://gitlab.archlinux.org/archlinux/devtools/-/blob/master/src/makechrootpkg.in?ref_type=heads#L235
+_chroot_run_namcap() {
+    sudo pacman -S --needed --noconfirm namcap
```
AladW, line 84: `sudo` is redundant, since `aur-chroot--helper` is already run as root

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+    for pkgfile in PKGBUILD /pkgdest/*; do
+        namcap "$pkgfile" 2>&1 | tee "/logdest/${pkgfile##*/}-namcap.log"
```
AladW, line 86: `makechrootpkg` also uses the build user to run namcap, with latter requiring no additional privileges.

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+        printf >&2 '%s: %s is not a directory\n' "$argv0" "$directory"/root
+        printf >&2 '%s: did you run aur chroot --create?\n' "$argv0"
+        exit 20
+    fi
+
+    # default to base-devel or multilib-devel, unless packages are
+    # specified on the command-line.
+    # available packages depend on the configured pacman configuration
+    if (( $# )); then
+        base_packages=("$@")
```
AladW, line 255: The the comment for `aur-build` on L129; this logic assumes makepkg arguments are specified with `--margs`, not specified after `--`.

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+        base_packages=('base-devel')
+    fi
+
+    # # create both the root container and a directory for pacman db
+    install -d "$directory"/root/var/lib/pacman -m 755
+
+    # # XXX: allow setting machine-id from host (and provide default) to enable .nspawn file usage
+    systemd-firstboot --root="$directory"/root --copy-locale --copy-timezone --setup-machine-id
+
+    pacman -Sy "${base_packages[@]}" \
```
AladW, line 268: I would put the `pacman -Sy` arguments on one line here

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+    bindmounts_rw+=(
+        "$PWD"
+        "${PKGDEST:-$PWD}:/pkgdest"
+        "${LOGDEST:-$PWD}:/logdest"
+        "${SRCDEST:-$PWD}:/srcdest"
+        "${SRCPKGDEST:-$PWD}:/srcpkgdest"
+    )
+
+    # create makepkg_user
+    systemd-sysusers --root "$directory"/root --replace=/usr/lib/sysusers.d/build.conf - \
```
AladW, line 301: Maybe it makes sense to use a second helper for the `--build` step, or overwrite (`>`) the contents. As it stands, the commands added for the `--create` and `--update` steps would be run a second time when both `--build` and `--create` / `--update` are specified.

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+        < <(printf 'u %s %s "makepkg user" /build /bin/bash\n' "$makepkg_user" "$AUR_CHROOT_UID")
```
AladW, line 302: `AUR_CHROOT_UID` seems like a good idea, but as written this relies on it being set by the caller, i.e. `aur-build`. When running `aur-chroot` on its own, the value is empty. I don't know if this is handled implicitly by `systemd-sysusers`, or if some explicit value should be set.

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+
+    # allow makepkg_user to install packages (e.g. --syncdeps)
+    tee "$directory/root/etc/sudoers.d/makepkg_user" \
```
AladW, line 305: `cat` also expands shell arguments:
```bash
    cat <<EOF >"$directory/root/etc/sudoers.d/makepkg_user"
$makepkg_user ALL = NOPASSWD: /usr/bin/pacman
EOF
```

lib/aur-chroot--standalone
```diff
@@ -0,0 +1,346 @@
...
+        < <(printf '%s ALL = NOPASSWD: /usr/bin/pacman' "$makepkg_user") \
+        1> /dev/null
+    chmod 440 "$directory/root/etc/sudoers.d/makepkg_user"
+
+    printf 'makepkg %s || exit\n' "${*@Q}" >> "$directory"/root/aur-chroot--helper
```
AladW, line 310: `makepkg_args` instead of `${*@Q}`

AladW (changes requested):

lib/aur-build
```diff
@@ -52,6 +52,10 @@ get_local_upgrades() {
     return "${PIPESTATUS[0]}"
 }
 
+auth_cmd() {
+    ${AUR_PACMAN_AUTH:-sudo --preserve-env} "$@"
```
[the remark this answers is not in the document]
chrooti: I'm not follwoing: `aur build` is the one calling `aur chroot` and has some prior infrastructure to call various commands with an `auth_cmd.` Once I call it with `sudo` without `--preserve-env` I can't do anything to get the envvars back.
I see a few ways of doing what you suggest:
- _not_ calling aur chroot with sudo and having a separate logic to sudo only the relevant commands. This is what I did in my first implementation, and you [suggested](https://github.com/aurutils/aurutils/pull/1155#discussion_r1594574587) to change it. I'm fine with changing it again, however.
- adding a separate `auth_cmd_chroot` function and the respective envvar inside `aur build`

Any third option I haven't thought about?

lib/aur-chroot--standalone
chrooti: I think this is what we do here... do I have to change anything?

lib/aur-chroot--standalone
chrooti: as long as we call aur chroot with sudo externally this needs to be passed, and I think the safest approach is just to raise if this variable doesn't exist. What do you think?

lib/aur-chroot--standalone
chrooti: I just disliked the weird indentation and it seemed a useless use of cat, but it's a matter of taste. I'll change it.

lib/aur-build
```diff
@@ -122,8 +126,7 @@ while true; do
             shift; pacman_conf=$1
             chroot_args+=(--pacman-conf "$1") ;;
         --pkgver)
-            run_pkgver=1; makepkg_args+=(--noextract)
-            chroot_build_args+=(--margs '--holdver') ;;
+            run_pkgver=1; makepkg_args+=(--noextract) ;;
```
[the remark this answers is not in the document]
chrooti: I thought you didn't particularly like the `--margs` interface with the comma as separator and it felt like a good moment for breaking changes, more so because there is still --create/--update to add more packages to the environment.
Just put a thumbs up and I'll do what you asked, no biggie.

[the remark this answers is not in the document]
AladW: The challenge is when combining `--build` with `--create` / `--update`. Then the additional arguments could be both `makepkg` arguments, and package names. With the end-of-options approach, to pass on additional packages to be installed in the chroot you'd need an additional flag like `--extra-packages`. Since that approach gives some special meaning for `--` when `--margs` doesn't, I think `--margs` is preferable overall.

lib/aur-chroot--standalone
AladW: Wouldn't `namcap` run as root here?

lib/aur-build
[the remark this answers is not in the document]
AladW: The discussion you linked mentions `check_root` in `aur-chroot` to reexec, but it seems that's no longer part of the script?

AladW:
My responses are still delayed, sorry about that. I still hope to see this in v20.

lib/aur-chroot--standalone
AladW: Let's return to this then after solving https://github.com/aurutils/aurutils/pull/1155#pullrequestreview-2230853262

AladW:
Superseded by #1179.

PR #1156, opened 2024-05-07
docs: ccache with aur-chroot needs --bind-rw

VannTen:
Otherwise you'd get 'read-only filesystem' errors when trying to build.


AladW:
Thanks.

PR #1157, opened 2024-05-08
man: fix glob in example

ammgws:
It should be a recursive glob in order to get all the .SRCINFO files


Earnestly:
The example assumes your directory structure is the more typical case of having a top level full of package directories with their `PKGBUILD` and `.SRCINFO` files directly below.

AladW:
Right. Considering the specific nature of the example I'd leave it as is.

ammgws:
Sorry about that, I have my pkgbuild repo setup the same but you're right - the glob should work as-is, not sure what I was doing last night

PR #1170, opened 2024-06-29
repo: warn when local repo is not configured

AladW:
In #1110 a mandatory check was added for the local repository to be configured in pacman.conf. However, the bulk of the aur-repo functionality is available without configuration in pacman.conf, provided the repository root (`--root`) is specified by the user.

@BrainDamage

### DIFF
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -186,14 +186,16 @@ if [[ ! $db_name ]]; then
 
 # check $db_name is a configured pacman repository (local or remote, #1113)
 elif ! [[ ${conf_sync_serv[$db_name]} ]]; then
-    printf >&2 "%s: repository %q not configured\n" "$err_prefix" "$db_name"
-    exit 2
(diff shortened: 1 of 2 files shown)


lib/aur-repo
```diff
@@ -186,14 +186,16 @@ if [[ ! $db_name ]]; then
 
 # check $db_name is a configured pacman repository (local or remote, #1113)
 elif ! [[ ${conf_sync_serv[$db_name]} ]]; then
-    printf >&2 "%s: repository %q not configured\n" "$err_prefix" "$db_name"
-    exit 2
+    printf >&2 "warning: repository %q not configured\n" "$db_name"
```
Earnestly, line 189: May want to take the opportunity here to replace `%q` with `%s`.

AladW: Made the change. There's some other places left with `%q` that might be worth a look later.

PR #1171, opened 2024-07-18
`repo`: add `--init`

### LINKED ISSUE #1168: `repo-add` does not create an empty database
I recently did a fresh install of Arch and could no longer create an empty local repository following the steps in `man aur`.

After some digging I came across a change made in the upstream `repo-add` script ([f91fa546](https://gitlab.archlinux.org/pacman/pacman/-/commit/f91fa546f65af9ca7cdbe2b419c181df609969b7)) that causes it to not create a database if it is used without specifying any packages to add.

A quick workaround was simply adding the `aurutils` package (although it could work with any other AUR package) to the database after the package was built manually using `makepkg`, i.e.

```
sudo install -d /home/custompkgs -o $USER
repo-add /home/custompkgs/custom.db.tar.gz /path/to/aurutils.pkg.tar.zst
```

This has since been fixed in [6dddb1f7](https://gitlab.archlinux.org/pacman/pacman/-/commit/6dddb1f78cc1728d3842bc9b5e12ed9b0ff6a2a5?merge_request_iid=183) but hasn't been merged yet.

I thought mentioning the issue here will help others until the fix is merged or maybe there is an alternative solution that does not require adding any prebuilt packages to create the database.

AladW:
Workaround for a regression in `pacman` where `repo-add` no longer creates empty databases.

Closes #1168

### DIFF
--- a/lib/aur-repo
+++ b/lib/aur-repo
@@ -21,7 +21,7 @@ usage() {
 # option parsing
 opt_short='c:f:d:i:r:s:F:almqtuJS'
 opt_long=('config:' 'database:' 'root:' 'all' 'list' 'path' 'list-path' 'list-repo'
-          'search:' 'search-by:' 'list-attr' 'sync' 'upgrades' 'table' 'quiet' 
+          'search:' 'search-by:' 'list-attr' 'sync' 'upgrades' 'table' 'quiet' 'init'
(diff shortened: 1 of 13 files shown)


AladW:
This should no longer be needed with pacman 7.0.

PR #1179, opened 2024-08-29
Replace devtools functionality (part II)

AladW:
Follow-up to #1155.


AladW:
The option parsing and configuration logic is duplicated from `aur-chroot`, so I'm thinking to split off the `create`/`update` and `build` parts to `aur-chroot--create` and `aur-chroot--build`, respectively. Then `aur-chroot` would call them if `AUR_CHROOT_STANDALONE` is set.

AladW:
I just merged the implementation directly to `aur-chroot` directly. About 120 lines for all we use from devtools is a decent trade-off.

Now this just needs some more testing and it can be merged.

PR #1180, opened 2024-09-09
`sync`: automatic `gnupg` key retrieval

### LINKED ISSUE #1129: `sync`: automatic key retrieve
Compared to the `auto-key-retrieve` gpg option, the idea is to import package keys to a separate keyring, with location defined by `AUR_SYNC_GNUPGHOME`. A partial implementation is in https://github.com/aurutils/aurutils/commit/c7624040c2781c07b7af7a4b490db21de8a4389d. What remains is passing this variable to `makepkg`, but not the `gpg` commands in `aur-build` (used for signing build artifacts with the user keyring).

AladW:
This uses a separate gpg keyring for signature verification with `makepkg(8)`. Since this will not affect the default keyring, the behavior is enabled by default.

Closes #1129.

### DIFF
--- a/lib/aur-sync
+++ b/lib/aur-sync
@@ -6,8 +6,10 @@ argv0=sync
 XDG_CACHE_HOME=${XDG_CACHE_HOME:-$HOME/.cache}
 XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/.config}
 XDG_STATE_HOME=${XDG_STATE_HOME:-$HOME/.local/state}
+XDG_DATA_HOME=${XDG_DATA_HOME:-$HOME/.local/share}
 AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/$argv0}
 AUR_SYNC_USE_NINJA=${AUR_SYNC_USE_NINJA:-0}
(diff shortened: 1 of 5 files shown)


lib/aur-sync
```diff
@@ -386,6 +390,28 @@ else
     xargs -a "$tmp"/queue stat >/dev/null || exit 2 # ensure all directories are available
 fi
 
+if (( auto_key_retrieve )); then
+    # shellcheck disable=SC2174
+    mkdir -pm 0700 -- "$AUR_SYNC_GNUPGHOME"
+    declare -A keys_uniq
+
+    # Retrieve unique set of gpg keys to be imported
+    while IFS= read -r path; do
```
Earnestly, line 399: Mostly because this isn't nul delimited but perhaps a good idea anyway to check if `"$path"/.SRCINFO` exists before attempting to use it.

``` bash
while IFS= read -r path; do
    if [[ -f $path/.SRCINFO ]]; then
        mapfile -t < <(pacini "$path"/.SRCINFO 'validpgpkeys')

        for key in "${MAPFILE[@]}"; do
            keys_uniq[$key]=1
        done
    fi
done < "$tmp"/queue
```

PR #1181, opened 2024-09-12
chroot: unbreak `--path`

intelfx:
Avoid triggering the implied `--status` when `--path` is given.

Fixes: 3c282892 ("chroot: make --status default if no modes are specified")

### DIFF
--- a/lib/aur-chroot
+++ b/lib/aur-chroot
@@ -186,7 +186,7 @@ fi
 
 # Print paths to container and used makepkg/pacman paths. This does
 # not require a priorly created container.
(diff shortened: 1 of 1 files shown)


AladW:
#1179 will remove `--path` anyway, is it still needed?

intelfx:
Oh.

I guess not then, feel free to close the PR. But I wish there was (much) less churn in aurutils command interfaces and semantics. It is kinda hard to build stuff on top of aurutils when interfaces come and go and semantics change every few months.

AladW:
I've re-added `--path`, 21 will have enough changes as it stands.

PR #1182, opened 2024-09-23
`lib`: split to subdirectories

AladW:
This PR introduces a subdirectory structure for `lib`. There is no effect on the installed package (the result of running the `Makefile`), and mainly serves to improve overview of the different components. Every subdirectory can have its own documentation with a `README.md` file.

Every subdirectory is self-contained; any programs within only relies on programs from the same directory, or the `AUR` perl library. `aur-sync` uses programs from all subdirectories, and is thus left in the main `lib` directory.


Earnestly (approved):
Seems reasonable to me. Was a bit apprehensive about calling it `pacman` at first but really those tools very are pacman specific.

PR #1183, opened 2024-09-23
`lib`: add `build--sync`

AladW:
This moves the `pacman` database synchronization to a helper script `aur-build--sync`, which can be run independently.

* The proposed implementation uses `pacsync` / `pactrans` with a temporary pacman configuration file. This matches older `aur-build` implementations (see quote below). 

> An interesting side-effect is that pacman considers packages inside the official repositories "local", and warns if they are newer than any custom counterpart.

* Existing sudoers rules may need to be changed.
* Locking is implemented so that `aur-build` instances can be run in parallel, provided package queues do not overlap.


lib/pacman/aur-build--sync
```diff
@@ -0,0 +1,93 @@
...
+
+# option parsing
+opt_short='d:n'
+opt_long=('database:' 'config:' 'noconfirm')
+orig_argv=("$@")
+
+if opts=$(getopt -o "$opt_short" -l "$(args_csv "${opt_long[@]}")" -n "$argv0" -- "$@"); then
+    eval set -- "$opts"
+else
+    printf >&2 'usage: %s [-n] [-d repo]\n' "$argv0"
```
Earnestly, line 56: The code suggests `-d` can be repeated, as it uses `db_names+=("$1")`. If this is the case the way to express that in the usage would be:

```
usage: %s [-n] [-d database]...\n
```

I'm not sure what `"$@"` is being used for in the calls to `pactrans`. That might need to be reflected in the usage as well.

lib/pacman/aur-build--sync
```diff
@@ -0,0 +1,93 @@
...
+else
+    printf >&2 'usage: %s [-n] [-d database]... [pkgname...]\n' "$argv0"
+    exit 1
+fi
+
+while true; do
+    case "$1" in
+        -d|--database)
+            shift; db_names+=("$1") ;;
+        --config)
```
Earnestly, line 64: Would it be better to set a variable `--config) conf=$1`? Then after the option parsing loop add:

```sh
if [[ -f $conf ]]; then
    conf_args+=(--config "$conf")
fi
```

Likewise for `-n|--noconfirm` as it eliminates potential for repeated flags while only accepting the last given option (which is typical convention).

lib/pacman/aur-build--sync
```diff
@@ -0,0 +1,93 @@
...
+# build--sync - helper for upgrading local repository
+set -o errexit
+[[ -v AUR_DEBUG ]] && set -o xtrace
+argv0=build--sync
+PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'
+
+# default options
+conf_args=() sync_args=() db_names=()
+
+args_csv() {
```
Earnestly, line 12: This could be rewritten as

```sh
args_csv() {
    local IFS=,
    printf %s "$*"
}
```

lib/pacman/aur-build--sync
```diff
@@ -0,0 +1,93 @@
...
+}
+
+get_local_config() {
+    #global conf_args
+    local repo
+
+    printf '[options]\n'
+    pacconf "${conf_args[@]}" --options --raw
+
+    for repo in "$@"; do
```
Earnestly, line 25:
```sh
for repo do
    printf '[%s]\n' "$repo"
    pacconf "${conf_args[@]}" --repo="$repo" --raw
done
```

lib/pacman/aur-build--sync
```diff
@@ -0,0 +1,93 @@
...
+if (( EUID != 0 )); then
+    exec ${AUR_PACMAN_AUTH:-sudo} "${BASH_SOURCE[0]}" "${orig_argv[@]}"
+fi
+
+lockfile=$(paclock "${conf_args[@]}" --print)
+
+wait_lock "$lockfile" pacsync "${conf_args[@]}" "${db_names[@]}"
+wait_lock "$lockfile" pacsync "${conf_args[@]}" "${db_names[@]}" --dbext=.files
+
+if [[ ${#db_names[*]} ]]; then
```
Earnestly, line 82: For this check to work you'll need to set `IFS` to nothing, e.g. `IFS=` (or `IFS=''` to make shellcheck happy, if it complains).

lib/pacman/aur-build--sync
AladW: This could be done in a few other places as well, but not everyone might know about the implied "$@"

lib/pacman/aur-build--sync
AladW: `args_csv` is used all over the codebase:
```
source/repos/aurutils/lib % grep -R 'args_csv()'
pacman/aur-build:args_csv() {
pacman/aur-repo-filter:args_csv() {
pacman/aur-repo:args_csv() {
pacman/aur-build--sync:args_csv() {
pacman/aur-chroot:args_csv() {
pacman/aur-srcver:args_csv() {
pacman/aur-vercmp:args_csv() {
util/aur-view:args_csv() {
aur-sync:args_csv() {
aurweb/aur-fetch:args_csv() {
aurweb/aur-pkglist:args_csv() {
```
if we change it, it should be changed consistently

lib/pacman/aur-build--sync
AladW: Specifically package names - although the use for `build--sync` is limited. I updated the usage as follows:
```
printf >&2 'usage: %s [-n] [-d database]... -- [pkgname...]\n' "$argv0"
```

lib/pacman/aur-build--sync
AladW: I went back to `[[ ${db_names[*]} ]]` for now

PR #1192, opened 2025-01-26
sync-asroot: rewrite in Python for more argument flexibility

FichteFoll:
Hard-coding single-letter arguments that are individually forwarded to either the aur-sync or aur-build calls does not scale well and I was missing a few, notably --pkgver for aur-build.

Instead of adding yet another single-letter argument, I wanted to make argument parsing more generic and since I'm not that much into bash, I converted the script into Python and used some special logic to parse multiple sections of arguments that are then forwarded to either the aur-sync call, the aur-build call, or both. Since we already have an example script in Python, I assume this not to be an issue.

Outside of this, functionality remains largely unchanged.

Options `-k` and `-U` are kept since they are required for the wrapper's functionality.

All other options can be specified in either the aur-sync section (started by `--S`) or the aur-build section (started by `--B`). Package names *must* be specified after a `--` separator argument in order to support forwarding unrecognized arguments to both aur sub-commands, e.g. `-d repo`.

```
~ Δ ~/bin/sync-asroot --help
usage: sudo sync-asroot [options] [options for sync and build] {--S [options for sync]} {--B [options for build]} -- {packages}

Batch-sync multiple AUR packages as root.

Example usages:
    sync-asroot -U build -d aur --S --nover --B --pkgver -- pkg1 pkg2
    sync-asroot -k 0 -d aur --S -u

options:
  -h, --help          show this help message and exit
  -U, --user USER     Build as USER; defaults to sudo user.
  -k, --keep-going N  Keep going until N failures occurred.

Any unrecognized options are forwarded to both `aur sync` and `aur build` sub-calls.
Arguments only for the `aur sync` call can be provided by being included after a single `--S` argument.
Arguments only for the `aur build` call can be provided by being included after a single `--B` argument.
Packages must be specified after a final `--`.

See also: aur-build(1), aur-sync(1)
```

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -44,7 +44,7 @@ directories are available. This suggests to use a persistent directory for
 * interacting with an nspawn container with `aur-chroot`.
 
 Instead of elevating to the root user for these tasks, `aur-build` can be run as
-root, dropping privileges where necessary. `sync-asroot` does do by running
+root, dropping privileges where necessary. `sync-asroot` does this by running
 `makepkg`, `gpg`, `repo-add` and `aur-build--pkglist` with `runuser -u <user>`.
 Sources are also retrieved this way with `runuser -u <user> aur sync`.
 
--- a/examples/sync-asroot
+++ b/examples/sync-asroot
@@ -1,71 +1,242 @@
-#!/bin/bash
-# run as: `sudo sync-asroot ...`
-[[ -v AUR_DEBUG ]] && set -o xtrace
-SUDO_HOME=$(eval echo "~$SUDO_USER")
-XDG_CACHE_HOME=${XDG_CACHE_HOME:-$SUDO_HOME/.cache}
-AURDEST=${AURDEST:-$XDG_CACHE_HOME/aurutils/sync}
-argv0=sync-asroot
+#!/usr/bin/env python
 
-# default options
-build_args=(-LR --chroot) sync_args=() keep_going=1
+import argparse
+import contextlib
+import os
+import re
+import shutil
+import subprocess
+import sys
+import textwrap
+from collections.abc import Generator
+from pathlib import Path
 
-# colors (for build summary)
-source /usr/share/makepkg/util/message.sh
 
-if [[ ! -v NO_COLOR ]] && [[ ! -v AUR_DEBUG ]]; then
-    [[ -t 2 ]] && colorize
-fi
+class Params(argparse.Namespace):
+    user: str | None
+    keep_going: int
 
-# option parsing
-unset build_user
-while getopts :U:d:k:fSuN OPT; do
-    case $OPT in
-        d) build_args+=(-d "$OPTARG"); sync_args+=(-d "$OPTARG") ;;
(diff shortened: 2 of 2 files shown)


FichteFoll:
The requirement for packages to be provided after `--` is actually fake since they are technically aur-sync arguments. However, this only surfaced a bit later in the implementation and I figured that specifying packages after a `--` would be more natural than after `--S` and before `--B`.

The requirement of `--S` having to occur before `--B` is more of an implementation detail that I'm also not too happy with, but the code to work around that wouldn't be so sleek. Another idea would be to replace `--S` and `--B` with just `--`, since you basically need to have sync args and build args would then simply be optional afterwards, but that also seemed a bit harder to read since there is now more context attached to `--`.

Suggestions welcome.

Earnestly:
In my opinion your python script is essentially a shell script, but requiring many more lines of code to express the same thing. Many standard library routines are not used, such as calling `mktemp` instead of tempfile.

What you're really after, and the way to properly express passing command-lines within command-lines is to use a repeatable flag and build an array. This is how `mpv` handles `yt-dlp` arguments for example.

Consider:

```sh
while getopts :s:b: opt; do
    case $opt in
    o) sync_flags+=("$OPTARG") ;;
    b) build_flags+=("$OPTARG") ;;
    *) # ...
    esac
done
```

With this you can then capture arguments as `aur-sync -s --nover -b --pkgver` without ever having to worry about parsing command-lines like a shell might. This will handle long option arguments as well, such as `-b --arg-file=foobar` but not `-a foobar`. Much like `mpv` and `yt-dlp` to work with `-a file` you would have to invent a format, e.g. `-b -a=file` where you then parse out `-a` and `file` over `=`. This isn't too bad if needed.

Earnestly:
> This would work if written as `-b -a -b foobar`, wouldn't it? I would hesitate to add `-b --argfile=foobar` syntax for this wrapper script only when it is not supported by `aur-sync` and `aur-build` upstream.

The `-b -a -b foobar` form is essentially dead on arrival as it would need to have prior knowledge of every option that could take an argument for every tool you intend to wrap, essentially reimplementing the cli for every tool, which scales quite poorly and doesn't respond very well to changes. It does give you full control though, because you're somewhat forced to take it.

The `key=value` notation is quite typical in cases like this, a good example is `mpv` and how it handles `--ytdl-raw-options`. I do agree it's not the most ideal but by attempting to essentially tunnel one tool's cli through another cli means you must somehow encode the inner command line in some form that retains its structure; this is imo the simplest.

The next most reasonable approach (from a ui point of view) is to take a string and then implement a parser for all of the shell's quoting syntax, which I don't recommend doing.

I would suspect most users are reasonable enough to understand the need for `--flag=value` for this use-case (mpv even requires `--ytdl-raw-options=flag=` (no value) for flags that do not take an argument; which I dislike).

AladW:
I'm impartial on the argument issue, but to me there's already sufficiently python code in the `examples` directory. Going above that seems worthy of a separate project; see e.g. https://github.com/VorpalBlade/aurutilsutils.

Issue #1196, opened 2025-04-19
`chroot`: preserve `LOGDEST` and `SRCDEST` when running `makechrootpkg`


lahwaacz:
Right now `aur-chroot` preserves only `PKGDEST` when running `makechrootpkg` via `sudo`:

https://github.com/aurutils/aurutils/blob/e5dced1d74649891c07f15b89cb6a30b14f22809/lib/pacman/aur-chroot#L199

However, I may want to explicitly set `LOGDEST` and `SRCDEST` in the environment to temporarily put these artifacts in a different location than what I have set in `~/.config/pacman/makepkg.conf`. I have to work around this by setting `AUR_PACMAN_AUTH="sudo --preserve-env=GNUPGHOME,SSH_AUTH_SOCK,SRCDEST,PKGDEST,LOGDEST"` but this is applied to `mkarchroot` and `arch-nspawn` as well, so I have to reconfigure sudoers.

I believe all `SRCDEST,PKGDEST,LOGDEST` should be preserved by default.

AladW:
Thanks for the report. Fixed in #1197

Toolybird:
Hey, this appears to have caused a regression for me. In `~/.config/pacman/makepkg.conf` I have `LOGDEST=/path/to/my/build/logs` which has worked fine for years. Not any more! Will take a look...

Toolybird:
Yep, just recreated a clean `aurutils` setup in a fresh VM (without any hacks) and sure enough, it's busted. @AladW, could you please take a look?

lahwaacz:
@Toolybird Do you have `LOGDEST` set also in your environment? Otherwise I don't see what would override the value from your `makepkg.conf`...

Toolybird:
> Maybe the LOGDEST="${LOGDEST:-$PWD}" could be removed here?

Yeah, that does appear to fix my issue. But whether it's the right fix in all cases... dunno.

AladW:
Environment variables do override variables defined in `makepkg.conf`
```
	# preserve environment variables to override makepkg.conf
	local restore_envvars=$(
		for var in PKGDEST SRCDEST SRCPKGDEST LOGDEST BUILDDIR PKGEXT SRCEXT GPGKEY PACKAGER CARCH; do
			# the output of 'declare -p' results in locally scoped values when used within a function
			[[ -v $var ]] && printf '%s=%s\n' "$var" "${!var@Q}"
		done
	
```
[...]

PR #1197, opened 2025-04-23
chroot: preserve LOGDEST and SRCDEST by default

AladW:
Issue #1196


lahwaacz:
One last nitpicK: it is also mentioned in the man page :sweat_smile: 

https://github.com/aurutils/aurutils/blob/aa94b0368ed0ac45f43030be4188ec9546a1755f/man1/aur-chroot.1#L204-L213

PR #1212, opened 2025-08-29
aur-fetch: support mixed url and pkgbase names

Earnestly:
Support both pkgbase names and urls given via stdin or on the command-line.

This is done by prefixing pkgbases with AUR_LOCATION and capturing the basename of paths from urls. Each pkgbase and url is then associated together so that the main loop can use the remote url specified for each given package.

This commit introduces a requirement on trurl(1) which is used to accurately parse out url path components.

Addresses https://github.com/aurutils/aurutils/issues/1104

### DIFF
--- a/lib/aurweb/aur-fetch
+++ b/lib/aurweb/aur-fetch
@@ -1,13 +1,37 @@
 #!/bin/bash
 # aur-fetch - retrieve build files from the AUR
-[[ -v AUR_DEBUG ]] && set -o xtrace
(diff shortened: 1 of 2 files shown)


AladW:
Looks pretty good, we just need to fix the `parseopt-consistency` test.

Is it worthwhile to add an EditorConfig file to the repo, to ensure consistent formatting across scripts? (e.g. applied with `shfmt`)

PR #1232, opened 2025-10-21
aur-fetch: guard packages array assignment to avoid empty pkgbase errors

Strykar:
Add check for empty pkgbase before assigning remote.

The simple guard fixes issues when a malformed or empty URL is passed through `trurl`, without changing behavior for valid package URLs.


AladW:
I might replace the process substitution with pipes as suggested here:
https://github.com/aurutils/aurutils/issues/1230#issuecomment-3425906879

PR #1253, opened 2026-04-11
tests: add depends.t with fixture auto-discovery

### LINKED ISSUE #1132: `perl`: improve test coverage
The dependency code (`Depends.pm`) is fairly complicated and there's all kinds of edge cases that should be documented in tests.

Strykar:
## Summary

Test suite for `Depends.pm` covering `recurse()`, `graph()`, `prune()`, `tsort()` — closes #1132.

- **42 tests** (synthetic unit, pipeline integration, fixture invariants)
- Confirms dead code in #1252 (`$pkgdeps->{$name}` always an arrayref)
- Documents self-edge provider redirection bug (explicit target's self-edge rerouted when `$provides=1`)
- `perl/Makefile` uses `prove -I.` so tests run against repo modules

### Fixtures

`perl/t/fixtures/<pkg>/*.json` — dir name is the primary target, all JSON files merged into the mock DB (mirrors aur-depends RPC rounds). Auto-discovered and validated against 6 invariants (self-edge, pipeline survival, no dangling nodes, no foreign in main DAG, tsort completeness, no tsort duplicates).

Included: ruby-activemodel (cyclic checkdeps), python2-cryptography (provides + equality constraints), liri-git-meta (AUR meta-package, replaces the removed ros-noetic-desktop).

### Test plan
- [x] `make test` passes
- [x] `perlcritic --severity 3` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/perl/t/depends.t
+++ b/perl/t/depends.t
@@ -1,0 +1,803 @@
+#!/usr/bin/env perl
+use strict;
+use warnings;
+use v5.20;
+use Carp;
+use File::Basename qw(dirname);
+use File::Temp qw(tempfile);
+use Test::More;
+use AUR::Json qw(parse_json);
+
+# Check if module can be imported
+require_ok "AUR::Depends";
+
+AUR::Depends->import(qw(recurse graph prune tsort));
+
+# --- Helper: build $results hash from a list of packages ---
+# Each entry: { Name => ..., Version => ..., Provides => [...] }
+sub make_results {
+    my @packages = @_;
+    my %results;
+    for my $pkg (@packages) {
+        $results{$pkg->{Name}} = $pkg;
+    }
+    return %results;
+}
+
+# --- Helper: build $pkgdeps like recurse() does ---
+# Seeds Self edges for targets, then adds explicit deps.
+# $targets: arrayref of target names
+# $deps:    hashref  { pkgname => [[$spec, $type], ...] }
+sub make_pkgdeps {
+    my ($targets, $deps) = @_;
+    my %pkgdeps;
+
+    # Seed Self edges the same way recurse() does (line 69-71)
+    for my $t (@{$targets}) {
+        push @{$pkgdeps{$t}}, [$t, 'Self'];
+    }
+
+    # Add explicit dependencies
+    for my $name (keys %{$deps}) {
+        for my $pair (@{$deps->{$name}}) {
+            push @{$pkgdeps{$name}}, $pair;
+        }
+    }
+    return %pkgdeps;
+}
+
+subtest 'graph: single target, no deps' => sub {
+    my %results = make_results(
+        { Name => 'foo', Version => '1.0-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['foo'], {});
+    my %pkgmap;
+
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    # Self edge should exist
+    is($dag->{foo}{foo}, 'Self', 'self edge for target foo');
+    is(scalar keys %{$dag_foreign}, 0, 'no foreign deps');
+};
+
+subtest 'graph: linear chain A -> B -> C' => sub {
+    my %results = make_results(
+        { Name => 'A', Version => '1.0-1' },
+        { Name => 'B', Version => '2.0-1' },
+        { Name => 'C', Version => '3.0-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['A'], {
+        'A' => [['B', 'Depends']],
+        'B' => [['C', 'Depends']],
+    });
+    my %pkgmap;
+
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is($dag->{A}{A}, 'Self',    'A self edge');
+    is($dag->{B}{A}, 'Depends', 'B required by A');
+    is($dag->{C}{B}, 'Depends', 'C required by B');
+};
+
+subtest 'graph: target with empty depends' => sub {
+    my %results = make_results(
+        { Name => 'leaf', Version => '0.1-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['leaf'], {});
+    my %pkgmap;
+
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is(scalar keys %{$dag}, 1, 'only one node in DAG');
+    ok(exists $dag->{leaf}, 'leaf exists as a DAG key');
+    is($dag->{leaf}{leaf}, 'Self', 'self edge exists');
+};
+
+subtest 'graph: foreign dependency' => sub {
+    my %results = make_results(
+        { Name => 'mypkg', Version => '1.0-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['mypkg'], {
+        'mypkg' => [['glibc', 'Depends']],
+    });
+    my %pkgmap;
+
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is($dag->{mypkg}{mypkg}, 'Self', 'self edge');
+    is($dag_foreign->{glibc}{mypkg}, 'Depends', 'glibc in foreign DAG');
+    ok(not(defined $dag->{glibc}), 'glibc not in main DAG');
+};
+
+subtest 'graph: provider ($provides=1)' => sub {
+    my %results = make_results(
+        { Name => 'mypkg',    Version => '1.0-1' },
+        { Name => 'libfoo',   Version => '2.0-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['mypkg'], {
+        'mypkg' => [['libfoo', 'Depends']],
+    });
+    # provider-pkg provides libfoo
+    my %pkgmap = ( 'libfoo' => ['provider-pkg', '2.0'] );
+
+    # $provides=1: provider takes precedence
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 1);
+
+    is($dag->{'provider-pkg'}{mypkg}, 'Depends',
+        'edge goes through provider-pkg when $provides=1');
+    ok(not(defined $dag->{'libfoo'}),
+        'libfoo not in DAG (replaced by provider-pkg)');
+};
+
+subtest 'graph: provider disabled ($provides=0)' => sub {
+    my %results = make_results(
+        { Name => 'mypkg',  Version => '1.0-1' },
+        { Name => 'libfoo', Version => '2.0-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['mypkg'], {
+        'mypkg' => [['libfoo', 'Depends']],
+    });
+    my %pkgmap = ( 'libfoo' => ['provider-pkg', '2.0'] );
+
+    # $provides=0: use the package itself
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is($dag->{libfoo}{mypkg}, 'Depends',
+        'edge goes to libfoo directly when $provides=0');
+    ok(not(defined $dag->{'provider-pkg'}),
+        'provider-pkg not in DAG when $provides=0');
+};
+
+subtest 'graph: multiple dep types' => sub {
+    my %results = make_results(
+        { Name => 'pkg',      Version => '1.0-1' },
+        { Name => 'libdep',   Version => '1.0-1' },
+        { Name => 'buildtool', Version => '2.0-1' },
+        { Name => 'checker',  Version => '0.5-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['pkg'], {
+        'pkg' => [
+            ['libdep',    'Depends'],
+            ['buildtool', 'MakeDepends'],
+            ['checker',   'CheckDepends'],
+        ],
+    });
+    my %pkgmap;
+
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is($dag->{libdep}{pkg},    'Depends',      'Depends edge');
+    is($dag->{buildtool}{pkg}, 'MakeDepends',  'MakeDepends edge');
+    is($dag->{checker}{pkg},   'CheckDepends', 'CheckDepends edge');
+};
+
+subtest 'graph: multiple targets each get self edges' => sub {
+    my %results = make_results(
+        { Name => 'X', Version => '1.0-1' },
+        { Name => 'Y', Version => '1.0-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['X', 'Y'], {});
+    my %pkgmap;
+
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is($dag->{X}{X}, 'Self', 'X self edge');
+    is($dag->{Y}{Y}, 'Self', 'Y self edge');
+};
+
+subtest 'graph: pkgdeps values are always arrayrefs (dead code proof)' => sub {
+    my %results = make_results(
+        { Name => 'test', Version => '1.0-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['test'], {});
+
+    # Verify the data structure: pkgdeps{test} is an arrayref
+    is(ref($pkgdeps{test}), 'ARRAY',
+        'pkgdeps entry is ARRAY ref, not a scalar (confirms #1252)');
+
+    # The dead code does: $pkgdeps->{$name} eq $name
+    # This can never be true for an arrayref
+    ok($pkgdeps{test} ne 'test',
+        'arrayref ne string is always true');
+};
+
+subtest 'prune: removes installed packages' => sub {
+    # Build a DAG: C <- B <- A (Self)
+    my %dag = (
+        'A' => { 'A' => 'Self' },
+        'B' => { 'A' => 'Depends' },
+        'C' => { 'B' => 'Depends' },
+    );
+
+    my @removed = prune(\%dag, ['C']);
+
+    ok((grep { $_ eq 'C' } @removed), 'C was pruned');
+    ok(not(defined $dag{C}), 'C removed from DAG');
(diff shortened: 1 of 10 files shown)


Strykar:
Dropped the bug-related content from this PR so it can merge without requiring verification of behavioral claims:

- Removed the three `pipeline: *` subtests that encoded the "target lost when provider exists" scenario, along with the `solve_pipeline` helper they used.
- Aligned the fixture-invariant harness to current upstream `solve()` semantics (prune every `$pkgmap` key). Dropped the "target survives pipeline" invariant and its aspirational-scope comment.

What remains is purely additive test infrastructure: unit tests for `graph()`, `prune()`, `tsort()`, `recurse()`; dep-type filtering; versioned-op boundary cases; cycle detection; and fixture JSON for the wiki comparison-table packages (`aws-cli-git`, `clion`, `liri-git-meta`, `nxagent`, `nxproxy`, `python2-cryptography`, `ruby-activemodel`, `samsung-unified-driver`).

All 44 subtests pass on current master.

The pipeline/provider behavior will be tracked separately as an issue + draft PR so it can be reviewed on a different timeline.

AladW:
Looks surprisingly good. I still have to verify `depends.t` for errors, but then this can be merged.

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+subtest 'graph: target with empty depends' => sub {
+    my %results = make_results(
+        { Name => 'leaf', Version => '0.1-1' },
+    );
+    my %pkgdeps = make_pkgdeps(['leaf'], {});
+    my %pkgmap;
+
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    # Count before sub-key access so autoviv can't mask an empty $dag.
```
AladW: Gibberish comment

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    my %pkgdeps = make_pkgdeps(['mypkg'], {
+        'mypkg' => [['libfoo', 'Depends']],
+    });
+    # provider-pkg provides libfoo
+    my %pkgmap = ( 'libfoo' => ['provider-pkg', '2.0'] );
+
+    # $provides=1: provider takes precedence
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 1);
+
+    is($dag->{'provider-pkg'}{mypkg}, 'Depends',
```
AladW, line 126: If provider-pkg provides `libfoo`, `libfoo` should not be in the DAG (`libfoo` could be virtual, or an actual package named as such.) 
```
ok(not(defined $dag->{'libfoo'}), 'libfoo provided by provider-pkg when $provides=1');
```
similar to subtest `graph: provider disabled`

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    # $provides=0: use the package itself
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is($dag->{libfoo}{mypkg}, 'Depends',
+        'edge goes to libfoo directly when $provides=0');
+    ok(not(defined $dag->{'provider-pkg'}),
+        'provider-pkg not in DAG when $provides=0');
+};
+
+# foo is the target, bar provides foo. Two subtests: one pins
```
AladW: This is way too much just for the "Self" issue. The goal here is the behavior on empty (!) depends - not the implementation detail through `Self`. Leave out provider_redirect_inputs and both `Self` subtests

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    my ($results, $pkgdeps, $pkgmap) = provider_redirect_inputs();
+    my ($dag) = graph($results, $pkgdeps, $pkgmap, 0, 1);
+
+    TODO: {
+        local $TODO = 'waiting on #1256';
+        is($dag->{foo}{foo}, 'Self', 'foo keeps its own self-edge');
+    }
+};
+
+subtest 'graph: versioned dependency passes vercmp' => sub {
```
AladW: Add subtests for other operators:

* `<` (`libbar`, `Version => 2.0-1`)
* `>` (`libbar`, `Version => 3.0-1`)
* `<=` (`libbar`, `Version => 3.0-1`)
* `=` (`libbar`, `Version => 3.0-1`)

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+
+    my @removed = prune(\%dag, ['C']);
+
+    ok((grep { $_ eq 'C' } @removed), 'C was pruned');
+    ok(not(defined $dag{C}), 'C removed from DAG');
+    ok(defined $dag{A}, 'A remains in DAG');
+    ok(defined $dag{B}, 'B remains in DAG');
+};
+
+subtest 'prune: cascading removal of orphaned deps' => sub {
```
AladW, line 220: A proper test for recursive ("cascading") removal should include more than 1 level. This has been an issue in previous implementations.
```perl
my %dag = (
'A' => { 'A' => 'Self' },
'B' => { 'A' => 'Depends' },
'C' => { 'B' => 'Depends' }
);

my @removed = prune(\%dag, ['A']);

ok((grep { $_ eq 'A' } @removed), 'A was pruned');
ok((grep { $_ eq 'B' } @removed), 'B was pruned (cascading)');
ok((grep { $_ eq 'C' } @removed), 'C was pruned (cascading)');
is(scalar keys %dag, 0, 'DAG is empty after cascading prune');
```

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+subtest 'recurse: single target, no deps' => sub {
+    my $callback = sub {
+        my ($deps) = @_;
+        my %db = (
+            'pkg-a' => {
+                Name    => 'pkg-a',
+                Version => '1.0-1',
+            },
+        );
+        return map { $db{$_} } grep { defined $db{$_} } @{$deps};
```
AladW: The `grep` does little here. If we make a typo in the `recurse` call, the test would fail, but without explicit `Use of uninitialized value`.

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+                Version  => '1.0-1',
+                Provides => ['virt=1.0'],
+            },
+            'second' => {
+                Name     => 'second',
+                Version  => '2.0-1',
+                Provides => ['virt=2.0'],
+            },
+        );
+        # Return in deterministic order: first before second
```
AladW, line 464: This is a good catch actually. Future tests of AUR::Query.pm and AUR::Json.pm should verify that outputs match input order, since these are the callbacks we use in real use cases.

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+        my %results = make_results(
+            { Name => 'app',    Version => '1.0-1' },
+            { Name => 'libold', Version => '1.0-1' },
+        );
+        my %pkgdeps = make_pkgdeps(['app'], {
+            'app' => [['libold>=5.0', 'Depends']],
+        });
+        my %pkgmap;
+
+        # verify=1: vercmp(1.0, 5.0, >=) should fail
```
AladW: As above we want to also compare against the other operators (`<=`, `<`, `>`, `==`). Either
* Make sure the operators are tested in `AUR::Vercmp`; then we can assume operators to be correct here
* We duplicate test cases with different data
* or use an array of cases per operator.

Compare `subtest 'graph: versioned dep operators'`

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    );
+
+    is_deeply($results->{'new-app'}{'Conflicts'}, ['old-app', 'rival'],
+        'Conflicts preserved verbatim in results');
+    is_deeply($results->{'new-app'}{'Replaces'}, ['old-app'],
+        'Replaces preserved verbatim in results');
+};
+
+subtest 'graph: versioned dep fails vercmp' => sub {
+    my $pid = fork();
```
AladW: Instead of using `fork` and exit code tricks, we should 

1. use `croak` in the Depends.pm module on invalid inputs (cf. https://www.perlmonks.org/?node_id=685452)
2. use `eval` or `Tiny::Try` in the tests.

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    my $exit_code = $? >> 8;
+    is($exit_code, 1, 'graph exits with EX_FAILURE on version mismatch');
+};
+
+subtest 'tsort: cycle detection' => sub {
+    # A -> B -> A (cycle), plus self-loops
+    my @input = ('A', 'A', 'A', 'B', 'B', 'B', 'B', 'A');
+    my @sorted;
+
+    # Capture stderr via tempfile
```
AladW: We're testing a Perl module, not an executable - we should retrieve the error/cycle from `croak()` called in `AUR::tsort`.

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    # Neither A nor B can be output since both have remaining predecessors
+    is(scalar @sorted, 0, 'no nodes output for pure cycle');
+};
+
+subtest 'tsort: partial cycle with sortable nodes' => sub {
+    # C -> A -> B -> A (A-B cycle), C has no predecessors
+    my @input = ('C', 'C', 'C', 'A', 'A', 'B', 'B', 'A');
+    my @sorted;
+
+    my ($tmp_fh, $tmp_fn) = tempfile(UNLINK => 1);
```
AladW: We're testing a Perl module, not an executable - we should retrieve the error/cycle from croak() called in `AUR::tsort`.

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    my ($dag, $dag_foreign) = graph(\%results, \%pkgdeps, \%pkgmap, 0, 0);
+
+    is(scalar keys %{$dag}, 0, 'empty DAG');
+    is(scalar keys %{$dag_foreign}, 0, 'empty foreign DAG');
+};
+
+subtest 'recurse: no results exits with EX_FAILURE' => sub {
+    my $callback = sub { return (); };
+
+    my $pid = fork();
```
AladW: Instead of using `fork` and exit code tricks, we should 

1. use `croak` in the Depends.pm module on invalid inputs (cf. https://www.perlmonks.org/?node_id=685452)
2. use `eval` or `Tiny::Try` in the tests.

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+subtest 'tsort: single non-self pair' => sub {
+    my @input = ('A', 'B');
+    my @sorted = tsort(0, \@input);
+
+    is(scalar @sorted, 2, 'two elements');
+    is($sorted[0], 'A', 'A first (no predecessors)');
+    is($sorted[1], 'B', 'B second (successor of A)');
+};
+
+subtest 'prune: max fan-in removal' => sub {
```
AladW: Fan-in is jargon (for logic gates?), find some better title. Also this should be moved to the other `prune` tests (L254+)

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    my @removed = prune(\%dag, ['base']);
+
+    ok((grep { $_ eq 'base' } @removed), 'base was pruned');
+    ok((grep { $_ eq 'dep-a' } @removed), 'dep-a cascaded');
+    ok((grep { $_ eq 'dep-b' } @removed), 'dep-b cascaded');
+    ok((grep { $_ eq 'dep-c' } @removed), 'dep-c cascaded');
+    is(scalar keys %dag, 0, 'DAG fully emptied');
+};
+
+subtest 'prune: leaf removal does not cascade upward' => sub {
```
AladW: Move to the other prune tests (L254+)

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+    for my $fixture_name (@fixtures) {
+        my $fixture_path = "$fixture_dir/$fixture_name";
+
+        subtest "fixture: $fixture_name" => sub {
+            my ($callback, $db) = fixture_callback($fixture_path);
+            my @targets = sort keys %{$db};
+
+            # Primary target is the directory name; fall back to
+            # the first package if the dir-name package isn't in
+            # the DB (malformed fixture)
```
AladW: Error instead of falling back on malformed fixtures

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+                );
+                ($dag, $dag_foreign) = graph(
+                    $results, $pkgdeps, $pkgmap, 0, 1
+                );
+
+                open(STDERR, '>&', $save) or croak "restore stderr: $!";
+            }
+
+            # --- Invariant 1: target has self-edge (pre-prune) ---
+            # Use exists + explicit inner lookup; a plain
```
AladW: Gibberish comment

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+                if (exists $dag->{$target}) {
+                    is($dag->{$target}{$target}, 'Self',
+                        "target $target has self-edge");
+                }
+                else {
+                    fail("target $target has self-edge (no DAG node)");
+                }
+            }
+
+            # Mirror upstream solve(): prune every $pkgmap key.
```
AladW: `prune every $pkgmap key ($provides=1)`

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+
+            # --- Invariants 4 & 5: tsort properties ---
+            my @pairs;
+            for my $dep (keys %{$dag}) {
+                for my $name (keys %{$dag->{$dep}}) {
+                    push @pairs, $dep, $name;
+                }
+            }
+
+            seek($tmp_fh, 0, 0);
```
AladW: As above; use `eval` or `Tiny::Try`

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+            }
+
+            # Invariant 4: no duplicates in tsort output
+            my %seen;
+            my @dupes = grep { $seen{$_}++ } @sorted;
+            is(scalar @dupes, 0, 'tsort output has no duplicates');
+
+            # Invariant 5: in the no-cycle case tsort emits every
+            # DAG node; on cycles it emits fewer. Check the stderr
+            # we captured for a cycle warning, then assert the
```
AladW: Second comment is gibberish

perl/t/depends.t
```diff
@@ -0,0 +1,925 @@
...
+            # right shape — not just a weak upper bound.
+            seek($tmp_fh, 0, 0);
+            my $cycle_stderr = do { local $/ = undef; <$tmp_fh> };
+            my $dag_nodes    = scalar keys %{$dag};
+            my $sorted_nodes = scalar @sorted;
+            if ($cycle_stderr =~ /cycle detected/) {
```
AladW: As above; use `croak` in `AUR::tsort` instead of parsing stderr

AladW (changes requested):

AladW:
Also, it should be added that the Claude commit messages are comically bad -- throwing around jargon for the sake of it or just being plain gibberish. Before merge all commits should be squashed with a manually edited commit message.

lib/aurweb/aur-depends
```diff
@@ -155,8 +155,17 @@ unless(caller) {
 
     # Retrieve AUR results
     #    JSON -> hash -> extract depends[] -> repeat until none -> prune
-    my ($results, $dag, $dag_foreign) = solve(\@ARGV, \@types, \&callback_query,
-        $opt_verify, $opt_provides, $opt_installed);
+    my ($results, $dag, $dag_foreign) = eval {
+        solve(\@ARGV, \@types, \&callback_query,
+              $opt_verify, $opt_provides, $opt_installed);
+    };
+    if ($@) {
```
AladW: Alternative to eval: https://perldoc.perl.org/perlsyn#Try-Catch-Exception-Handling

The regex parsing of the errors is baffling (in a bad way.) With the `try` feature above:
```
try {
... = solve(...);
} catch ($e) {
  exit($e =~ /out of range/ ? 34 : 1);
}
```
You could also croak on the exit code (leaving EX_FAILURE, EX_OUT_OF_RANGE in Depends.pm) and print the messages in `aur-depends`. Either way, it's brittle, but a general approach like `Exception:Class` is heavy on dependencies.

AladW: If you leave `EX_*` in Depends.pm, you could use it as prefix:
```
croak EX_FAILURE . __PACKAGE__ . ...
```
for a more predictable error format.

AladW: Or, we follow https://perldoc.perl.org/perlobj, implement an AUR::Exception class and use `bless`:
```perl
package AUR::Exception;
use strict;
use warnings;

# Called when printing the object (nomethod)
use overload '""' => \&message;

sub new {
    my ($class, $message, $exit_code) = @_;
    # Default to exit code 1 if none is provided
    $exit_code //= 1; 
    return bless { message => $message, exit_code => $exit_code }, $class;
}

sub message   { return $_[0]->{message} }
sub exit_code { return $_[0]->{exit_code} }
```
in `AUR:Depends`
```perl
croak AUR::Exception->new("out of range", EX_OUT_OF_RANGE);
```
in `lib/aur-depends`
```perl
use Scalar::Util 'blessed';
...

# Or feature 'try' / Tiny::Try
eval {
    ... = solve
};

# Catch the exception
if (my $error = $@) {
    # Check if the error is an AUR:Exception
    if (blessed($error) && $error->isa('AUR::Exception')) {
        exit($error->exit_code);
    } else {
        # Fallback for unspecified errors
        warn "Unexpected error: $error";
        exit(1);
    }
}
```

Strykar: Went with the EX_FAILURE-prefix suggestion.

perl/t/depends.t
Strykar: Acknowledged and made a note; out of scope for this PR

lib/aurweb/aur-depends
```diff
@@ -155,8 +155,19 @@ unless(caller) {
...
+        my $err = $@;
+        $err =~ s/\s+at\s+\S+\s+line\s+\d+\.?\s*\z//;
```
AladW: Let's go with AUR:Exception (AUR:Except?) which is more general than parsing stderr messages, regardless of how you format them.

AladW (changes requested):

lib/aurweb/aur-depends
AladW: Note that AUR::Exception.pm can be a separate PR, it doesn't have to be implemented in this PR.

Strykar: Will create a new PR for `AUR::Exception` as a follow-up.

AladW:
Needs rebase

Strykar:
Rebased

AladW:
:+1:

PR #1256, opened 2026-04-13
aur-depends: anchor self-edges to explicit targets (fix #1255)

### LINKED ISSUE #1255: aur-depends: explicit target stripped from DAG when another resolved package Provides it
## Summary

When an explicit target of `aur-depends` happens to be `Provides:`d by another package that `solve()` also resolves, the target is removed from the DAG by the `prune(%$pkgmap)` step, leaving it with no build-order position. Its own dependencies are then orphaned and pruned cascadingly.

I'm filing this as a repro only — happy to defer any behavioral fix to maintainer judgement.

## Repro (self-contained)

Requires `AUR::Depends` on `@INC` (run from repo root with `-Iperl` or drop the script into `perl/`):

```perl
#!/usr/bin/env perl
use strict;
use warnings;
use AUR::Depends qw(recurse graph prune);

my $callback = sub {
    my ($deps) = @_;
    my %db = (
        foo  => { Name => 'foo',  Version => '1.0-1', Depends  => ['libx'] },
        bar  => { Name => 'bar',  Version => '2.0-1', Provides => ['foo=1.0'],
                  Depends => ['liby'] },
        libx => { Name => 'libx', Version => '1.0-1' },
        liby => { Name => 'liby', Version => '1.0-1' },
    );
    return map { $db{$_} } grep { defined $db{$_} } @{$deps};
};

# User asks to build "foo" and "bar". bar happens to Provide foo.
my ($results, $pkgdeps, $pkgmap) = recurse(['foo','bar'], ['Depends'], $callback);
my ($dag) = graph($results, $pkgdeps, $pkgmap, 0, 1);   # $provides=1

# Same prune solve() does in aur-depends:
prune($dag, [keys %$pkgmap]);

for my $t (qw(foo bar libx liby)) {
    printf "%-5s  results=%s  dag_node=%s\n",
        $t,
        (defined $results->{$t} ? 'yes' : 'no '),
        (defined $dag->{$t}     ? 'yes' : 'no ');
}
```

## Observed output

```
foo    results=yes  dag_node=no
bar    results=yes  dag_node=yes
libx   results=yes  dag_node=no
liby   results=yes  dag_node=yes
```

- `foo` remains in `%results` but has no node in the DAG, so it won't appear in the `tsort` build order.
- `libx` (foo's dep) is orphaned and disappears from the DAG too.
- `bar` and `liby` survive because `bar` has its own self-edge through its own pkgdeps.

## Where this happens

In [`perl/AUR/Depends.pm`](https://github.com/aurutils/aurutils/blob/master/perl/AUR/Depends.pm):

- `graph()` at lines 211-220: when `$provides` is on and `$pkgmap` has an entry for a dep, the edge is rewritten to go through the provider. The Self edge seeded for `foo` in `recurse()` becomes `$dag{bar}{foo} = 'Self'` because `foo` has a provider (`bar`). So the DAG never holds a `$dag{foo}{...}` row.
- `solve()` in `aur-depends` then prunes `keys %$pkgmap` (which includes `foo` as a virtual). The inner-key `$dag{bar}{foo}` gets deleted, and `foo` has no own row to fall back on.

Note also the dead-looking guard at line 192: `defined $pkgdeps->{$name} and $pkgdeps->{$name} eq $name` — `$pkgdeps->{$name}` is always an arrayref seeded at line 70, so that string comparison can never be true. The Self edge reaches the DAG via the inner loop at line 220 instead.

## Impact

Only triggers when a user-named target is also `Provides`'d by a sibling target or a transitively-pulled package — uncommon but not unheard of when users pass versioned splits plus their meta-package, or provider/real-package pairs.

## Not proposing a fix

Two obvious shapes (exclude targets from the prune list; give targets a non-redirected Self edge in `graph()`) both have tradeoffs around other `provides=1` behavior. Deferring to maintainer judgement.

## Related

- Executable tests for this behavior will live in a separate draft PR so they can be reviewed on a different timeline from #1253 (purely additive test infra).

Strykar:
Fix #1255: when one explicit target Provides another, the explicit target loses its node in the resolved graph.

Reproduction: `bitlbee-libpurple-git` is an AUR package that Provides: bitlbee, and bitlbee is also an AUR package. On master, `aur depends bitlbee bitlbee-libpurple-git --json` returns a JSON object where `.bitlbee.RequiredBy` is null.

### `graph()` in `perl/AUR/Depends.pm`

Skip provider redirection on Self edges:

```perl
if ($provides and $dep_type ne 'Self' and defined $pkgmap->{$dep_name}) {
    ($prov_name, $prov_ver) = @{$pkgmap->{$dep_name}};
}
```

Without this, the Self edge that `recurse()` seeds for each command-line target was rewritten from `dag{target}{target} = 'Self'` to `dag{provider}{target} = 'Self'`, leaving the explicit target with no node of its own.

### `solve()` in `lib/aurweb/aur-depends`

When building the prune list from `pkgmap` keys, exclude explicit targets:

```perl
if ($opt_provides) {
    my %is_target = map { $_ => 1 } @{$targets};
    push @to_prune, grep { not $is_target{$_} } keys %{$pkgmap};
}
```

Without this, the just-anchored Self edge gets wiped, since the target name is also a `pkgmap` key and `prune()` removes anything keyed under one.

## Why both halves

Either one alone produces a different broken state. Verified by running the actual `aur-depends` script in four configurations against the same recursed AUR data:

| variant | bitlbee in results? | bitlbee.RequiredBy |
|---|---|---|
| master (no fix) | yes | absent (the bug) |
| graph fix only | **no** (target dropped from output) | n/a |
| solve filter only | yes | absent, edge attributed to wrong owner |
| both fixes | yes | `{bitlbee: 'Self'}` (correct) |

## Verification

```
$ aur depends bitlbee bitlbee-libpurple-git --json | jq '.bitlbee.RequiredBy'
{
  "bitlbee": "Self"
}
```

## Test plan

- [x] `perl/t/depends.t` passes (42 tests, including the fixture suite from #1253)
- [x] Real-world reproduction `aur depends bitlbee bitlbee-libpurple-git` produces the correct DAG
- [ ] Fixture-based regression tests (`test-depends-provides-regression` branch) to be submitted as a follow-up PR after this lands

### DIFF
--- a/lib/aurweb/aur-depends
+++ b/lib/aurweb/aur-depends
@@ -24,8 +24,12 @@ sub solve {
     # Targets to be removed from the graph
     my @to_prune;
 
-    # Remove virtual dependencies from dependency graph (#1063)
-    push @to_prune, keys %{$pkgmap} if $opt_provides;
+    # Remove virtual dependencies from dependency graph (#1063), but
+    # keep explicit targets even if a sibling target Provides them (#1255).
+    if ($opt_provides) {
+        my %is_target = map { $_ => 1 } @{$targets};
+        push @to_prune, grep { not $is_target{$_} } keys %{$pkgmap};
+    }
 
     # Remove transitive dependencies for installed targets (#592)
     push @to_prune, @{$opt_installed} if $opt_installed && @{$opt_installed};
--- a/perl/AUR/Depends.pm
+++ b/perl/AUR/Depends.pm
@@ -206,7 +206,7 @@ sub graph {
                 # $provides is false.
                 my  ($prov_name, $prov_ver) = ($dep_name, $dep_ver[0]);
 
-                if ($provides and defined $pkgmap->{$dep_name}) {
+                if ($provides and $dep_type ne 'Self' and defined $pkgmap->{$dep_name}) {
                     ($prov_name, $prov_ver) = @{$pkgmap->{$dep_name}};
                 }
 


fredericboileau:
The push to prune of the keys of pkgmap is redundant. If opt_provides is true then virtual packages are resolved in graph subroutine:
```
if ($provides and defined $pkgmap->{$depname}) {
    ($prov_name, $prov_ver) = @{$pkgmap->{$dep_name}
}
```
and then `$prov_name` is what is added to the dag

AladW:
Was this tested with `tests/depends`? That includes a real case for `provides` - assuming the relevant packages still exist.

Strykar:
Reproduces today against `bitlbee` `bitlbee-libpurple-git`: `bitlbee-libpurple-git` Provides: `bitlbee`.

On master:

    $ aur depends bitlbee bitlbee-libpurple-git --json | jq '.bitlbee.RequiredBy'
    null

After this PR:

    $ aur depends bitlbee bitlbee-libpurple-git --json | jq '.bitlbee.RequiredBy'
    {
      "bitlbee": "Self"
    }

@fredericboileau Can you confirm?

I was wrong earlier to drop the `solve()` change. Both halves are necessary: `graph()` must not redirect Self edges through `pkgmap`, and `solve()` must not push explicit targets onto the prune list.

Verified the four combinations (no fix / graph only / solve only / both) and only both-together produces the correct DAG.

fredericboileau:
I'm not sure I understood correclty but from what I got if you're going to push to prune keys from pkgmap you *do* need the grep to exclude self edges. What I was saying is you don't need to push the keys of pkgmap at all since they are never inserted in the dag in the first place. All it needs is the already installed packages

PR #1257, opened 2026-04-13
tests: regression coverage for #1255 (provider prunes explicit target)

Strykar:
Draft companion to issue [#1255](https://github.com/aurutils/aurutils/issues/1255) and fix PR [#1256](https://github.com/aurutils/aurutils/pull/1256). Adds `perl/t/depends_provides.t` which loads the real `lib/aurweb/aur-depends` via `do` and exercises in-tree `solve()` plus the formatter layer (`pairs()` / `RequiredBy`) against three small in-memory package DBs.

## Structure

The aspirational DAG-shape and `pairs()` assertions are wrapped in `TODO: { local \$TODO = '...'; ... }` blocks scoped as tightly as possible. Everything that already holds on current master — `\$dag_foreign` emptiness, `%results` membership of explicit targets — lives **outside** the TODO guard so any regression there fails loudly.

Today the TODO-wrapped assertions are expected-fail. Once [#1256](https://github.com/aurutils/aurutils/pull/1256) (or any equivalent fix) lands, they flip to **unexpected-pass** and `prove` forces an intentional TODO removal.

## Scenarios

1. **Sibling target provides an explicit target** (`['foo','bar']`, `bar` provides `foo`) — DAG shape under TODO.
2. **Provides disabled** (`\$provides=0`) — control path, no TODO, passes today.
3. **Target's own dependency provides it** (`['foo']`, `foo` depends on `bar`, `bar` provides `foo`) — DAG shape + `pairs()` output under TODO.
4. **Formatter control** — pins `pairs()` output (by `Name` and by `PackageBase`) on the clean `\$provides=0` DAG outside any TODO, so a regression in `RequiredBy` population or pair dedup surfaces without a guard.

## Why draft

No fix is proposed here. This PR is executable documentation of the behavior described in #1255, safe to ship independently of the fix's merge timeline. Marked draft so it can sit alongside the issue on the maintainer's schedule.

## Test plan
- [x] `cd perl && prove -I. t/depends_provides.t` — all 4 subtests pass; TODO assertions correctly report as expected-fail with meaningful `is_deeply` diffs.
- [x] Cherry-picking PR #1256 on top flips the TODO assertions to unexpected-pass, matching the intended signal.
- [x] Control subtests (`\$provides=0` path, formatter output) pass today with no TODO guard.

## Notes on test hardening included here

- `is_deeply` against the full DAG shape instead of chained `\$dag->{X}{Y}` probes — avoids autovivification side effects that would mask an empty-DAG regression.
- `do \$script` bootstrap distinguishes compile errors (`\$@`), I/O failures (`\$!`), and scripts that load but don't define `solve()`.
- Formatter-layer coverage drives `pairs()` through the same `RequiredBy` population the main block does.

### DIFF
--- a/perl/t/depends_provides.t
+++ b/perl/t/depends_provides.t
@@ -15,8 +15,15 @@ use FindBin qw($Bin);
 use File::Spec;
 
 my $script = File::Spec->catfile($Bin, '..', '..', 'lib', 'aurweb', 'aur-depends');
-do $script;
-BAIL_OUT("unable to compile $script: $@") if $@;
+my $rv = do $script;
+if (not defined $rv) {
+    # do() returns undef for both compile errors ($@ set) and
+    # I/O failures ($! set). Keep them distinct so a misplaced
+    # script doesn't masquerade as a syntax problem.
+    BAIL_OUT("unable to compile $script: $@") if $@;
+    BAIL_OUT("unable to read $script: $!")    if $!;
+    BAIL_OUT("do $script returned undef (script ended in falsy value?)");
+}
 BAIL_OUT("$script loaded but solve() not defined") if not exists &solve;
 
 sub solve_with_db {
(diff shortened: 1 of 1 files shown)


Strykar:
Closing this draft for now. I'll keep any follow-up regression coverage on my fork and revisit depends.t later if it still seems useful.

PR #1259, opened 2026-04-14
Update bash-completion to the 2.12 API

Strykar:
Brings the bash completion close to zsh's fidelity and moves it onto
the modern bash-completion 2.12 API.

Split into two commits so the test harness in the second can be
dropped without affecting the completion itself.

## Commit 1 — bash completion proper
`completions/bash/aurutils.in`, `completions/command_opts.sh`

- Typed argument completion: file, dir, repo, pkgname, pkgbase,
  local_pkg, user, attr. Tables live in `command_opts.sh`, values
  transcribed from the zsh completion so both shells share one
  metadata source.
- Static choice sets for `--by`, `--type`, `--key`, `--format`,
  `--sync`, `--upair`.
- Positional completion (pkgname / pkgbase / local package / dir)
  per subcommand.
- `_comp_initialize -s` replaces `_get_comp_words_by_ref`: handles
  COMPREPLY reset, redirect detection, and `--opt=val` splitting in
  one call. Drops the hand-written `--flag=VALUE` branch.
- `_comp_compgen`, `_comp_compgen_filedir`, `_comp_delimited`
  replace each `COMPREPLY=($(compgen ...))` and `_filedir` call.
- Comma-separated value completion for `aur sync --ignore=a,b,c`
  (local packages) and `--provides-from=/a,/b` (absolute + relative
  paths, mid-list suggestions).

## Commit 2 — optional test harness
`tests/completion`, `Makefile`

35 cases: subcommands, static choice sets, `--opt=val` split,
dynamic repo / attr / local-pkg completion, mid-list delimited
paths, negative cases. Wired into `make test`. Skips gracefully on
systems without bash-completion 2.12 or without the `aur` binary.

If the test harness isn't wanted, cherry-pick only the first commit.

## Test plan
- [x] `make -C completions bash`
- [x] `make test` — 35/0
- [x] `shellcheck` clean (same two warnings as master: SC2148, SC2206)
- [x] Manual verification in bash with real `aur` 20.5.8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ shellcheck: aur
 
 test: aur shellcheck
 	@tests/parseopt-consistency
+	@tests/completion
 	@$(MAKE) -C perl test
 
 install-aur: aur
--- a/completions/bash/aurutils.in
+++ b/completions/bash/aurutils.in
@@ -1,79 +1,127 @@
 #!/bin/bash
+#
+# Template for the aurutils bash completion. Generates `bash/aur`,
+# installed to /usr/share/bash-completion/completions/. Requires
+# bash-completion 2.12+ at runtime.
 
 cat <<'EOF'
+# aurutils bash completion — requires bash-completion >= 2.12
 _aur_completion()
 {
-    local cur prev words cword
-    _get_comp_words_by_ref cur prev words cword
+    # comp_args is populated by _comp_initialize and declared local
+    # here so it doesn't leak into the global scope.
+    # shellcheck disable=SC2034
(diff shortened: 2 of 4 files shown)


AladW:
I don't know enough about completion to review an LLM completion PR. In general I fail to see the benefits here - the code becomes (even more) complicated just to use some new bash API.

If the completion test works with the current bash completion, that could be merged seperately.

Strykar:
Closing this. Following your suggestion to land the test against the current completion separately — opening a new PR with just tests/completion-bash and the make-test wire-up.

PR #1260, opened 2026-04-14
completions: add support for the fish shell

Strykar:
New `completions/fish/aur.fish` with per-subcommand flag sets, descriptions, static choice lists, and dynamic completion for repositories, AUR packages, local-repo packages, and repo-add attributes. Helpers __aur_list_* wrap the same shell-outs the zsh helpers use.

Installs via the new `install-fish` target to the standard `fish/vendor_completions.d` path.

Co-Authored-By: Claude Opus

### DIFF
--- a/completions/Makefile
+++ b/completions/Makefile
@@ -1,6 +1,6 @@
 PREFIX ?= /usr
 SHRDIR ?= $(PREFIX)/share
-.PHONY = bash install-bash install-zsh
+.PHONY = bash install-bash install-zsh install-fish
 
 ZSH_SITE_FUNCTIONS = \
 	zsh/_aur \
@@ -23,3 +23,6 @@ install-zsh: zsh/run-help-aur $(ZSH_SITE_FUNCTIONS)
 
 	@install -d '$(DESTDIR)$(SHRDIR)/zsh/site-functions/'
 	@install -m644 -t '$(DESTDIR)$(SHRDIR)/zsh/site-functions/' $(ZSH_SITE_FUNCTIONS)
+
+install-fish: fish/aur.fish
+	@install -Dm644 fish/aur.fish -t '$(DESTDIR)$(SHRDIR)/fish/vendor_completions.d'
--- a/completions/fish/aur.fish
+++ b/completions/fish/aur.fish
@@ -1,0 +1,266 @@
+# Fish completion for aurutils (aur wrapper)
+#
+# Transcribed from the zsh completion with parity as close as
+# fish's model allows. Third-party aur-<subcommand> extensions
+# are expected to ship their own completions.
+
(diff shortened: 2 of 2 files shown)


Earnestly:
Could more of these be generated from the hidden `--dump-options` flag that each of the utils exposes for this purpose?

Strykar:
Good call — looked into this properly. Two-part answer:

**Fish side:** I can mirror the same build-time generation the bash template already does (`aurutils.in` + `command_opts.sh`), driven by `--dump-options`. Descriptions (fish's main value-add over bash) would live in a hand-maintained table keyed by `cmd,flag`, since `--dump-options` only emits names. That gets fish to parity with bash for the 10 utils that currently implement `--dump-options`.

**Perl side:** The remaining 5 utils (`aur-depends`, `aur-search`, `aur-query`, `aur-format`, `aur-repo-parse`) don't implement `--dump-options` at all. To add it cleanly, each `GetOptions(...)` call would be refactored to take a pre-declared spec array, and a shared `AUR::Options::dump_options` helper in `perl/AUR/Options.pm` would walk that array — same source-of-truth pattern the bash utils use with their `opt_long` arrays. I read every `GetOptions` call: they only use `=s` and `|`-aliases (no `!`, `+`, `=i`, `:s`, `=s@`), so the dumper stays small and easy to verify.

**Out of scope:** `aur-graph` is gawk and takes `-v VAR=value` assignments, not CLI flags, so there's nothing for `--dump-options` to emit.

**Scope:** ~11 files — 3 for fish (new `.in` + template + Makefile), 7 for the Perl refactor (`Options.pm` + 5 utils + `command_opts.sh` to move them from `no_optdump` to `have_optdump`), plus `tests/completion`. Happy to split into two PRs if you'd prefer to review them independently.

This okay?

AladW:
Yeah, I would not merge this if the completions are not updated automatically. If this cannot be done with fish, I'm fine to not have fish support.

----

We can leave out the descriptions.

Having a `--dump-option` equivalent in `AUR::Options` sounds like a good idea.

Strykar:
Before I start the rewrite, should we extend the `--dump-options` format to include type hints?

One PR per group: bash scripts first, then Perl scripts once the `AUR::Options` helper lands. Both shells pick up typed completion simultaneously once the data is there.

Or would you rather keep `--dump-options` as-is and land the fish template against the current flat output first?

AladW:
Maybe you can try to rephrase the robot's output yourself, because the overly complicated, jargon-focused style is increasingly annoying to deal with. 🫤 

In any case, an AUR::Options.pm implementation should match the bash --dump-options. Things can be extended later.

Strykar:
Closing, will open this after two refactor PRs.

PR #1263, opened 2026-04-18
Options.pm: add dump_options and enable --dump-options in Perl scripts

Strykar:
The five Perl-based `aur-` scripts (`depends`, `search`, `query`, `format`, `repo-parse`) had no `--dump-options` support, so their completions were hardcoded as empty in `command_opts.sh`.

Add a `dump_options()` helper to `AUR::Options` that parses the `GetOptions` spec and prints flags in the same format the bash scripts already use. Each Perl script now accepts `--dump-options` and passes its spec through the helper.

Update `command_opts.sh` to include them in `have_optdump`. Only `aur-graph` (`gawk`, no flags) remains in `no_optdump`.

Co-Authored-By: Claude Opus

### DIFF
--- a/completions/command_opts.sh
+++ b/completions/command_opts.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
 
-have_optdump=('build' 'chroot' 'fetch' 'pkglist' 'repo' 'repo-filter'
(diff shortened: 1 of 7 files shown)


AladW:
LGTM

